PR #195, opened 2017-07-26
PDF Extract from full specs compliant json renderer - GSOREB-212

voisardf:
Linked to https://jira.camptocamp.com/browse/GSOREB-212

DO NOT MERGE YET! you may play with it though... 😃 

PDF templates based on the renderer of PR https://github.com/camptocamp/pyramid_oereb/pull/137 GSOREB-181 Add PDF extract.
* [x] Localization in function of the selected language (@sbrunner)
* [x] Get graphics from URL's (@sbrunner)
* [x] Get appendices list working in the table of content (@sbrunner)
* [x] Include legend data (@voisardf)
* [x] Include references list (@voisardf)
* [x] Include ExclusionOfLiability texts (@voisardf)
* [ ] Complete test data for "all" cases (@voisardf)


voisardf:
ok, the make serve works now and I get a reduced extract! thanks
but not the full PDF - null pointer exception... I'll have a look at it tomorrow.
I also get the topic map, but without the cadastral baselayer underneath

voisardf:
Allright, I got a very interesting behaviour on my instance:
- The NullPointerException does only happen when calling the full PDF for the first time after starting to serve
- I changed the names of the Logo files in the pyramid_oereb_standard.yml, then stopped the app and started serving again - and got the never ending loop on filling up wsgi1 like two weeks ago. Undo the changes in the yaml and restart gets the app back to normal working

voisardf:
I now found out, where the loop comes from: 
The referenced image file has to be on the same level as the pyramid_oereb_standard.yml as defined by now, otherwise we got the loop - so I think there is a error handling to improve here, maybe?
I also find it a bit confusing that the logos are in the pyramid_oereb folder on first level and again in the pyramid_oereb\pyramid_oereb\standard folder.
Is it possible to have only one copy of those files?

sbrunner:
The one at the first level is just an instantiation of the on in `pyramid_oereb\standard` ... Then I don't understand the issue...

PR #219, opened 2017-08-23
GSOREB-226 Add travis config to get cron build (daily)

ger-benjamin:
@sbrunner to discuss


ger-benjamin:
It can magically get the password with the "secure" params you add in the .travis.yaml ? (For now, I expect a "TRAVIS_PASSWORD" env value)

PR #227, opened 2017-08-30
Add deployment on a Docker container

sbrunner:
Also cleanup the git ignores


Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM camptocamp/c2cwsgiutils:0
+LABEL maintainer "info@camptocamp.org"
+
+WORKDIR /app
+
+RUN mkdir --parent /app/docker
```
pvalsecc: Not needed. COPY will create it

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+COPY . /app
+
+RUN pip install --disable-pip-version-check --no-cache-dir --editable . && \
+    mkdir /etc/pyramid_oereb && \
+    mv docker/config.yaml /etc/pyramid_oereb/ && \
+    mv docker/production.ini .
+
+ENV LOGO_CANTON logo_canton.png
+ENV DEVELOPMENT false
+ENV LOG_LEVEL INFO
```
pvalsecc: Faster and lighter (single layer):
```
ENV LOGO_CANTON=logo_canton.png \
    DEVELOPMENT=false \
    LOG_LEVEL=INFO
```

Makefile
```diff
@@ -168,21 +168,13 @@ test-db/12-create.sql: pyramid_oereb_standard.yml .venv/install-timestamp
...
-	docker run --rm --name $(DOCKER_CONTAINER_BASE)-print-dev \
-	    --publish=8280:8080 \
-	    $(DOCKER_BASE)-print-dev
-
 .PHONY: deploy
 deploy:
 	$(VENV_BIN)python setup.py sdist bdist_wheel upload
+
+logo_%.png: pyramid_oereb_standard.yml
+	touch --no-create $@
```
pvalsecc: I don't understand why this is needed

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM camptocamp/c2cwsgiutils:0
```
pvalsecc: You don't call the includeme?

Makefile
sbrunner: The rule who create `pyramid_oereb_standard.yml` also creates the required logos, and If I don't add a command it don't works :-)

Dockerfile
sbrunner: Just added :-)

Issue #234, opened 2017-09-06
PDF report name is hardcoded


ger-benjamin:
And should be configurable:
https://github.com/camptocamp/pyramid_oereb/blob/361e2e96973aa2c0c4d1cedb0dc61a5bcea4d148/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L113-L117

PR #236, opened 2017-09-07
Can configure the print template name to use

ger-benjamin:
Fix: #234


ger-benjamin:
Thanks @voisardf  @sbrunner

PR #238, opened 2017-09-07
Shapely update

voisardf:
Just updated the shapely wheels for the windows environnement


sbrunner:
This should also be updated...
https://github.com/camptocamp/pyramid_oereb/blob/master/dev-requirements-windows.txt#L2

ger-benjamin:
Have you tested it ? I means... I can't really review that :-)

voisardf:
Yes, I tested both versions on my machine :)
and yes, I'll update the requirement straight away

PR #247, opened 2017-09-15
added background layer to topic maps

### LINKED ISSUE #243: Background map on topicpages
In the PDF extract there is still the background layer (cadastral map) missing for the restriction topic maps. Normally it is the same that we use for the situation on the title page.

Could you have a look at that or give me instructions? thanks :)

voisardf:
Added the plan for land register as basemap to the topic map.

Optional extension: Generalize it to get all baselayers if extract_dict['baseLayers']['layers'] has more than one entry...

Fix #243

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -74,6 +74,7 @@ class Renderer(JsonRenderer):
                 'imageFormat': 'image/png',
             }]
         }
+        basemap = extract_dict['baseLayers']['layers'][0]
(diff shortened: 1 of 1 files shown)


pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -74,6 +74,7 @@ def __call__(self, value, system):
                 'imageFormat': 'image/png',
             }]
         }
+        basemap = extract_dict['baseLayers']['layers'][0]
```
sbrunner: Shouldn't we create this variable directly, and use it in the `baseLayers` building...?

voisardf: Yeah, that is fine with me and probably the better solution! Do you adapt it or do you want me to do it?

PR #249, opened 2017-09-20
The Host header should be forwarded

sbrunner:


voisardf:
Is that why I get this error on the productif instance for now? 
MaxRetryError("HTTPConnectionPool(host='print', port=8080): Max retries exceeded with url: /print/oereb/buildreport.pdf (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x07A42AF0>: Failed to establish a new connection: [Errno 11001] getaddrinfo failed',))",)

sbrunner:
Yes :-/

sbrunner:
New solution proposed :-)

sbrunner:
It's not needed in every cases, it's needed only when we use Apache vhost.
Then I should possible to add it, but I didn't see a good default value.

voisardf:
ok, thanks

sbrunner:
Thanks for the review :-)

PR #259, opened 2017-09-27
added opacity, styles and transparency params, reordered layers

voisardf:
Title says it all and should fix issue https://github.com/camptocamp/pyramid_oereb/issues/253 (at least partially).


voisardf:
merging

PR #262, opened 2017-09-29
Group the restrictions by theme

### LINKED ISSUE #254: multiple pages in case of multiple restrictions of same type
Actually, should only generate one page/theme per restriction type.

Needs print python webservice to aggregate results before sending JSON info to print

### LINKED ISSUE #263: "None given" not translated
The string here: https://github.com/camptocamp/pyramid_oereb/blob/9dfb92c17ca2dc29cd4be7dcbbf7be580937c4cd/print/print-apps/oereb/mainpage.jrxml#L352

Is hard coded in the jasper report. Should I translate it or anyway, if the value is not present we have a problem and I can replace it by "Null" ?

sbrunner:
fix #254
fix #263


pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -142,6 +142,34 @@ def __call__(self, value, system):
...
+                for element in text_element:
+                    restriction_on_landownership[element] = set([restriction_on_landownership[element]])
+                continue
+            curent = theme_restriction[theme]
+            # Number or array
+            for element in ['Area', 'Article', 'LegalProvisions', 'PartInPercent', 'Reference']:
+                curent[element] += restriction_on_landownership[element]
+            # Text
+            for element in text_element:
+                curent[element] += restriction_on_landownership[element]
```
voisardf: My instance says: TypeError: unsupported operand type(s) for +=: 'set' and 'unicode'

sbrunner: Should be fixed :-)

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -142,6 +142,34 @@ def __call__(self, value, system):
...
+            theme = restriction_on_landownership['Theme_Code']
+            if theme not in theme_restriction:
+                theme_restriction[theme] = restriction_on_landownership
+                for element in text_element:
+                    restriction_on_landownership[element] = set([restriction_on_landownership[element]])
+                continue
+            curent = theme_restriction[theme]
+            # Number or array
+            for element in ['Area', 'Article', 'LegalProvisions', 'PartInPercent', 'Reference']:
+                curent[element].add(restriction_on_landownership[element])
```
voisardf: Thanks, now it's Area creating an error
```File "c:\mapfish\pyramid_oereb\pyramid_oereb\contrib\print_proxy\mapfish_print.py", line 161, in __call__
    curent[element].add(restriction_on_landownership[element])
AttributeError: 'float' object has no attribute 'add'
```

sbrunner: Arg..., fixed :-)

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -142,6 +142,34 @@ def __call__(self, value, system):
...
+        for restriction_on_landownership in extract_dict.get('RealEstate_RestrictionOnLandownership', []):
+            theme = restriction_on_landownership['Theme_Code']
+            if theme not in theme_restriction:
+                theme_restriction[theme] = restriction_on_landownership
+                for element in text_element:
+                    restriction_on_landownership[element] = set([restriction_on_landownership[element]])
+                continue
+            curent = theme_restriction[theme]
+            # Number or array
+            for element in ['Area', 'Article', 'LegalProvisions', 'PartInPercent', 'Reference']:
```
voisardf: Area and PartInPercent may be None for linear and point restrictions.
In the case of linear restrictions, we should have a length value instead and nothing for a point restriction.

sbrunner: Than what should I do for linear restrictions, Do we have an other property?

voisardf: Yes, if it's a linear restriction, we do have 'length' I'm not sure about the 'PartInPercent':
https://github.com/camptocamp/pyramid_oereb/blob/361e2e96973aa2c0c4d1cedb0dc61a5bcea4d148/pyramid_oereb/lib/renderer/extract/json_.py#L242-L267
:)

sbrunner: Thanks, `Length` added :-)

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -142,6 +142,39 @@ def __call__(self, value, system):
...
+        text_element = [
+            'Information', 'Lawstatus_Code', 'Lawstatus_Text', 'ResponsibleOffice_Name',
+            'ResponsibleOffice_OfficeAtWeb', 'SymbolRef', 'TypeCode'
+        ]
+        for restriction_on_landownership in extract_dict.get('RealEstate_RestrictionOnLandownership', []):
+            theme = restriction_on_landownership['Theme_Code']
+            if theme not in theme_restriction:
+                theme_restriction[theme] = restriction_on_landownership
+                for element in text_element:
+                    restriction_on_landownership[element] = set([restriction_on_landownership[element]])
```
voisardf: Almost there :) I just get a  KeyError: 'SymbolRef' on line 156 now I haven't had before

voisardf: I'll check on a new branch if that's related to our new data or also happening with the sample data

sbrunner: Effectively, just fixed :-)

voisardf:
Thanks, the call 
goes through now! But there should be
- [ ] the surface and part for each type of restriction (page 3 atteched pdf)
- [ ] the length and the unit in m; no part indication in case of linear restrictions (page 4) - ForestPerimeters and ForestDistanceLines
- [ ] No indication for surface/length and part for point restriction

[20171002042237_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1349202/20171002042237_extract.pdf)
Knowing that in the LandUsePlans theme you could have the three types of restrictions...

sbrunner:
Thanks @voisardf for your tests I will work on it :-)

sbrunner:
@voisardf I do some new changes does it works ?

voisardf:
Sorry, I've got an key error again on line 187: 
restriction_on_landownership['Legend'].append(legend)

print/print-apps/oereb/Report_de.properties
```diff
@@ -28,3 +28,8 @@ CreationDateLabel = Erstellungsdatum
 CertificationLabel = Beglaubigung
 Certificationtext = Certification text
 TamponDateSignature = (Stempel)                                                                                   (Datum)                                                                         (Unterschrift)
+NoneGiven = None given
+Type = Typ
+Area = Fläche
+Length = Length
```
voisardf: Länge

sbrunner: Thanks :-)

sbrunner:
Should be fixed :-)

voisardf:
Thanks, the PDF gets created again. And for some topics I even get the legend symbol, for others I don't - I do not know yet if it's a problem with my DB entries or not.
[20171004032304_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1355947/20171004032304_extract.pdf)

There is a point I need to clarify with the others though: If we got multiple restrictions of the same kind - like 3x "zone de protection 2" we probably should also group and sum them IF they have the same publication date/law status - BUT that we can amend in an other PR

sbrunner:
Many thanks @voisardf :-)

PR #266, opened 2017-10-03
Some little fix found during OEREBlex integration

sbrunner:


voisardf:
I'm not using OEREBlex for our legal docs, so I can't tell if it's working or not... 😀

sbrunner:
It's working with custom code :-)

PR #272, opened 2017-10-06
Fix has_data value

kdeininger:
The parameter `has_data` signals, if the `EmptyPlrRecord` stands for a "not concerned theme" or a "theme without data".

It has to be set `True` in line 406. This means, the topic is available but doesn't intersect the real estate. So it's listed in the `not_concerned_theme` list.

Line 427 is for unavailable topics, listed in `theme_without_data`.


kdeininger:
@vvmruder ?

PR #274, opened 2017-10-06
Fix supported geometry types and unwrap multipart geometries

kdeininger:
The geometry type `MultiPolygon` should be available but is missing in the list of allowed types.

Do we need to add `MultiPoint` and `MultiLineString`, too? I'dont think, they are really necessary, but they are listed in the default `pyramid_oereb.yml` template. If we don't enable these types, they should be removed there, except the `GeometryCollection` which is handled specifically in the PLR source.

### DIFF
--- a/pyramid_oereb/lib/records/geometry.py
+++ b/pyramid_oereb/lib/records/geometry.py
@@ -2,7 +2,7 @@
 import logging
 from datetime import datetime
 from pyramid_oereb.lib.config import Config
-from shapely.geometry import Point, LineString, Polygon, MultiPolygon
+from shapely.geometry import Point, LineString, Polygon
 
 log = logging.getLogger('pyramid_oereb')
 
@@ -14,7 +14,7 @@ class GeometryRecord(object):
     Args:
         law_status (pyramid_oereb.lib.records.law_status.LawStatusRecord): The law status of this record.
         published_from (datetime.date): Date from/since when the PLR record is published.
-        geom (Point or LineString or Polygon or MultiPolygon):
+        geom (Point or LineString or Polygon):
             The geometry which must be of type POINT, LINESTRING or POLYGON, everything else
              will raise an error.
         geo_metadata (uri): The metadata.
@@ -32,7 +32,7 @@ class GeometryRecord(object):
         self.law_status = law_status
         self.published_from = published_from
         self.geo_metadata = geo_metadata
-        if isinstance(geom, (Point, LineString, Polygon, MultiPolygon)):
+        if isinstance(geom, (Point, LineString, Polygon)):
             self.geom = geom
         else:
             raise AttributeError(u'The passed geometry is not supported: {type}'.format(type=geom.type))
--- a/pyramid_oereb/lib/records/geometry.py
+++ b/pyramid_oereb/lib/records/geometry.py
@@ -2,7 +2,7 @@
 import logging
 from datetime import datetime
 from pyramid_oereb.lib.config import Config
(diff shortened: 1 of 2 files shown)


pyramid_oereb/lib/records/geometry.py
```diff
@@ -32,7 +32,7 @@ def __init__(
         self.law_status = law_status
         self.published_from = published_from
         self.geo_metadata = geo_metadata
-        if isinstance(geom, Point) or isinstance(geom, LineString) or isinstance(geom, Polygon):
+        if isinstance(geom, (Point, LineString, Polygon, MultiPolygon)):
```
sbrunner: We have nearly all the geom type, then shouldn't be `BaseGeometry`?

kdeininger: For me, it would be ok to use `BaseGeometry`.
@vvmruder and @voisardf, what do you think?

voisardf: Yes, MultiPolygon and MultiLineString are necessary - not sure about MultiPoint though. But in case of doubt, let's add that one too.
If I understand you right, you want to create a new BaseGeometry Class?

sbrunner: No I speak about using this class https://github.com/Toblerity/Shapely/blob/b23ffb63dbcc4c622de0e42cbcbaf056fe6272fa/shapely/geometry/base.py#L184 :-)

vvmruder: Since we have Point, Line, Polygon at this stage only we should clearly test for this. So -1 for general BaseType-Testing.

vvmruder:
@kdeininger: As we discussed, I would appreciate the handling of multi geometries at source level. Since the extract model does not support multi geometries directly but supports multiple geometries related to one public law restriction I would go the way to unwrap multi geometries to their single basic types (point, line, polygon) and assign them to the original public law restriction. This way we can avoid limitations of the extract model and support multi geometries we clearly have as data.

vvmruder:
@kdeininger: As we discussed, I would appreciate the handling of multi geometries at source level. Since the extract model does not support multi geometries directly but supports multiple geometries related to one public law restriction I would go the way to unwrap multi geometries to their single basic types (point, line, polygon) and assign them to the original public law restriction. This way we can avoid limitations of the extract model and support multi geometries we clearly have as data.

kdeininger:
@vvmruder, I've updated the source to unwrap the multi geometries and geometry collections. Could you have a look at it, please.

pyramid_oereb/lib/records/geometry.py
```diff
@@ -14,7 +14,7 @@ class GeometryRecord(object):
     Args:
         law_status (pyramid_oereb.lib.records.law_status.LawStatusRecord): The law status of this record.
         published_from (datetime.date): Date from/since when the PLR record is published.
-        geom (Point or LineString or Polygon or MultiPolygon):
+        geom (Point or LineString or Polygon):
```
voisardf: Isn't that confusing to remove MultiLineString and MultiPolygon here when adding MultPolygon support ? 😀

vvmruder: Indeed it is. But you know. We are related to the federal data model...

kdeininger: Fixed it :stuck_out_tongue_winking_eye:

kdeininger:
@sbrunner, did you already review the changes? Can I merge this PR?

PR #280, opened 2017-10-12
Encodingfix with special chars in legend

voisardf:
This PR solves encoding issues in the symbol call for an legend entry with special chars. For an reason unknown to me casting the legend_text dict to JSONType got us the utf-8 url escaped in the DB query...?!

### DIFF
--- a/pyramid_oereb/standard/hook_methods.py
+++ b/pyramid_oereb/standard/hook_methods.py
@@ -102,30 +102,14 @@ def get_symbol(request):
         raise HTTPBadRequest('Missing parameter TEXT.')
 
     try:
-        print 'START'
         model = DottedNameResolver().resolve('{module_}.{class_}'.format(
             module_=source_params.get('models'),
             class_='LegendEntry'
         ))
-        results = session.query(model).filter(
-            model.type_code == type_code
-        ).first()
-        print '*****'
-        print json.dumps(results.legend_text).decode('unicode-escape')
-        print type(json.dumps(results.legend_text).decode('unicode-escape'))
-        print '*****'
-        print '#####'
-        print  json.dumps(json.loads(base64.b64decode(legend_text))).decode('unicode-escape')
-        print type( json.dumps(json.loads(base64.b64decode(legend_text))).decode('unicode-escape'))
-        print '#####'
         legend_entry = session.query(model).filter(
             cast(model.type_code, Text) == cast(type_code, Text),
             cast(model.legend_text, Text) == json.dumps(json.loads(base64.b64decode(legend_text).decode('unicode-escape'))).decode('unicode-escape')
         ).first()
-        print '++++++'
-        print legend_entry.legend_text
-        print '++++++'
-        print 'INBETWEEN'
         if legend_entry:
             symbol = getattr(legend_entry, 'symbol', None)
             if symbol:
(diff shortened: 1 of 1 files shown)


sbrunner:
Looks a little hashish why so many cast...

sbrunner:
Why an encapsulated loads dumps ...

sbrunner:
Why the legend_text is encoded in base64?

voisardf:
Feel free to change... 😀 
I tried different encoding/decoding/cast for two days even with Michael and that is the only one that worked for me yet.
If you have a simple and proper solution, I take it 👍

pyramid_oereb/standard/hook_methods.py
```diff
@@ -107,9 +106,9 @@ def get_symbol(request):
             class_='LegendEntry'
         ))
         legend_entry = session.query(model).filter(
-            model.type_code == type_code,
-            cast(model.legend_text, Text) == cast(cast(
-                json.loads(base64.b64decode(legend_text).decode('utf-8')), JSONType), Text)
```
kdeininger: The problem is, that there's no comparison operator for the JSON datatype. So the `JSONType` attributes have to be converted to `Text` for filtering. That's basically the same as `json.dumps()`, but I'm not sure, if the ordering of the JSON elements is always identical. Maybe a JSON structure like

```
{
  "element1": "value1",
  "element2": "value2"
}
```

could be converted to `{"element1": "value1", "element2": "value2"}` by the one method, while the other method would return `{"element2": "value2", "element1": "value1"}`. This would cause the text comparison to fail, though the contents are the same. That's why I've decided to use the database casts on both sides.

But maybe I'm wrong and `json.dumps()` works exactly like casting a `JSONType` as `Text`.

voisardf: Problem is, that on my instance that creates issues with the encoding (python 2.7) ...
I'm open to any other solution getting me the symbols back and into the PDF... :)

sbrunner: And why not using jsonb postgres type?

voisardf: I'm absolutely not familiar with jsonb. I just found https://stackoverflow.com/questions/22654170/explanation-of-jsonb-introduced-by-postgresql
@kdeininger @vvmruder : any opinion on this?

voisardf:
OK guys, almost no pressure, but I'd really need an answer, review or counter proposition here. :)
In two weeks we all have to have things ready for the confederation and were not done with the PDF yet...

vvmruder:
For me this looks ok. But keep in mind that this strange comparison was only introduced because the fedaral data had wrong values for the type_code. That's why we couldn't  compare only on the type_code but needed also the legend_text.
As far as I know this malformed data will be fixed already this year. So the solution you made @voisardf  is a short term solution. but ok. If it is definitivly working for you I agree to merge it.

pyramid_oereb/standard/hook_methods.py
kdeininger: I think we should continue using the `json` type.

The comparison of the legend texts is just a workaround as there is not unique type code for each legend text in the federal data until now. I as far as I know, this issue should be solved with the switch to the new specification on December 1st. After that switch, the federal data is delivered corresponding to the new model and with unique type codes. Then the legend entries can be filtered using only the type code.

As mentioned above in the answer on stackoverflow, the type `json` should be preferred, if the JSON data is only stored and queried as it is. This should be the case after the federal data has been fixed.

sbrunner:
It don't looks good, but OK it it's only for short therm :-)

voisardf:
thanks for the feedback. I'll merge the PR.
@all: Shoud I open an issue in Jira or Github so we don't forget about fixing this workaround, onces the new specs are fully applied? :)

PR #284, opened 2017-10-17
enable encoding setting for pyconizer

vvmruder:
It turns out, that some WMS are not providing the correct encoding in the SLD response. So I introduced the encoding parameter to pass it to the pyconizer package for parsing the xml. That way we should be able to avoid errors in an elegant workaround :grimacing:


vvmruder:
@sbrunner: Will you merge this?

vvmruder:
gracie

sbrunner:
:-)

PR #287, opened 2017-10-20
Don't get multiple identical item

### LINKED ISSUE #286: Remove redundant OEREBlex links
With following example request, we get redundant OEREBlex links:
https://map-d.geo.sz.ch/oereb/wsgi/oereb/extract/reduced/pdf/SZ0200001372/4876?getspec=true

In my estimation there should be only following items:

- https://oereblex.sz.ch/api/attachments/577 (1 time instead of 3 times)
- https://oereblex.sz.ch/api/attachments/728 (1 time instead of 3 times)
- http://www.lexfind.ch/dtah/125640/2 (1 time instead of 6 times)
- http://www.lexfind.ch/dtah/125641/2 (1 time instead of 6 times)
- http://www.lexfind.ch/dtah/144693/2 (1 time instead of 6 times)
- http://www.lexfind.ch/dtah/147565/2 (1 time instead of 6 times)

According to François this can be done in the component called "mapfish_print_renderer" (or "mapfish_print_proxy"?)

The resulting PDF is attached.
[oereb_extract(18).pdf](https://github.com/camptocamp/pyramid_oereb/files/1394615/oereb_extract.18.pdf)

sbrunner:
fix #286 

@voisardf  does it looks good for you?

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -118,8 +118,8 @@ class Renderer(JsonRenderer):
                 self._multilingual_text(item, 'ResponsibleOffice_Name')
 
             legal_provisions = []
-            reference = []
-            article = []
+            references = {}
+            articles = {}
             if 'LegalProvisions' in restriction_on_landownership:
                 finish = False
                 while not finish:
@@ -128,23 +128,27 @@ class Renderer(JsonRenderer):
                         if 'Base64TextAtWeb' in legal_provision:
                             del legal_provision['Base64TextAtWeb']
                         if 'Reference' in legal_provision:
-                            reference += legal_provision['Reference']
+                            for reference in legal_provision['Reference']:
+                                references[reference['OfficialNumber']] = reference
                             del legal_provision['Reference']
                             finish = False
                         if 'Article' in legal_provision:
-                            article += legal_provision['Article']
+                            for article in legal_provision['Article']:
+                                articles[article['OfficialNumber']] = article
                             del legal_provision['Article']
                             finish = False
 
                 legal_provisions += restriction_on_landownership['LegalProvisions']
                 del restriction_on_landownership['LegalProvisions']
 
+            references = list(references.values())
+            articles = list(articles.values())
             self.lpra_flatten(legal_provisions)
-            self.lpra_flatten(reference)
-            self.lpra_flatten(article)
+            self.lpra_flatten(references)
+            self.lpra_flatten(articles)
             restriction_on_landownership['LegalProvisions'] = legal_provisions
-            restriction_on_landownership['Reference'] = reference
-            restriction_on_landownership['Article'] = article
+            restriction_on_landownership['Reference'] = references
+            restriction_on_landownership['Article'] = articles
 
         # One restriction entry per theme
         theme_restriction = {}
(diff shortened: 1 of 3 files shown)


pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -202,7 +206,20 @@ def __call__(self, value, system):
...
-        extract_dict['RealEstate_RestrictionOnLandownership'] = list(theme_restriction.values())
+        restrictions = list(theme_restriction.values())
+        for restriction in restrictions:
+            legends = {}
+            for legend in restriction['Legend']:
+                type = legend['TypeCode']
+                if type in legends:
+                    for item in ['Area', 'Length', 'PartInPercent']:
+                        if item in legend:
+                            legends[type][item] += legend[item]
```
voisardf: I get an `KeyError: 'Length'` here because there are legen[item'] = None coming in.

sbrunner: This mean that we can have tow element of the same `TypeCode`, with deferent geometry type?

voisardf: No, normally not. But I think there might be items where Area is set and Length defined as None? where the attribute Length should not exist

sbrunner: ok ...

sbrunner: Corrected, can you try again?

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -206,7 +210,20 @@ def __call__(self, value, system):
...
 
-        extract_dict['RealEstate_RestrictionOnLandownership'] = list(theme_restriction.values())
+        restrictions = list(theme_restriction.values())
+        for restriction in restrictions:
+            legends = {}
+            for legend in restriction['Legend']:
+                type_ = legend['TypeCode']
+                if type_ in legends:
+                    for item in ['Area', 'Length', 'PartInPercent']:
+                        if legend.get(item) is not None:
```
voisardf: should be either `type_` or `type` from L#217 down to L#223 in my opinion (and given my tests 😉)

and we have an other issue here. Somewhere on the way 'PartInPercent' is set to None instead of 0.0 when the area of a restriction is very small compared to the whole parcel: I have a case where a restriction area is 2.75m2 and the overall property is 158'596m2 >> None

So until we find where this assignement happens, I suggest this workaraound:
 ```
if legend.get(item) is not None:
    if legends[type_][item] is None:
        legends[type_][item] = 0.0
    legends[type_][item] += float(legend[item])
else:
    legends[type_] = legend
```

voisardf: that gives me this PDF: 
[20171024031649_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1410932/20171024031649_extract.pdf)

voisardf:
The legal provisions aspect seems to be fine: for the 7 contaminated sites of public transportation I get the seven individual references - no duplicates. 👍 
[20171024034230_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1411083/20171024034230_extract.pdf)

pyramid_oereb/contrib/print_proxy/mapfish_print.py
sbrunner: The "Distances légales à la forêt" don't have any values, does it have any value here?
https://github.com/camptocamp/pyramid_oereb/pull/287/files#diff-cfb7a95e0585c750deca6eabb5c5119bR167

sbrunner: I try to have a better fix for the 0 values :-)

voisardf:
Travis is unhappy :(

sbrunner:
Travis fixed :-)

sbrunner:
Thanks :-)

PR #302, opened 2017-11-07
[WIP] add more logging

jwkaltz:
Using name for all loggers. Adding some log statements.

### DIFF
--- a/pyramid_oereb/lib/readers/municipality.py
+++ b/pyramid_oereb/lib/readers/municipality.py
@@ -4,6 +4,7 @@ from pyramid.path import DottedNameResolver
 
 log = logging.getLogger(__name__)
 
+
 class MunicipalityReader(object):
     """
     The central reader for municipalities. It is directly bound to a so called source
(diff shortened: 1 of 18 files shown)


jwkaltz:
@sbrunner I would like to add more logging. Does is make sense for you like this?
I can then add some more.

pyramid_oereb/lib/readers/municipality.py
```diff
@@ -38,5 +41,7 @@ def read(self):
             The list of all found records. Since these are not filtered by any criteria the list simply
             contains all records delivered by the source.
         """
+        log.debug("read()")
```
ger-benjamin: Are these kind of log really useful ?

jwkaltz: Regarding PR #97 : you mean closing the PR 97 and instead I will suggest logging entries in error cases within this PR (302)? Yes, that would be ok for me.

Regarding your question "really useful": this one I guess probably not... except to have a trace (in debug Level) to understand where a request went through. What do you think about the traces with parameter values, would these be (more) useful?

jwkaltz:
Removing - interesting parts of this PR have now been addressed in other PRs.

PR #307, opened 2017-11-10
Remove mention of OpenLayers in CLA

jwkaltz:


cla/ccla.txt
```diff
@@ -19,7 +19,7 @@ If you have not already done so, please complete and sign the Agreement
 and mail a scanned copy of it to info < at > camptocamp.com.
 
 The signed Agreement may also be sent by facsimile if necessary. Please
-contact the OpenLayers developers for assistance if this is necessary.
+contact the Maintainer for assistance if this is necessary.
```
sbrunner: For me the capital letter isn't needed ... (=> maintainer)

jwkaltz: In a normal english sentence, yes -
but here, this is used as a special term (see above, "(the "Maintainer")"), likewise for "Contributions", "Agreement", etc.

sbrunner: It not so special for me but previously in the document the capital is also used, then merge :-)

PR #313, opened 2017-11-15
Fix the geom type detection

sbrunner:
When an element hasn't any Length or Area it should be None, not 0

Should fix https://github.com/camptocamp/pyramid_oereb/pull/311#issuecomment-344311324

### DIFF
--- a/pyramid_oereb/lib/records/plr.py
+++ b/pyramid_oereb/lib/records/plr.py
@@ -129,7 +129,7 @@ class PlrRecord(EmptyPlrRecord):
                 length = geometry.length
                 if length > self.min_length:
                     lengths_to_sum.append(length)
-        return sum(lengths_to_sum)
+        return sum(lengths_to_sum) if len(lengths_to_sum) > 0 else None
 
     def _sum_area(self):
         """
@@ -142,7 +142,7 @@ class PlrRecord(EmptyPlrRecord):
                 area = geometry.area
                 if area > self.min_area:
                     areas_to_sum.append(area)
-        return sum(areas_to_sum)
+        return sum(areas_to_sum) if len(areas_to_sum) > 0 else None
 
     @property
     def area(self):
@@ -169,7 +169,10 @@ class PlrRecord(EmptyPlrRecord):
                 tested_geometries.append(geometry)
                 inside = True
         self.geometries = tested_geometries
-        self._length = round(self._sum_length(), self.length_precision)
-        self._area = round(self._sum_area(), self.area_precision)
-        self.part_in_percent = round(((self._area / real_estate.limit.area) * 100), self.percentage_precision)
+        length = self._sum_length()
+        self._length = None if length is None else round(length, self.length_precision)
+        area = self._sum_area()
+        self._area = None if area is None else round(area, self.area_precision)
+        self.part_in_percent = None if area is None else \
+            round(((area / real_estate.limit.area) * 100), self.percentage_precision)
         return inside
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -246,7 +246,10 @@ class Renderer(JsonRenderer):
                 if type_ in legends:
                     for item in ['Area', 'Length', 'PartInPercent']:
                         if item in legend:
-                            legends[type_][item] += legend[item]
+                            if item in legends[type_]:
+                                legends[type_][item] += legend[item]
+                            else:
+                                legends[type_][item] = legend[item]
                 else:
                     legends[type_] = legend
             for legend in legends.values():
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -125,8 +125,8 @@ class Renderer(JsonRenderer):
                 self._multilingual_text(legend_item, 'LegendText')
 
             for legend_entry in restriction_on_landownership['OtherLegend']:
-                for element in legend_entry.keys():
-                    if element not in ['LegendText', 'SymbolRef', 'TypeCode']:
+                for element in ['LegendText', 'SymbolRef', 'TypeCode']:
+                    if element in legend_entry:
                         del legend_entry[element]
 
             del restriction_on_landownership['Map']  # /definitions/Map
@@ -177,8 +177,8 @@ class Renderer(JsonRenderer):
         for restriction_on_landownership in extract_dict.get('RealEstate_RestrictionOnLandownership', []):
             theme = restriction_on_landownership['Theme_Code']
             geom_type = \
-                'Area' if restriction_on_landownership.get('Area') is not None else \
-                'Length' if restriction_on_landownership.get('Length') is not None else 'Point'
+                'Area' if 'Area' in restriction_on_landownership else \
+                'Length' if 'Length' in restriction_on_landownership else 'Point'
 
             if theme not in theme_restriction:
                 current = dict(restriction_on_landownership)
@@ -245,7 +245,7 @@ class Renderer(JsonRenderer):
                 type_ = legend['TypeCode']
                 if type_ in legends:
                     for item in ['Area', 'Length', 'PartInPercent']:
-                        if legend.get(item) is not None:
+                        if item in legend:
                             legends[type_][item] += legend[item]
                 else:
                     legends[type_] = legend


sbrunner:
@voisardf can you test it?

pyramid_oereb/lib/records/plr.py
```diff
@@ -169,7 +169,10 @@ def calculate(self, real_estate):
                 tested_geometries.append(geometry)
                 inside = True
         self.geometries = tested_geometries
-        self._length = round(self._sum_length(), self.length_precision)
-        self._area = round(self._sum_area(), self.area_precision)
-        self.part_in_percent = round(((self._area / real_estate.limit.area) * 100), self.percentage_precision)
+        length = self._sum_length()
+        self._length = None ir length is None else round(length, self.length_precision)
```
voisardf: spelling error :)

sbrunner:
Just fixed :-)

voisardf:
Still get the NoneType error for this
`current['Legend'] =[{'PartInPercent': 99.9, 'Information': u'zone de vignes et gr\xe8ves', 'Area': 373.37, 'TypeCode': u'5202', 'SymbolRef': 'http://localhost:6543/oereb/image/symbol/LandUsePlans?TEXT=eyJmciI6ICJ6b25lIGRlIHZpZ25lcyBldCBnclx1MDBlOHZlcyJ9&CODE=5202', 'Geom_Type': 'Area'}, {'PartInPercent': 99.9, 'Information': u'zone agricole', 'Area': 373.37, 'TypeCode': u'2101', 'Length': None, 'SymbolRef': 'http://localhost:6543/oereb/image/symbol/LandUsePlans?TEXT=eyJmciI6ICJ6b25lIGFncmljb2xlIn0%3D&CODE=2101', 'Geom_Type': 'Area', 'TypeCodelist': None}, {'PartInPercent': 99.9, 'Information': u'zone viticole', 'Area': 373, 'TypeCode': u'2301', 'Length': None, 'SymbolRef': 'http://localhost:6543/oereb/image/symbol/LandUsePlans?TEXT=eyJmciI6ICJ6b25lIHZpdGljb2xlIn0%3D&CODE=2301', 'Geom_Type': 'Area', 'TypeCodelist': None}]`

```
2017-11-16 12:11:06,517 ERROR [pyramid_debugtoolbar:267][waitress] Uncaught Exception at http://localhost:6543/oereb/extract/reduced/pdf/121234?getspec=
traceback url: http://localhost:6543/_debug_toolbar/313136333133343234/exception
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid_debugtoolbar\toolbar.py", line 247, in toolbar_tween
    response = _handler(request)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid_debugtoolbar\panels\performance.py", line 72, in noresource_timer_handler
    result = handler(request)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\tweens.py", line 41, in excview_tween
    response = _error_handler(request, exc)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\tweens.py", line 16, in _error_handler
    reraise(*exc_info)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\tweens.py", line 39, in excview_tween
    response = handler(request)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\router.py", line 156, in handle_request
    view_name
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\view.py", line 642, in _call_view
    response = view_callable(context, request)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\viewderivers.py", line 390, in attr_view
    return view(context, request)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\viewderivers.py", line 368, in predicate_wrapper
    return view(context, request)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\viewderivers.py", line 410, in viewresult_to_response
    result = view(context, request)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\viewderivers.py", line 124, in _class_requestonly_view
    response = getattr(inst, attr)()
  File "c:\mapfish\pyramid_oereb\pyramid_oereb\views\webservice.py", line 226, in get_extract_by_id
    request=self._request
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\renderers.py", line 146, in render_to_response
    result = helper.render_to_response(value, None, request=request)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\renderers.py", line 458, in render_to_response
    result = self.render(value, system_values, request=request)
  File "c:\mapfish\pyramid_oereb\.venv\lib\site-packages\pyramid\renderers.py", line 454, in render
    result = renderer(value, system_values)
  File "c:\mapfish\pyramid_oereb\pyramid_oereb\contrib\print_proxy\mapfish_print.py", line 258, in __call__
    legend[item] = int(legend[item])
TypeError: int() argument must be a string or a number, not 'NoneType'
```

sbrunner:
I can't reproduce it, then an other question is coming what's the status of the sample data?
For me there is too many thing and we needs some more complex theme.
Is it OK If I completely remove the existing to replace by some more useful thing?
Any opposition @vvmruder @kdeininger?

vvmruder:
Sample data is really complex. I think it will grow to an amount of size where we will blow the repo... What do you think? @sbrunner @kdeininger

vvmruder:
By the way... Iam really wondering what was wrong with geometry type detection before? Wasn't it working in some way? Cause we couldn't find any error on that in the version we use :eyes:

sbrunner:
@voisardf do a test with polygon and lines in the same theme, I don't know if we have it in the sample data but I need it to debug it.

For me the sample data is to huge and don't offer good things to test the application.

vvmruder:
@sbrunner I think we do not will face this case... If I'am not wrong, this is not in the specification. But I might be wrong.

@kdeininger @voisardf ???

kdeininger:
The sample data is not used for the unit tests, as far as I know. So, for me it could be replaced or updated. I think, it's not necessary to provide the whole topic as sample data. Maybe it could be reduced to a smaller region. Data of additional topics can be added for this region too to cover more cases (geometry types, intersections, ...).

vvmruder:
But again Stephane: Is this related to a real existing bug or is it for code-clean-style-reason?

sbrunner:
It's a real issue that has @voisardf ...

vvmruder:
Ok. But this is print related. And I dont think this should be solved in the central PLR handling, cause this will affect the (working) extract itself directly. This is not good. Print related problem should be solved on the print side. Since we use the extract successfully with our client and have no problems summing up the amounts of lenghts and areas I think we DO NOT touch the PLR stuff...

I'am on it and will have a look. Ok?

sbrunner:
For me it an issue that when we don't have a line we have a length of 0, in this case we shouldn't have any length...

vvmruder:
I understand :)

But... Since we are already using this we need to plan a step like this. Its completely affecting the whole application. You agree with it?

sbrunner:
Except making those if working:
* https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/lib/renderer/extract/json_.py#L244
* https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/lib/renderer/extract/templates/xml/public_law_restriction.xml#L39

and the same for area, I don't see any effect in the application...

kdeininger:
According to the [specification](https://github.com/camptocamp/pyramid_oereb/blob/master/tests/resources/schema/draft/ExtractData.xsd#L397-L398), the attributes for length and area not mandatory. I think they are defined this way because polygons shouldn't have a length, lines shouldn't have an area and points should have none of these attributes.

So, for me it's OK to set these attributes to `None` if no values exist. The intersection and thresholds for each topic should ensure, that every public law restriction has a resulting length or area depending on its geometry type (except points).

vvmruder:
I talkred to @kdeininger. It should be fine to change this like @sbrunner introduced it. The specification will accept it.

vvmruder:
The errors mentioned by @voisardf are still existing?

vvmruder:
Regarding to short talk with @voisardf the problem with still exist. @sbrunner Can you stay on it to fix it with him?

sbrunner:
Then I should refactor the sample data to be able to reproduce the issue locally, is it OK that I do and commit that?

voisardf:
ok for me

sbrunner:
Excellent, tanks :-)

vvmruder:
For me too now :)

voisardf:
I think that the problem is that in fix2 we are missing the test for None type on line
https://github.com/camptocamp/pyramid_oereb/blob/fix2/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L250
while we have it in
https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L248

sbrunner:
I don't think so because we shouldn't have a None value here ...

sbrunner:
It's what I should investigate :-)

voisardf:
so with https://github.com/camptocamp/pyramid_oereb/blob/fix2/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L242-L259
adapted to
```
        for restriction in restrictions:
            legends = {}
            for legend in restriction['Legend']:
                type_ = legend['TypeCode']
                if type_ in legends:
                    for item in ['Area', 'Length', 'PartInPercent']:
                        if item in legend:
                            if item in legends[type_]:
                                if legend[item] is not None:
                                    legends[type_][item] += legend[item]
                            else:
                                legends[type_][item] = legend[item]
                else:
                    legends[type_] = legend
            for legend in legends.values():
                for item in ['Area', 'Length']:
                    if item in legend:
                        if legend[item] is not None:
                            legend[item] = int(legend[item])
            restriction['Legend'] = list(legends.values())
```
I get the specs and some pdf's again :)

sbrunner:
But it hide the real reason why we have a null Length attribute...

voisardf:
Is this one ready to merge?

sbrunner:
For me it can be merged :-)

voisardf:
ok, then let's go :)

PR #314, opened 2017-11-15
OEREBlexSource: Add checks for mandatory attributes and extend configuration

kdeininger:
The `OEREBlexSource` is extended

- to check if all mandatory attributes are set, because this isn't handled in the `geolink_formatter` library anymore (since version 1.1.2).
- with the default mapping for the relevant attributes of the new geoLink schema (20171108).
- with a configuration flag to put documents of the category `related` and the doctype `decree` into the list of main documents (directly to the public law restriction) instead of the list of referenced documents. This is useful to avoid referenced legal provisions being listed with the other legal documents (doctype `edict`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ install: $(PYTHON_VENV)
 	$(VENV_BIN)pip$(PYTHON_BIN_POSTFIX) install --requirement $(REQUIREMENTS)
 	touch $@
 
-.venv/install-timestamp: .venv/timestamp setup.py $(REQUIREMENTS)
+.venv/install-timestamp: .venv/timestamp setup.py requirements.txt
 	$(VENV_BIN)pip$(PYTHON_BIN_POSTFIX) install --requirement requirements.txt
 	$(VENV_BIN)pip$(PYTHON_BIN_POSTFIX) install --editable .
 	touch $@
--- a/pyramid_oereb/standard/pyramid_oereb.yml.mako
+++ b/pyramid_oereb/standard/pyramid_oereb.yml.mako
@@ -132,7 +132,13 @@ pyramid_oereb:
     canton: BL
     # Mapping for other optional attributes
     mapping:
-      municipality: subtype
+      official_number: number
+      abbreviation: abbreviation
+    # Handle related decree also as main document
+    # By default a related decree will be added as reference of the type "legal provision" to the main
+    # document. Set this flag to true, if you want the related decree to be added as additional legal
+    # provision directly to the public law restriction. This might have an impact on client side rendering.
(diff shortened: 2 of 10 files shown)


pyramid_oereb/standard/pyramid_oereb.yml.mako
```diff
@@ -131,7 +131,13 @@ pyramid_oereb:
     canton: BL
     # Mapping for other optional attributes
     mapping:
-      municipality: subtype
+      official_number: number
+      abbreviation: abbreviation
```
ger-benjamin: Only for my understanding, why do we need to map the `abbreviation` (and not the other) (For instnce, the title is automatically mapped, no ?)

kdeininger: All mandatory attributes are mapped automatically. For the optional attributes a mapping has to be defined, because their meaning might be different for each OEREBlex installation.

ger-benjamin: Ok, thanks :-)

ger-benjamin:
@kdeininger Can you also update geolink-formatter to 1.2.0 within this PR ?

kdeininger:
@ger-benjamin, could we set the version to `geolink-formatter>=1.1.2`? Version 1.2.0 won't work for the older geoLink schema. But we have to use the old schema until our next deploy in December because of other dependencies. This might affect other users too.

jwkaltz:
@kdeininger regarding old / new schema: this sounds like we should start a branch now, or soon. We will need it anyway to implement the new schema. I will discuss it today with @sbrunner

Makefile
```diff
@@ -54,7 +54,7 @@ install: $(PYTHON_VENV)
 	$(VENV_BIN)pip$(PYTHON_BIN_POSTFIX) install --requirement $(REQUIREMENTS)
 	touch $@
 
-.venv/install-timestamp: .venv/timestamp setup.py $(REQUIREMENTS)
+.venv/install-timestamp: .venv/timestamp setup.py requirements.txt
```
ger-benjamin: But we use $(REQUIREMENTS) everywhere in the makefile. Why do we use only requirements.txt here ?

sbrunner: Because we install the version from requirements.txt :-)

ger-benjamin: Oh, yes, sorry. Forget this comment :-)

kdeininger:
Thanks for your review! :-)
Is it OK for you to merge this PR now?

ger-benjamin:
For me, yes. @sbrunner ?

sbrunner:
+1

PR #321, opened 2017-11-17
add switch to model creation method for string or integer type primar…

vvmruder:
…y keys


vvmruder:
@sbrunner I got several issues on testing suddenly. mostly: 'E722 do not use bare except' in files i was not touching...

Was there something changed?

vvmruder:
Addition: It is about lint errors...

.gitignore
```diff
@@ -9,6 +9,7 @@ __pycache__/
 /.cache/
 /.coverage
 /coverage_report/
+build/
```
sbrunner: /build/

pyramid_oereb/standard/create_standard_models.py
```diff
@@ -39,12 +48,21 @@ def create_standard_model():
         type='string',
         help='The schema name. If not specified, it will be derived from the code.'
     )
+    parser.add_option(
+        '-s', '--primary_key_is_string',
+        dest='primary_key_is_string',
+        metavar='PRIMARY KEY IS STRING',
```
sbrunner: This should be in one word or separate with _ with space it usually mean that we have 4 parameters...

pyramid_oereb/standard/create_standard_models.py
```diff
@@ -9,6 +9,15 @@ def create_standard_model():
         description='Create a python file which contais all sqlalchemy models in to hold oereb '
                     'specification like data'
     )
+
+    def str2bool(v):
```
sbrunner: never used ...

pyramid_oereb/standard/create_standard_models.py
```diff
@@ -39,12 +48,21 @@ def create_standard_model():
...
         parser.error('No oereb code set.')
     if not options.geometry_type:
         parser.error('No geometry_type set.')
     if not options.target_path:
         parser.error('No target_path set.')
+    if options.primary_key_is_string:
+        primary_key_is_string = True
     _create_standard_configuration_models_py_(options.code, options.geometry_type, options.target_path,
-                                              options.schema)
+                                              options.schema, primary_key_is_string)
```
sbrunner: Why having a local variable we can directly use `str2bool(options.primary_key_is_string)`

pyramid_oereb/standard/create_standard_models.py
sbrunner: => will mean that's his usage is missing :-)

vvmruder: Ok :)) I understand :P

pyramid_oereb/standard/create_standard_models.py
vvmruder: Old leftover. I changed it.

.gitignore
vvmruder: changed it

pyramid_oereb/standard/create_standard_models.py
sbrunner: Argparse support boolean variable, you can add `action='store_true'` in the add_option, an directly use `options.primary_key_is_string` here :-)

PR #328, opened 2017-11-23
WIP Don't sum legend values for the same element

ger-benjamin:
Fix GEO-711

~~Before, some values was summed and we got x time the result of the real size for each legend element.~~

~~EDIT: perhaps not work with multiple items of the same kind -> wip~~

Real problem is that we get all geometries on one real estates. Then we make on plr for each geometry.
But after that, for each plr, we get all geometry related.
And so, we have all geometries in each plr (plr generated by all geometries)

### DIFF
--- a/pyramid_oereb/standard/sources/plr.py
+++ b/pyramid_oereb/standard/sources/plr.py
@@ -181,32 +181,27 @@ class DatabaseSource(BaseDatabaseSource, PlrBaseSource):
 
         return geometry_records
 
-    def from_db_to_geometry_records(self, geometries_from_db):
-        geometry_records = []
-        for geometry_from_db in geometries_from_db:
-
-            # Create law status record
-            law_status = LawStatusRecord.from_config(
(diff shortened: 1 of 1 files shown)


ger-benjamin:
@sbrunner 
Okay, the final question is: Why do we have one PLR by geometry. The result is multiple PLR with exactly the same content. According with the database model, we can have one PLR for multiple geometries. I think we must handle it like that. But for now, can I merge this PR and reduce the (sur)number of plr in another PR ?  (that can be long / difficult)

ger-benjamin:
In this PR, I have one geometry by PLR.
The result is a lot of identical plr with always one geometry. Then, in the print, we merge these plr by type and sum their "geometry" (so, the length, area, etc)

Now, I've another idea.
I think that the best will to have one PLR (by type) with an array of gemetries. I think that will be cleaner.

I close this PR and open another that will supersede this one with the new idea.

PR #329, opened 2017-11-28
Fix geom type, ...

sbrunner:
see commits


voisardf:
I still can't install shapely from an remote - I use wheels on windows because of this error:
Collecting shapely==1.6.2.post1 (from -r requirements.txt (line 7))
  Using cached Shapely-1.6.2.post1.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "c:\users\voisardf\appdata\local\temp\pip-build-51y80u\shapely\setup.py", line 80, in <module>
        from shapely._buildcfg import geos_version_string, geos_version, \
      File "shapely\_buildcfg.py", line 200, in <module>
        lgeos = CDLL("geos_c.dll")
      File "c:\applications\python27\Lib\ctypes\__init__.py", line 365, in __init__
        self._handle = _dlopen(self._name, mode)
    WindowsError: [Error 193] %1 n'est pas une application Win32 valide

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in c:\users\voisardf\appdata\local\temp\pip-build-51y80u\shapely\
make: *** [Makefile:58: .venv/install-timestamp] Error 1

is there a workaround?

and it seems shapely 1.6.2.post1 is not available as an whl yet?

sbrunner:
Manually change the version of Shaply in the `requirements.txt`, this version is not mendatory ...

voisardf:
I already did that ;0)
still it would be nice if the windows install could be done without manual interventions... but it's not high priority

sbrunner:
Isn't it possible to add automated check for that?

voisardf:
probably. we'll have a look at it once the system is running :)

Sorry, but no point elements are output in the json/pdf as it seems...
[20171129100957_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1513137/20171129100957_extract.pdf)

voisardf:
linear are ok, I just checked

sbrunner:
I just fix the points :-)

voisardf:
ok, I'll give it a try :)

voisardf:
no symbol in my pdf, but the line is there 👍 
[20171129054545_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1514641/20171129054545_extract.pdf)

sbrunner:
I don't speak about the symbol, but about the line in the legend,now we have a new line with objet protégé :-)

sbrunner:
I will investigate about the symbol :-)

voisardf:
it might just be that I have not created the icon yet... so if the others have it, don't worry about it

sbrunner:
OK :-)

sbrunner:
Then is it OK to be merged?

voisardf:
for me yes. if there is a problem with the symbol we'll fix it later :) 👍

sbrunner:
Thanks :-)

PR #330, opened 2017-11-28
first fix for wrong xml rendering

vvmruder:
The XML is malformed and need to be updated.

A big question about this is, on which version we want to rely on.

Compared to the service response provided by canton zürich I found some attributes which already included from the new specification.

These are:
PlanForLandRegisterMainPage in RealEstate
AreaShare/LengthShare in RestrictionOnLandownership

Any suggestions how to deal with it?


vvmruder:
My personal one is to stay on old version and plan a real switch later.

voisardf:
Hi, for the confederation v 0.4 is absolutely acceptable for now. So I vote for just make it compliant as we have to have it ready on this friday latest... 🙈 :)

vvmruder:
@voisardf Can you confirm, that this version is satisfying your needs?

voisardf:
I'll check this version and let you know!

voisardf:
Looks quite good to me. I did an extract on this base and sent it to Rolf Zuercher for validation. I'll give you the feedback once he gets back to me.
In the mean time I think we could merge this branch and open a new one, if there is bugfixing needed?

Thanks a lot! 👍

vvmruder:
Let us merge this when feedback from rolf is there. It should be enough to merge this evening or tomorrow.

voisardf:
where does the `<script/>` come from? I couldn't find it yet...
And when validating the XML against the schema it says that xsi:type is not valid either...

voisardf:
This one, is the one. Confirmed by the Confederation. Thanks a lot!! 😃

kdeininger:
Finally... :laughing:

The tests should succeed. So I think we can merge this PR when they have finished.

Still OK for you, @sbrunner and @ger-benjamin?

sbrunner:
Looks good to me :-)

PR #332, opened 2017-11-30
Be able to install the recommended dependencies

### LINKED ISSUE #331: need a flexible dependency mechanism
(issue born out of #267 )

pyramid_oereb lists its required libraries in requirements.txt. Currently, specific version numbers of the libraries are defined here. However, in pyramid_oereb's setup.py, the version numbers are then ignored. Consequences of this situation are:
* when you build an application that uses pyramid_oereb, the build will ensure that the libraries required by pyramid_oereb are installed, but you don't know in which version they will be installed. If you already have a required library, but actually a newer version would be required, it won't get installed. Vice-versa, if you don't have a required library, it will always be the newest version that gets installed, even if the newest version hasn't been tested with pyramid_oereb and might not work.
* note that the testing suite and the CI are not affected by this problem, because here pyramid_oereb's setup.py does not get executed. So the tests do actually get executed with specific version numbers.

work-around: manually define in your application's setup.py the required versions of the dependencies used by pyramid_oereb (use the same version numbers as listed in the requirements.txt of the version of the pyramid_oereb library that you are including).

goal: pyramid_oereb should provide a way to list the versions of the libraries it needs, as listed in requirements.txt. The build process for applications using pyramid_oereb can then take this information into account.

sbrunner:
Fix  #331

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -127,7 +127,11 @@ the ``-d`` or ``--dir=`` argument.
 .........................
 
 To include ``pyramid_oereb`` into your existing Pyramid application, you first have to include the plugin in
-your application's main method. Open the ``__init__.py`` of your main module and add the following statement
+your application's main method.
+
+.. note:: Use the name pyramid_oereb[recommand] to gets the recommanded versions of the dependencies.
+
+Open the ``__init__.py`` of your main module and add the following statement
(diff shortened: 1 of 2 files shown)


sbrunner:
No pip don't check the available version it it's not needed

sbrunner:
No pip don't check the available version it it's not needed

PR #333, opened 2017-11-30
WIP Use one PLR (by type) with multiple geometries

ger-benjamin:
Supersede #328 

If that work well, I'll comment that and we will be able to remove a lot of code in MapFishPrint.

### DIFF
--- a/pyramid_oereb/standard/sources/plr.py
+++ b/pyramid_oereb/standard/sources/plr.py
@@ -472,11 +472,14 @@ class DatabaseSource(BaseDatabaseSource, PlrBaseSource):
                     if len(geometry_results) == 0:
                         self.records = [EmptyPlrRecord(self._theme_record)]
                     else:
+                        relevant_plrs = {}
                         self.records = []
                         for geometry_result in geometry_results:
-                            self.records.append(
-                                self.from_db_to_plr_record(geometry_result.public_law_restriction)
-                            )
+                            plr = geometry_result.public_law_restriction
+                            plr_key = '{}_{}'.format(self._theme_record.code, plr.type_code)
+                            if plr_key not in relevant_plrs:
+                                relevant_plrs[plr_key] = plr
+                                self.records.append(self.from_db_to_plr_record(plr))
 
             finally:
                 session.close()


pyramid_oereb/standard/sources/plr.py
```diff
@@ -472,11 +472,14 @@ def read(self, real_estate, bbox):
...
                         self.records = [EmptyPlrRecord(self._theme_record)]
                     else:
+                        relevant_plrs = {}
                         self.records = []
                         for geometry_result in geometry_results:
-                            self.records.append(
-                                self.from_db_to_plr_record(geometry_result.public_law_restriction)
-                            )
+                            plr = geometry_result.public_law_restriction
+                            plr_key = '{}_{}'.format(self._theme_record.code, plr.type_code)
```
ger-benjamin: I'm note sure about that. Perhaps it's not safe but it's safe enough for the test.

kdeininger:
I think, grouping PLRs by type is not possible. There might be multiple PLRs of the same type on one real estate, but with different other attributes, like the responsible office. Also the referenced legal provisions will probably not be the same for all PLRs. You could aggregate these documents, but then you would loose the information about which PLR the document belongs to.

Apart from that, I think we would be breaking with the specification by grouping the PLRs.

Or did I misunderstand something, @ger-benjamin?

ger-benjamin:
Thanks for the comment.
Ok, I can imagine the problem (even if in my test data, I always have identical information for all PLR (for a same type)).

But I'm still confuse on this part of the code.
Here: https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/standard/sources/plr.py#L476 we loop on all geometry in the real estate.
But we don't use the geometry, we recollect all the geometry for this PLR here: https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/standard/sources/plr.py#L324

Perhaps I miss something but I'm sure that create duplicate information.
**If we can't regroup PLR, why PLR can have multiple geometries ?** (an array of geometry that potentially doesn't belong to the same PLR.) Perhaps this PR was a good idea: https://github.com/camptocamp/pyramid_oereb/pull/328 (read the code, it's more clear than my comments :-) even if I haven't updated the PLR record)

ger-benjamin:
@kdeininger , someone ?

kdeininger:
The first loop contains only the returned geometries of the intersection (all geometries which intersect the real estate). For these geometries we need to collect the concerned public law restrictions.

But as the relation between public law restrictions and geometries is 1:N, there might be other geometries belonging to the same restriction, but not intersecting the real estate. These geometries need to be delivered in the extract too. So we need the second iteration to collect all geometries of the concerned public law restrictions.

You could also say, the first one is just the spatial query whereas the second one collects the extract data.

ger-benjamin:
Thanks for the explanation. (It can be good to have some comments in the code to explains why we loop on geometry twice).

So Grouping PLR is not possible, but we have a bug with "duplicated" geometry. This PLR will resolve that: https://github.com/camptocamp/pyramid_oereb/pull/377
(and resolve schwyz issue: GEO-711)

PR #334, opened 2017-12-01
Update install process for Windows

kalbermattenm:
I updated some makefile stuff for Windows. Just some remarks concerning this PR:

* As I understand the whole makefile process, the dev-requirements are run whatever happens.
* I did not look into it deeper, but isn't there some kind of a loop in the install process defined in the makefile (it checked the requirements twice...)
* Shapely is a primary dependency. It comes from `requirements.txt` and not `dev-requirements.txt`, thus we should treat as this in Windows as well.
* Using `system-site-packages` is really not a good idea, even in Windows...


voisardf:
@sbrunner : could we put the shapely requirement to 1.6.2 without post1?
If yes, could you just fix the conflicts, then merge this branch?
cheers!

sbrunner:
Done, closing :-)

voisardf:
@sbrunner : did you report it somewhere else? or did you just miss the merging?  ;0)

sbrunner:
I directly commit it on master ...

sbrunner:
9159945d0fade42be0788291a89a815222b8a8b3 and ce16a5b11cf96cea5352f7f6dcb4a928757336a2

voisardf:
👍

PR #335, opened 2017-12-04
Fix extra can't redefined dependency version

sbrunner:


doc/source/installation.rst
```diff
@@ -126,10 +126,13 @@ the ``-d`` or ``--dir=`` argument.
 6. Include in application
 .........................
 
-To include ``pyramid_oereb`` into your existing Pyramid application, you first have to include the plugin in
-your application's main method.
+In your ``setup.py`` you shouldn't directly depends on ``pyramid_oereb`` because you will not depends on any
```
jwkaltz: I don't understand the second part of the sentence.

doc/source/installation.rst
```diff
@@ -126,10 +126,13 @@ the ``-d`` or ``--dir=`` argument.
...
+other package. You should use ``pyramid_oereb[recommend]`` to gets the recommended versions of the
```
jwkaltz: ... "to get the recommended versions of the dependencies, or ``pyramid_oereb[no-version]`` to get the dependencies without specifying which versions of these dependencies shall be used."

doc/source/installation.rst
sbrunner: ... what don't you understand?

jwkaltz: the "because you will not depends on any..." part.

doc/source/installation.rst
sbrunner: Done :-)

sbrunner:
Unfortunately no, if we do `pip install pyramid_oereb` we don't have any dependency, we should do 
`pip install pyramid_oereb[recommend]` or `pip install pyramid_oereb[no-version]` for me pip is not as flexible as I want ...

doc/source/installation.rst
jwkaltz: Can we inverse the sentence?

sbrunner: Fixed :-)

PR #340, opened 2017-12-07
Fix legend entry matching

kdeininger:
Because of an error in the federal data (no unique information for each type code), the matching of restrictions and legend entries is done by `type_code` **and** `information`/`legend_text`.

As the error in the data is fixed since the beginning of December, the matching can now be done by `type_code` only. This is how it is supposed to be according to the specification.

### DIFF
--- a/pyramid_oereb/routes.py
+++ b/pyramid_oereb/routes.py
@@ -12,18 +12,18 @@ def includeme(config):  # pragma: no cover
     """
 
     # Service for logo images
-    config.add_route('{0}/image/logo'.format(route_prefix), '/image/logo/{logo}.png')
+    config.add_route('{0}/image/logo'.format(route_prefix), '/image/logo/{logo}')
     config.add_view(Logo, attr='get_image', route_name='{0}/image/logo'.format(route_prefix),
                     request_method='GET')
 
     # Service for municipality images
-    config.add_route('{0}/image/municipality'.format(route_prefix), '/image/municipality/{fosnr}.png')
+    config.add_route('{0}/image/municipality'.format(route_prefix), '/image/municipality/{fosnr}')
     config.add_view(Municipality, attr='get_image', route_name='{0}/image/municipality'.format(route_prefix),
                     request_method='GET')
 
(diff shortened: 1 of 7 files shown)


pyramid_oereb/routes.py
```diff
@@ -12,18 +12,18 @@ def includeme(config):  # pragma: no cover
...
     # Service for municipality images
-    config.add_route('{0}/image/municipality'.format(route_prefix), '/image/municipality/{fosnr}')
+    config.add_route('{0}/image/municipality'.format(route_prefix), '/image/municipality/{fosnr}.png')
     config.add_view(Municipality, attr='get_image', route_name='{0}/image/municipality'.format(route_prefix),
                     request_method='GET')
 
     # Service for symbol images
     config.add_route('{0}/image/symbol'.format(route_prefix),
-                     '/image/symbol/{theme_code}')
+                     '/image/symbol/{theme_code}/{type_code}.png')
```
voisardf: I'm not sure if we should hardcode the image type or leave it configurable...

pyramid_oereb/standard/hook_methods.py
```diff
@@ -90,30 +89,20 @@ def get_symbol(request):
...
-                base64.b64decode(legend_text).decode('unicode-escape'))).decode('unicode-escape')
+            cast(model.type_code, Text) == cast(type_code, Text)
         ).first()
         if legend_entry:
             symbol = getattr(legend_entry, 'symbol', None)
             if symbol:
                 response = request.response
                 response.status_int = 200
-                response.content_type = 'image/*'
+                response.content_type = 'image/png'
```
voisardf: same as above :)

pyramid_oereb/routes.py
kdeininger: OK. According to the specification, symbol are allowed as PNG and SVG. I will remove the format extension.

voisardf:
👍

kdeininger:
@ger-benjamin, @sbrunner, OK for you?

PR #346, opened 2017-12-11
Integration issues

kdeininger:
Add improvements and fixes for legend entry creation.

**Legend entry creation**
- [X] Import cast method
- [X] Log exceptions with traceback

**Publish/unpublish themes**
- [X] Municipality not published => all themes in `themes_without_data`

### DIFF
--- a/pyramid_oereb/standard/load_legend_entries.py
+++ b/pyramid_oereb/standard/load_legend_entries.py
@@ -24,7 +24,7 @@ log = logging.getLogger('pyramid_oereb')
 
 def create_legend_entries_in_standard_db(config, topic_code, temp_creation_path='/tmp/pyconizer',
                                          language='de', section='pyramid_oereb', image_format='image/png',
-                                         image_height=36, image_width=72, encoding=None):
+                                         image_height=36, image_width=72, encoding=None, replace_host=None):
     """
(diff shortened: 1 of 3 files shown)


pyramid_oereb/standard/load_legend_entries.py
```diff
@@ -102,8 +106,11 @@ def create_legend_entries_in_standard_db(config, topic_code, temp_creation_path=
                     layer_config.get('layer') == params.get('LAYERS')[0]:
                 layer_existent = True
         if not layer_existent:
+            service_url = urlunsplit((url.scheme, url.netloc, '', '', ''))
+            if replace_host is not None:
+                service_url = replace_host
```
sbrunner: This can be also be done by:
```python
if replace_host is None:
  service_url = urlunsplit((url.scheme, url.netloc, '', '', ''))
else:
  service_url = replace_host
```
or shorter:
```python
service_url = urlunsplit((url.scheme, url.netloc, '', '', '')) \
   if replace_host is None else replace_host
```
With that we don't calculate `urlunsplit` if not needed :-)

kdeininger:
Sorry, we had to add some other fixes that we need for our deploy today. So we needed to use the same branch. I'll add the suggested improvement. Could you please review again? @sbrunner @ger-benjamin @voisardf

PR #354, opened 2017-12-19
PDF template improvements

voisardf:
First batch of template improvements. This PR should lead us to a final PDF answering various issues like
https://github.com/camptocamp/pyramid_oereb/issues/348
https://github.com/camptocamp/pyramid_oereb/issues/350
https://github.com/camptocamp/pyramid_oereb/issues/338
https://github.com/camptocamp/pyramid_oereb/issues/297

maybe more...


voisardf:
One of the annoying issues is, that Cadastra 8.5pt does not work well with Jasper and the PDF, beause it's an interpolated font size. We have discussed this point with the responsibles from the federal office and it's ok to use 9 or 8pt

voisardf:
Here is a commented PDF with the issues of this morning: 
[20171219103535_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1572448/20171219103535_extract.pdf)

and that is the state of this evening
[20171219053913_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1572450/20171219053913_extract.pdf)

more to come tomorrow

voisardf:
@sbrunner : thanks - A question though: I saw that version 3.11.3 of MFP is out. 
Should we update to this version? What are the changes?

sbrunner:
The main change it that we update most of the dependencies, and we requires Java 8.

I suspect that the requirement of Java 8 can actually be an issue, then I thing that we should stay on version 3.10.x.

voisardf:
ok, so we leave it at 3.10.x for now.

this PR is ready for reviewing and merging :)

PR #358, opened 2018-01-04
mapfish_print proxy: simplify language usage

jwkaltz:
Because the base class already has a _language, and because I could not find a reason for a second attribute about language, I propose that we remove the additional "lang".

This makes the code a bit easier, and also simplifies testing because objects can be reused more easily.

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -54,8 +54,8 @@ class Renderer(JsonRenderer):
         if value[1].images:
             raise HTTPBadRequest('With image is not allowed in the print')
 
-        # If language present in request, use that. Otherwise, keep language from base class
         self._request = self.get_request(system)
+        # If language present in request, use that. Otherwise, keep language from base class
         if 'lang' in self._request.GET:
             self._language = self._request.GET.get('lang')
 
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -54,13 +54,10 @@ class Renderer(JsonRenderer):
         if value[1].images:
             raise HTTPBadRequest('With image is not allowed in the print')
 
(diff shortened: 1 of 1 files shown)


jwkaltz:
Note that passing a parameter lang=fr, for example, in the current stand-alone installation will not work, independently of this pull request. I assume we should work on that, too, but that is a separate issue IMO.

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -54,13 +54,10 @@ def __call__(self, value, system):
...
 
+        # If language present in request, use that. Otherwise, keep language from base class
         self._request = self.get_request(system)
-
-        self.default_lang = Config.get('default_language')
         if 'lang' in self._request.GET:
-            self.lang = self._request.GET.get('lang')
-        else:
-            self.lang = self.default_lang
+            self._language = self._request.GET.get('lang')
```
kdeininger: I think `self._language` should still be set to the configured default language, if no language parameter has been defined in the URL.

```python
else:
    self._language = Config.get('default_language')
```

jwkaltz: This already happens in the base class:
https://github.com/camptocamp/pyramid_oereb/blob/26002ef4dfdb55cc0412988dcbb9365491bb180e/pyramid_oereb/lib/renderer/__init__.py#L25

jwkaltz:
@sbrunner 
default language is already set in the base class:
https://github.com/camptocamp/pyramid_oereb/blob/26002ef4dfdb55cc0412988dcbb9365491bb180e/pyramid_oereb/lib/renderer/__init__.py#L25

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -54,13 +54,10 @@ def __call__(self, value, system):
         if value[1].images:
             raise HTTPBadRequest('With image is not allowed in the print')
 
+        # If language present in request, use that. Otherwise, keep language from base class
```
sbrunner: missplaced

Issue #239, opened 2017-09-07
first PDF call does not work (NullPointerException or missing images)


voisardf:
After starting the sample environnement with make serve, then calling the PDF extract for the first time (full or reduced) with 
http://localhost:6543/oereb/extract/full/pdf/CH113928077734 (in my case)

I get a 
`Error while processing request:
java.lang.NullPointerException`
first. On the second try I get the PDF as expected

sbrunner:
I suspect that's an issue in mapfish print...

sbrunner:
Exception:
```
print_1  | 08:56:38.014 [ForkJoinPool-2-worker-3] DEBUG o.m.p.processor.ProcessorGraphNode - Succeeded in executing process: SetFeaturesProcessor
print_1  | 08:56:38.014 [ForkJoinPool-2-worker-23] DEBUG o.m.p.processor.ProcessorGraphNode - Succeeded in executing process: SetFeaturesProcessor
print_1  | 08:56:38.014 [ForkJoinPool-2-worker-23] INFO  o.m.p.processor.ProcessorGraphNode - Time taken to run processor: 'class org.mapfish.print.processor.map.SetFeaturesProcessor' was 130 ms
print_1  | 08:56:38.015 [ForkJoinPool-2-worker-3] INFO  o.m.p.processor.ProcessorGraphNode - Time taken to run processor: 'class org.mapfish.print.processor.map.SetFeaturesProcessor' was 130 ms
print_1  | 08:56:38.015 [ForkJoinPool-2-worker-23] DEBUG o.m.p.processor.ProcessorGraphNode - Executing process: CreateMapProcessor
print_1  | 08:56:38.016 [ForkJoinPool-2-worker-3] DEBUG o.m.p.processor.ProcessorGraphNode - Executing process: CreateMapProcessor
print_1  | 08:56:38.056 [ForkJoinPool-2-worker-23] ERROR o.m.p.processor.ProcessorGraphNode - Error while executing process: CreateMapProcessor
print_1  | java.lang.NullPointerException: null
print_1  |      at org.mapfish.print.processor.map.CreateMapProcessor.zoomToFeatures(CreateMapProcessor.java:581)
print_1  |      at org.mapfish.print.processor.map.CreateMapProcessor.execute(CreateMapProcessor.java:171)
print_1  |      at org.mapfish.print.processor.map.CreateMapProcessor.execute(CreateMapProcessor.java:104)
[...]

PR #359, opened 2018-01-09
Define an order for the Legend entries

jwkaltz:
Current, there is no order defined for the legend entries, so the order of these entries in the print spec depends on the execution environment.
This is not optimal because
-  the report for the identical extract varies according to the user environment, which is confusing for users
-  testing is hard: we can not do straight-forward tests, because we would need to program in the test case the assumption that the order should not matter for the legend entries. See also https://github.com/camptocamp/pyramid_oereb/issues/341.

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -258,7 +258,10 @@ class Renderer(JsonRenderer):
                 for item in ['Area', 'Length']:
                     if item in legend:
                         legend[item] = int(legend[item])
-            restriction['Legend'] = list(legends.values())
+            # After transformation, get the new legend entries, sorted by TypeCode
+            transformed_legend = list([value for (key, value) in sorted(legends.items())])
+            log.debug("transformed legend: {}".format(transformed_legend))
+            restriction['Legend'] = transformed_legend
 
         extract_dict['RealEstate_RestrictionOnLandownership'] = restrictions
         # End one restriction entry per theme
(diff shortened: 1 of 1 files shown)


pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -258,7 +258,10 @@ def __call__(self, value, system):
                 for item in ['Area', 'Length']:
                     if item in legend:
                         legend[item] = int(legend[item])
-            restriction['Legend'] = list(legends.values())
+            # After transformation, get the new legend entries, sorted by TypeCode
+            transformed_legend = list([value for (key, value) in sorted(legends.items())])
+            log.debug("transformed legend: {}".format(transformed_legend))
```
sbrunner: Debug message to be removed ...

voisardf:
On this first PDF extract, the map window zoom varies... why? ... no idea...
[20180109043732_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1615669/20180109043732_extract.pdf)

the second try of the same real estate got me the null pointer - again :(
```
Error while processing request:
java.lang.NullPointerException
```
then I printed another real estate and went back to the first and I got it twice the same way:
[20180109043916_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1615676/20180109043916_extract.pdf)
[20180109044059_extract.pdf](https://github.com/camptocamp/pyramid_oereb/files/1615678/20180109044059_extract.pdf)

strange, strange...

voisardf:
ok, the incomplete legend for polluted public transportation sites is a data issue - typecode is Staotyp2, so that's normal, but the zoom...

jwkaltz:
@voisardf I will try to work on the null pointer issue (#239) in the next days. On the stand-alone environment, the NPE is quite consistent, almost every time on first call, then almost never after that. That was like that before this PR, too.

jwkaltz:
@voisardf for the zoom problem, could you please open a separate issue?

PR #362, opened 2018-01-10
Print contrib test

jwkaltz:
First proposal for adding a test to the mapfish print conversion code.
To be able to test this code, some refactoring was done.

### DIFF
--- a/tests/contrib/print_proxy/test_mapfish_print.py
+++ b/tests/contrib/print_proxy/test_mapfish_print.py
@@ -1,0 +1,1470 @@
+# -*- coding: utf-8 -*-
+import pytest
+from pyramid_oereb.contrib.print_proxy.mapfish_print import Renderer
+from tests.renderer import DummyRenderInfo
+
+
+@pytest.fixture()
+def coordinates():
+    return [(((2608901.529, 1261990.655), (2608898.665, 1261991.598), (2608895.798, 1261992.53),
+              (2608892.928, 1261993.452), (2608890.054, 1261994.363), (2608880.256, 1261996.496),
+              (2608873.932, 1261998.379), (2608867.594, 1262000.213), (2608861.242, 1262001.998),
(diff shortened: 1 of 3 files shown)


voisardf:
thanks! 👍 
What do you think, should we get rid of all the 'complete' PDF related code as it's not requested anymore and planned to be removed as legal constraint for the PLR cataster?

jwkaltz:
@sbrunner commits are squashed now. I do wonder if there is a nicer way to handle the test data - but the new test is functional, and lint is happy too, so that is already a first step.

jwkaltz:
@voisardf ah yes, that is a good idea - but I think we should do that in a later step. Can you open an issue please?

tests/contrib/print_proxy/test_mapfish_print.py
```diff
@@ -0,0 +1,1470 @@
...
+    return symbol_ref_base() + 'ContaminatedPublicTransportSites/StaoTyp3'
+
+
+@pytest.fixture()
+def responsible_office_web():
+    return 'http://www.bav.admin.ch/themen/verkehrspolitik/00709/index.html'
+
+
+@pytest.fixture()
+def bav_pdf():
```
voisardf: I'd suggest to link a PDF that is sure not to be removed anytime soon like the certificat for no data for the railway topics of the PLR

voisardf:
issue created: https://github.com/camptocamp/pyramid_oereb/issues/364

tests/contrib/print_proxy/test_mapfish_print.py
jwkaltz: I see. But actually in this test case, it does not matter, better the test case supplies all the extract data directly (is just tests the way the extract is transformed for mapfish-print).
An additional test case, that actually generates an extract from the db and then calls the conversion code would be cool.
For now, I will merge this first test, because it involves restructuring the mapfish print code and so is a basis for further tests.

PR #365, opened 2018-01-15
update mapfish-print version

jwkaltz:


jwkaltz:
@voisardf between 3.10.4 and 3.10.6 the difference is two bug fixes for mapfish-print, without any library updates etc.. 
Java compatibility is an issue for the 3.11 versions (not yet planned for oereb).

voisardf:
👍

PR #370, opened 2018-01-16
new file for italian translations in report

voisardf:
I just created the initial file for the italian translation of the PDF and did some first translations.

@lopo977: to be completed with official terms fixed in Termdat (https://www.termdat.bk.admin.ch/Search/Search) and on https://www.cadastre.ch/it/manual-oereb.html


voisardf:
If that's fine with you I suggest to merge the file straight away without waiting for a complete translation so we have the file in the project and can work on it that way?

PR #373, opened 2018-01-17
complete italian translation from canton TI

voisardf:
title says it all.
Italian translation completed by canton TI


voisardf:
Thanks :)

PR #377, opened 2018-01-22
optimize loop (get plr details only once for each plr)

jwkaltz:


jwkaltz:
@vvmruder as done this morning - BUT the change makes one of the project tests fail. Can you please look at the test carefully? I did not want to just change the expected result of the test without you checking...

vvmruder:
We will have a look now on the broken test

vvmruder:
Was all the log stuff only for your debug @jwkaltz ?

jwkaltz:
@vvmruder actually IMO pyramid_oereb should generally be more verbose than it is now, when Log-Level is set to DEBUG or INFO.

For this specific new log statement (the one from the PR), I guess log.debug would be enough.

vvmruder:
Ok. It was only a little confusing while testing. It logged a lot then in the rest results. But I commented them temporarily.

vvmruder:
@jwkaltz I pushed a more convenient solution for this

Its already selecting the distinct on database. This prevents us from loading too many records to python to filter them distinctly in straight after query and decline them :)

What do you think? Can you test this in your SZ instance and provide results here?

jwkaltz:
@vvmruder in agreement with @ralphschulerSZ I have copied the proposed changes into the schwyz_oereb project, running here: https://map-d.geo.sz.ch/wkaltz_oereb/wsgi/oereb
For testing small parcels, I have been using SZ0200001372/1808
For big ones, SZ0200001372/2435

As far as I can tell, the new fix is working fine. It is hard to say what the impact on the performance is, the server is currently doing tile generation and therefore slower than usual.

jwkaltz:
@vvmruder @kdeininger as far as I'm concerned, we can apply this change now (i.e., merge the pull request). OK for you?

For your information, we did notice something strange with Ralph: when applying the first fix, there is a blank line within "Legend beteiligter Objekte" of "Grundwasserschutzzonen" in the report. The blank line is between "Grundwasserschutzzone S1" and "Grundwasserschutzzone S3". When applying the latest fix (commit by Clemens), this blank line disappears. As the phenomenon does not appear in the most recent version of this PR, I don't think it is worth investigating further.

What we also did notice is that the order within "Übrige Legende" and "Rechtsvorschriften" is not the same (the contents are the same, but the order of the entries is not). I assume that this is because in the mapfish-print proxy conversion code, we don't always impose an order, so the order is not guaranteed. I think we should change that. In PR359 we actually recently introduced an imposed ordering for the "Legend", I guess we should do something similar for "Übrige Legende" and "Rechtsvorschriften" - but that's a different issue (I will open one).

vvmruder:
Hi Wolfgang. Thanks for this check. @kdeininger and Iam glad that this works.

So you can merge this.

But I want to introduce to the fact that we have some strange behaviour about the spatial relations in special situations. For that please see #383. We are currently working on a fix there.

vvmruder:
:+1:

Issue #383, opened 2018-01-23
Spatial relation for calculation


kdeininger:
Until now, we use `intersects` to query the candidates and decide between inside and outside restrictions. This spatial relation also returns touching elements, but we don't need them for the extract. We even dismiss small intersections.

Maybe it would be better to use a combination of `overlaps` and `contains` to improve the processing performance.

vvmruder:
Good idea. This will be a good change for next version?

vvmruder:
It pointed out, that we had a missleading idea of using intersect when we used "intersect" for every geometry type.

See this example:
![grafik](https://user-images.githubusercontent.com/3714179/35321108-70e499ae-00e6-11e8-953b-52d815e59e42.png)


The parcel is intersecting with "Hecken, Feldgehölze" which is a polygon. Since it only shares points with the parcel the "intersects" operation is finding this. Later in the logic  the found element is really intersected with real estate. This gives a line as result. Thats why the symbology delivers a polygon symbol for a line of the length of 257m.

We will work on that to fix it.

vvmruder:
More input appreciated

vvmruder:
@jwkaltz can you confirm: I understood you like you think there should be a more detailed logging?

Especially for demystifying intersection and later following filter process I will include a more extensive logging.

vvmruder:
We will close this issue.

PR #390, opened 2018-01-24
Fix intersection

vvmruder:
We want to try to improve the intersection process in speed and decrease complexity.

#383


jwkaltz:
@vvmruder yes, I will try to find time tomorrow (it takes some time to adapt, we are still in copy-paste hell in that part)

vvmruder:
Ok. I only want to see if everything works before I merge this :)

jwkaltz:
Summary, as discussed: the contents are mostly identical with these changes.
The only difference is that after the change, "Nutzungsplanung kantonal (Siedlung)" and "Nutzungsplanung kantonal (Strassen)" appear in the TOC under "... welche das Grundstück nicht betreffen, whereas before they appear in "... zu denen noch keine Daten vorhanden sind".
It is not yet clear why there is an impact, and also not yet clear which is actually correct.

Another change is the order of the elements within "Übrige Legende", but this is an already known issue (currently, no sorting imposed here).

vvmruder:
I cross checked it now with our data. And we cant produce a behaviour like you explained @jwkaltz 

So I would merge this now.

jwkaltz:
Makes sense to me @vvmruder . If the order turns out to be incorrect in SZ reports, we will then analyze that more deeply. So merge is fine as far as I am concerned.

jwkaltz:
@vvmruder FYI the newest version (this one) is now in place at Schwyz (development environment), and good news: it turns out the new attribution within the TOC is correct! So the final version of this PR is indeed better than the intermediate version.

Issue #391, opened 2018-01-24
general information is missing in extract


vvmruder:
It turned out that "general_information" is missing in extract. This should be provided by pyramid_oereb configuration.

vvmruder:
@voisardf I will implement this as configurable text. Can you provide a translation to French later?

voisardf:
yes :)

voisardf:
Done.
Will be completed by https://github.com/camptocamp/pyramid_oereb/pull/394 and PR's https://github.com/camptocamp/pyramid_oereb/pull/396

@jwkaltz : You may close this issue once the PR's merged :)

voisardf:
Romanch is still missing, but the rest is there ;)

PR #393, opened 2018-01-25
add general information to extract, read it from configuration

vvmruder:
#391

### DIFF
--- a/pyramid_oereb/standard/pyramid_oereb.yml.mako
+++ b/pyramid_oereb/standard/pyramid_oereb.yml.mako
@@ -314,6 +314,12 @@ pyramid_oereb:
         methods:
           date: pyramid_oereb.standard.hook_methods.get_surveying_data_update_date
           provider:  pyramid_oereb.standard.hook_methods.get_surveying_data_provider
+    general_information:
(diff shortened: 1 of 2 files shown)


jwkaltz:
For me it is fine to do it like this.

FYI, I see that up to now, Schwyz has added this text to the config.yaml of their print configuration. Since this text is required only for the static extract, that way is acceptable too, I think - although that way is not convenient for multi-lingual text.

vvmruder:
@jwkaltz We need this text for all extract flavours.

pyramid_oereb/standard/pyramid_oereb.yml.mako
```diff
@@ -314,6 +314,12 @@ pyramid_oereb:
         methods:
           date: pyramid_oereb.standard.hook_methods.get_surveying_data_update_date
           provider:  pyramid_oereb.standard.hook_methods.get_surveying_data_provider
+    general_information:
+      text:
```
voisardf: This line has to been removed to get it working, otherwise the dict hirarchy does not work out for me...
I'm amending that in my french version, ok?

PR #395, opened 2018-01-25
Fix date in base data

kdeininger:
The update date for the cadastral data is only set for the first extract as the dictionary `base_data` gets updated. A new instance of `base_data` has to be created on each method call to fix this issue.


voisardf:
not sure I really got the issue? But the code is fine...

PR #397, opened 2018-01-25
Fix map bbox+buffer calculation

ger-benjamin:
Wait on the next release of MapFishPrint.

Will fix https://github.com/camptocamp/pyramid_oereb/issues/355

Note: I wonder if we can remove this config: https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/standard/pyramid_oereb.yml.mako#L79 I mean, we can calculate this information.

### DIFF
--- a/docker/config.yaml
+++ b/docker/config.yaml
@@ -68,8 +68,8 @@ vars:
       # extract. Here you can define the path to the logic which prepares the output as payload for print
       # service and returns the result to the user.
       renderer: pyramid_oereb.contrib.print_proxy.mapfish_print.Renderer
-      # The buffer on the map around the parcel in percent
-      buffer: 10
+      # The minimum buffer in pixel between the real estate and the map's border.
+      buffer: 30
       # The map size in pixel at 72 DPI (width, height), This is the defined size of a map image
       # (requested in wms urls) inside the static extract. On a pdf report, tha map size will
       # be calculated with the pdf_dpi and the pdf_map_size_millimeters below.
--- a/docker/config.yaml
+++ b/docker/config.yaml
@@ -68,7 +68,7 @@ vars:
       # extract. Here you can define the path to the logic which prepares the output as payload for print
       # service and returns the result to the user.
       renderer: pyramid_oereb.contrib.print_proxy.mapfish_print.Renderer
-      # The minimum buffer in pixel between the real estate and the map's border.
+      # The minimum buffer in pixel at 72 DPI between the real estate and the map's border.
       buffer: 30
       # The map size in pixel at 72 DPI (width, height), This is the defined size of a map image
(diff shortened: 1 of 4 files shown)


docker/config.yaml
```diff
@@ -68,8 +68,8 @@ vars:
       # extract. Here you can define the path to the logic which prepares the output as payload for print
       # service and returns the result to the user.
       renderer: pyramid_oereb.contrib.print_proxy.mapfish_print.Renderer
-      # The buffer on the map around the parcel in percent
-      buffer: 10
+      # The minimum buffer in pixel between the real estate and the map's border.
```
sbrunner: in pixel at 72 DPI

pyramid_oereb/standard/pyramid_oereb.yml.mako
```diff
@@ -67,8 +67,8 @@ pyramid_oereb:
     # extract. Here you can define the path to the logic which prepares the output as payload for print
     # service and returns the result to the user.
     renderer: pyramid_oereb.contrib.print_proxy.mapfish_print.Renderer
-    # The buffer on the map around the parcel in percent
-    buffer: 10
+    # The minimum buffer in pixel between the real estate and the map's border.
```
sbrunner: in pixel at 72 DPI

vvmruder:
Just a moment please: Since we have the dimensions in the specification in millimetres we should leave this config in mm. I think we can calculate the pixels dimension. Or do I have a missunderstanding here?

ger-benjamin:
For now, I've only changed the buffer for percent to pixels (at 72 DPI, like the map in pixel). Because the "buffer" (margin) in MapFishPrint is in pixels. Does it look acceptable for you ?

That all what'll do in this PR. But It's interesting to know why we have both the map dimension in pixel, and the map in mm. Because like you say, we are able to calculate the dimension of the map in the mm from pixels, or in pixels form mm.

ger-benjamin:
Look also: https://github.com/mapfish/mapfish-print/issues/651

ger-benjamin:
Now it work perfectly but you need the fix made for  mapfish/mapfish-print#651
The calculation of the min margin in MapFishPrint is also more logical.

vvmruder:
Just for my better understanding: We need to install new version of mapfish print for this?

IF: What version should it be?

ger-benjamin:
3.10.7 (not currently released) or  3.12.1 (already ready)

vvmruder:
Can we go further here? I downloaded 3.10.7 on our print instance.

ger-benjamin:
Yes, you can install the 3.10.7 (or better, the 3.2.1) then this PR should work (pyramid_oereb buffer = mapfishprint map's MinMargin).

PR #429, opened 2018-01-31
implement sorting for not_concerned_theme

jwkaltz:
Addresses https://github.com/camptocamp/pyramid_oereb/issues/428

Tested on stand-alone version, by forcibly simulating a case where ForestPerimeter gets moved from concerned to not_concerned. After the code changes, ForestPerimeter again appears at the expected position.


jwkaltz:
Sorry, I will look at codacy warnings first before asking for review.

pyramid_oereb/lib/processor.py
```diff
@@ -106,11 +106,45 @@ def is_inside_plr(theme_code):
...
+            new_not_concerned_theme = extract.concerned_theme.pop(idx)
+            log.debug("plr_tolerance_check() adding to not_concerned_theme theme: {}"
+                      .format(new_not_concerned_theme)
+                      )
+            extract.not_concerned_theme.append(new_not_concerned_theme)
 
         real_estate.public_law_restrictions = self.get_legend_entries(inside_plrs, outside_plrs)
         return extract
 
+    def order_extract_toc(self, extract):
```
vvmruder: just for convenience reason: shouldn't we call it "reorder_extract_toc"

pyramid_oereb/lib/processor.py
```diff
@@ -254,11 +288,14 @@ def process(self, real_estate, params, sld_url):
         for municipality in municipalities:
             if municipality.fosnr == real_estate.fosnr:
                 extract_raw = self._extract_reader_.read(real_estate, municipality, params)
-                # the selection of view services will be done whilst tolerance check. This enables us to take
+                extract = self.plr_tolerance_check(extract_raw)
+
+                self.order_extract_toc(extract)
```
vvmruder: is it really necessary to reorder TOC over all themes? I thought we have this problem only in the not concerned TOC.
Iam asking because this approach adds another iteration in iterations. This slows down response speed... And to me it seems better to hook in in the process at line 109. There we already know which themes we want to shift. So it should be easy to find their correct place in the array instead of simply appending it there.

pyramid_oereb/lib/records/theme.py
```diff
@@ -18,3 +18,8 @@ def __init__(self, code, text):
 
         self.code = code
         self.text = text
+
+    def __str__(self):
+        return '<%s -- code: %s text: %s>' % (
```
vvmruder: Iam more the `'hallo {0}'.format('du')` guy. Its a suggestion only ;)

pyramid_oereb/lib/processor.py
jwkaltz: Yes, and the name was misleading, too. It is only about the not_concerned part.

pyramid_oereb/lib/processor.py
jwkaltz: Actually the new code just iterates over the not_concerned part. The name was misleading, sorry.

Regarding where to hook it: I now moved the call into the process you mentioned. 
However, at that point there is no simple way the insert the themes in the right position, because the concerned_theme has just the indexes of the concerned_themes, the not_concerned_theme just the indexes of the not_concerned_themes, so we don't know in which position we should insert a theme we move from one list to the other. So we need to iterate, but this will iterate over only a very small number of elements. 
An alternative would be to remember more about the original ordering, in some data structure - for example, the ThemeRecord could remember its order in the whole list of themes, then we would need to iterate only once to insert a new element in the list.

pyramid_oereb/lib/records/theme.py
jwkaltz: I like that better too, as well -> changed.

pyramid_oereb/lib/processor.py
vvmruder: This is a good approach. For the moment we can leave it as is. But keep this idea in mind. Maybe we can add a "sort" attribute on the plr record which indicates a number. In the end this will stay valid through all processing and we can order lists after processing easily be this number.

pyramid_oereb/lib/records/theme.py
vvmruder: thanks ;)

PR #435, opened 2018-02-01
Be able to (easily) override get document_records methode in plr

ger-benjamin:
(For: GEO-941 - reduce code duplication)

I just moved out into its own function the logic to get document_records of a plr.

My final aim is to simplify the way to have oereblex documents in the plr. Before, for Schwyz, we replace almost the whole standard/sources/plr.py file only to have the oereblex documents (and that was hard to maintain). With this logic we can do it easily, just by inheritance and by overridden the "get_document_records" method. So,  like that in a project that want to use oereblex documents (and not more):

````
# -*- coding: utf-8 -*-                                                          
import base64                                                                    
import logging                                                                   
                                                                                 
from pyramid_oereb import Config                                                 
from pyramid_oereb.contrib.sources.document import OEREBlexSource                
from pyramid_oereb.lib.sources.plr import PlrBaseSource                          
                                                                                 
from pyramid_oereb.standard.sources.plr import DatabaseSource                    
                                                                                 
log = logging.getLogger(__name__)                                                                                                                         
                                                                                 
class DatabaseOEREBlexSource(DatabaseSource):                                    
    def __init__(self, **kwargs):                                                
        """                                                                      
        Keyword Arguments:                                                       
            name (str): The name. You are free to choose one.                    
            code (str): The official code. Regarding to the federal specifications.
            geometry_type (str): The geometry type. Possible are: POINT, POLYGON, LINESTRING,
                GEOMETRYCOLLECTION                                               
            thresholds (dict): The configuration of limits, units and precision which is used for processing.
            text (dict of str): The speaking title. It must be a dictionary containing language (as
                configured) as key and text as value.                            
            language (str): The language this public law restriction is originally shipped whith.
            federal (bool): Switch if it is a federal topic. This will be taken into account in processing
                steps.                                                           
            source (dict): The configuration dictionary of the public law restriction
            hooks (dict of str): The hook methods: get_symbol, get_symbol_ref. They have to be provided as
                dotted string for further use with dotted name resolver of pyramid package.
            law_status (dict of str): The multiple match configuration to provide more flexible use of the
                federal specified classifiers 'inForce' and 'runningModifications'.
        """                                                                      
        self._oereblex_source = OEREBlexSource(**Config.get_oereblex_config())   
        DatabaseSource.__init__(self, **kwargs)                                                                                                      
                                                                                 
    def get_document_records(self, public_law_restriction_from_db):
        """
        Override the parent get_document_records method to obtais the oereblex document instead. 
        """              
        return self.document_records_from_oereblex(public_law_restriction_from_db.geolink)                                                            
                                                                                 
    def document_records_from_oereblex(self, lexlink):                           
        """                                                                      
        Create document records parsed from the OEREBlex response with the specified geoLink ID and appends
        them to the current public law restriction.                              
                                                                                 
        Args:                                                                    
            lexlink (int): The ID of the geoLink to request the documents for.   
                                                                                 
        Returns:                                                                 
            list of pyramid_oereb.lib.records.documents.DocumentRecord:          
                The documents created from the parsed OEREBlex response.         
        """                                                                      
        log.warn("document_records_from_oereblex() start")                       
        self._oereblex_source.read(lexlink)                                      
        log.warn("document_records_from_oereblex() returning records")           
        return self._oereblex_source.records  
````

I've not passed a lot of time to do that, perhaps I miss something. What do you think @vvmruder ?

---

Then, a next point **may** to have the geolink column directly into the standard model. Do you think its 
feasible ?

### DIFF
--- a/pyramid_oereb/contrib/sources/plr_oereblex.py
+++ b/pyramid_oereb/contrib/sources/plr_oereblex.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+import logging
+
+from pyramid_oereb import Config
+from pyramid_oereb.contrib.sources.document import OEREBlexSource
+from pyramid_oereb.standard.sources.plr import DatabaseSource
+
+log = logging.getLogger(__name__)
+
+
+class DatabaseOEREBlexSource(DatabaseSource):
+    """
+    A source to get oereblex documents attached to public law restrictions in replacement
+    of standards documents. Be sure to use a model with an OEREBlex "lexlink" integer
+    column for plrs that use this source.
+    """
+    def __init__(self, **kwargs):
+        """
+        Keyword Arguments:
+            name (str): The name. You are free to choose one.
+            code (str): The official code. Regarding to the federal specifications.
+            geometry_type (str): The geometry type. Possible are: POINT, POLYGON, LINESTRING,
+                GEOMETRYCOLLECTION
+            thresholds (dict): The configuration of limits, units and precision which is used for processing.
+            text (dict of str): The speaking title. It must be a dictionary containing language (as
+                configured) as key and text as value.
+            language (str): The language this public law restriction is originally shipped whith.
+            federal (bool): Switch if it is a federal topic. This will be taken into account in processing
+                steps.
+            source (dict): The configuration dictionary of the public law restriction
+            hooks (dict of str): The hook methods: get_symbol, get_symbol_ref. They have to be provided as
+                dotted string for further use with dotted name resolver of pyramid package.
+            law_status (dict of str): The multiple match configuration to provide more flexible use of the
+                federal specified classifiers 'inForce' and 'runningModifications'.
+        """
+        super(DatabaseOEREBlexSource, self).__init__()
+        self._oereblex_source = OEREBlexSource(**Config.get_oereblex_config())
+
+    def get_document_records(self, public_law_restriction_from_db):
+        """
+        Override the parent's get_document_records method to obtain the oereblex document instead.
+        """
+        return self.document_records_from_oereblex(public_law_restriction_from_db.geolink)
+
+    def document_records_from_oereblex(self, lexlink):
+        """
+        Create document records parsed from the OEREBlex response with the specified geoLink ID and appends
+        them to the current public law restriction.
+
+        Args:
+            lexlink (int): The ID of the geoLink to request the documents for.
+
+        Returns:
+            list of pyramid_oereb.lib.records.documents.DocumentRecord:
+                The documents created from the parsed OEREBlex response.
+        """
+        log.warn("document_records_from_oereblex() start")
+        self._oereblex_source.read(lexlink)
+        log.warn("document_records_from_oereblex() returning records")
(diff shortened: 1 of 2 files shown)


sbrunner:
Shouldn't we add some documentation about OEREBlex integration?

vvmruder:
@ger-benjamin: This looks good in a first reading. I will check this in detail later that day.

vvmruder:
About the oereb-lex in standard database: This is not possible because we need this standard database structure to fit the federal minimal data model.

vvmruder:
For further development we should put your approach into the contrib section. This way its included in main package. Since we have a more or less generic API out of ÖREBLex I think this is arguable. If we do so, we can provide a mako file which is creating models files via console script. More or less the same way we went with the standard database models.

What do you think?

ger-benjamin:
> This is not possible because we need this standard database structure to fit the federal minimal data model.

Is this a real problem ? I mean, everybody is allowed to expand its database with specific column. We can add this column (useful for a lot of user, and if you don't need it, you can ignore it), that's not against the federal specifications. I know that's not the best for users that don't use it, but for this specific case, that seems acceptable to me.

> For further development we should put your approach into the contrib section. This way its included in main package. Since we have a more or less generic API out of ÖREBLex I think this is arguable.

I agree. I'll add this file in this PR.

> If we do so, we can provide a mako file which is creating models files via console script. More or less the same way we went with the standard database models.

That's a good idea, (even if that's more complicated to maintain that the "geonumber" column).
After this PR, I will create an issue "enhancement" to explain that (But currently, I don't have time / budget to do it in pyramid_oereb).

vvmruder:
Is this a real problem? Yes! We need to keep the overview here. The standard model is the mirror of the minimal federal data. If we change this we introduce eventualities here which are not necessary. Because this means we need to keep track of the documents. Minimal federal model has documents in database. We have at least 7 federal topics which are fitting exactly in this model. If you start now to add a conditional column e.g. "oereb_lex_link" in this models you need to make a mix interpetation and check if you want to use now the eventually existing documents in database or better interpret the attribute "oereb_lex_link" for reading all document related stuff from WebAPI.

For me it is important to keep the object oriented way. And this is not mixing up objects to later if...elif...elif...elif..elif...else them to get correct combination... Only splitting it in dedicated objects is the way to prosperity in this confusing and narrow-minded project ;)

ger-benjamin:
@sbrunner What is this error from codacy ? What can I do ?

sbrunner:
@jwkaltz what do you do with codacy? It seem that you configure a second instance for pyramid oereb witch is doing the mess ...

ger-benjamin:
@sbrunner thanks

@vvmruder (for models) Okay, I understand.

Then, Can I merge ? (I've added the contrib/source/plr_oereblex.py file since the firsts "approval" on this PR)

jwkaltz:
@sbrunner I am not aware of having done anything with codacy and/or a second pyramid_oereb instance. Any more infos?

sbrunner:
No, I will notify you if I get an issue again :-)

ger-benjamin:
Ok, so I merge and create related tickets.

Issue #388, opened 2018-01-24
unpredictable order on OtherLegend and LegalProvision


jwkaltz:
When printing an extract, there is no imposed order of the elements in OtherLegend ("Übrige Legende") and LegalProvision ("Rechtsvorschriften"). Therefore, the order can vary depending on the environment and the data selection implementation. This makes comparisons between / after difficult and also restricts test automation possibilities.

In #359 we introduced an imposed ordering for the "Legend", I guess we should do something similar for OtherLegend and LegalProvision.

jwkaltz:
Should we define an order (e.g. alphabetical according to "Text") for OtherLegend in the extract data already (before any transformation for printing)?
As far as I can tell, we don't have any ordering logic yet, and the specification does not say anything about that.
What do you think @voisardf @vvmruder @kdeininger ?

I guess this concerns not only OtherLegend, but other lists as well.

voisardf:
I'm generally in favor of an ordered display even if it is not mandatory 👍 thanks!

vvmruder:
I introduced the need of a sorting which is *necessary* in issue #380 

@kdeininger can you clarify where this sorting is coming from?

jwkaltz:
@vvmruder @kdeininger for the legend, I introduced a sorting in the mapfish print proxy, because there is actually logic rewriting which entries appear in the legend there, so there is an intermediate python data structure within the print conversion code. Right now, it is sorted by type code. But only for the legend.
AFAICT there is no sorting for any of the other lists, not in the extract preparation and not in the mapfish-print conversion. By the way, @ralphschulerSZ showed me the screenshot of the example in the specification, where the order appears random to me.

kdeininger:
I think, an alphabetical sorting of the elements in `OtherLegend` would be good improvement.

The list of affected public law restrictions should be sorted as described in #380. The sorting by `SubTheme` offers the possibility to group public law restrictions within a certain theme. For example, the topic `LandUsePlans` could be grouped into _basic land use_ and _overlaying land use_. The sorting could ensure, that the _basic land use_ is always shown first in the list, followed by the _overlaying land use_.

voisardf:
👍

jwkaltz:
Thanks for your feedback @kdeininger and @voisardf 
We will try to improve this for schwyz (and of course then do a PR here).

jwkaltz:
Because there have beem several issues about sorting, I would like to summarize the goal and the status of this issue:
[...]

PR #436, opened 2018-02-02
Ordering OtherLegends Alphabetically

marionb:
This should fix the ordering of the OtherlLegends described in #388.
(concerning the comment in multilingual_text.xml : it is something I noted while testing the code and I think it is good to keep in mind even though this is hopefully a very rare case)

Pleas review!

### DIFF
--- a/pyramid_oereb/lib/renderer/__init__.py
+++ b/pyramid_oereb/lib/renderer/__init__.py
@@ -159,9 +159,9 @@ class Base(object):
         Returns:
             new_text (string): with out umlaut
         """
-        new_text=text.lower()
-        new_text=new_text.replace(u'ä', u'a')
-        new_text=new_text.replace(u'ö', u'o')
-        new_text=new_text.replace(u'ü', u'u')
+        new_text = text.lower()
+        new_text = new_text.replace(u'ä', u'a')
+        new_text = new_text.replace(u'ö', u'o')
+        new_text = new_text.replace(u'ü', u'u')
 
         return new_text
--- a/pyramid_oereb/lib/renderer/__init__.py
+++ b/pyramid_oereb/lib/renderer/__init__.py
@@ -143,3 +143,25 @@ class Base(object):
             list of dict: List of dictionaries containing the multilingual representation.
         """
         return [self.get_localized_text(values)]
+
+    @staticmethod
+    def replace_de_umlaut(text):
+        """
+        Replaces all German special characters with the base character:
+        ä -> a
+        ö -> o
+        ü -> u
+        so that a proper sorting can be done according to the DIN 5007 norm
+
+        Args:
+            text (string): the string value
+
+        Returns:
+            new_text (string): with out umlaut
+        """
+        new_text=text.lower()
+        new_text=new_text.replace(u'ä', u'a')
+        new_text=new_text.replace(u'ö', u'o')
+        new_text=new_text.replace(u'ü', u'u')
+
+        return new_text
--- a/pyramid_oereb/lib/renderer/__init__.py
+++ b/pyramid_oereb/lib/renderer/__init__.py
@@ -168,10 +168,8 @@ class Base(object):
             element_list (pyramid_oereb.lib.records.view_service): The list of map.legends to sort.
 
         Returns:
-            element_list (pyramid_oereb.lib.records.view_service): Alphabetically and language
-                specific sorted elements if translations exist.
-            element_list (pyramid_oereb.lib.records.view_service): The list of unsorted map.legends
-                if sorting was not possible.
+            list of pyramid_oereb.lib.records.view_service: Alphabetically and language specific sorted
+                elements if translations exist or the list of unsorted map.elements if soring failed.
 
         """
         try:
--- a/pyramid_oereb/lib/renderer/__init__.py
+++ b/pyramid_oereb/lib/renderer/__init__.py
@@ -160,7 +160,7 @@ class Base(object):
         new_text = unicode(text.lower())
         return unicodedata.normalize('NFD', new_text)
 
-    def sort_elements_by_language(self, element_list):
+    def sort_by_localized_text(self, element_list):
         """
         Sort a list of translated text elements alphabeticaly.
 
@@ -173,10 +173,8 @@ class Base(object):
             (null): nothing is returned if the sort failed
 
         """
-        sorted_list = None
-
         try:
-            sorted_list = sorted(
+            return sorted(
                 element_list,
                 key=lambda text_element: self.unaccent_lower(
                     self.get_localized_text(text_element.legend_text)['Text']
@@ -185,5 +183,4 @@ class Base(object):
 
         except AttributeError as ex:
             log.warn('Other legend can not be sorted: {0}'.format(ex))
-
-        return sorted_list
+            return element_list
--- a/pyramid_oereb/lib/renderer/__init__.py
+++ b/pyramid_oereb/lib/renderer/__init__.py
@@ -162,18 +162,20 @@ class Base(object):
 
     def sort_by_localized_text(self, element_list):
         """
-        Sort a list of translated text elements alphabeticaly.
+        Sort a list of translated text elements alphabetically.
 
         Args:
-            element_list (pyramid_oereb.lib.records.view_service): The list of map.legends to sort accodring.
+            element_list (pyramid_oereb.lib.records.view_service): The list of map.legends to sort.
 
         Returns:
-            slegend_entry (pyramid_oereb.lib.records.view_service): Alphapetically and language
-                speciffic sorted elements.
-            (null): nothing is returned if the sort failed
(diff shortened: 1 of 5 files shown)


pyramid_oereb/lib/renderer/__init__.py
```diff
@@ -143,3 +143,25 @@ def get_multilingual_text(self, values):
             list of dict: List of dictionaries containing the multilingual representation.
         """
         return [self.get_localized_text(values)]
+
+    @staticmethod
+    def replace_de_umlaut(text):
```
sbrunner: => unaccent_lower

pyramid_oereb/lib/renderer/__init__.py
```diff
@@ -143,3 +143,25 @@ def get_multilingual_text(self, values):
...
+        Args:
+            text (string): the string value
+
+        Returns:
+            new_text (string): with out umlaut
+        """
+        new_text = text.lower()
+        new_text = new_text.replace(u'ä', u'a')
+        new_text = new_text.replace(u'ö', u'o')
+        new_text = new_text.replace(u'ü', u'u')
```
sbrunner: => 
```
unicodedata.normalize('NFD', 'Aéàè').encode('ascii', 'ignore')
```
To suppress all the accents ...

pyramid_oereb/lib/renderer/extract/templates/xml/view_service.xml
```diff
@@ -9,7 +9,13 @@
 <data:LegendAtWeb>${map.legend_at_web|u}</data:LegendAtWeb>
 %endif
 %if map.legends:
-%for legend_entry in map.legends:
+<%
+    def sorting_function(legend_entry):
```
sbrunner: Create this function out of the template an use the same as in the extract JSON ...

marionb:
@vvmruder I added the function in the rendering process because this is where all the rest of the logic is that is needed. Also the translations, which are ordered are fetched in the specific template (xml/json) itself.
If I understood the code correct it would require to make a lot of changes also to the templates (xml/json). For example where the translations are fetched and the the legend list is assembled.

vvmruder:
Ok thats true. I only dont like the fact doing the same stuff in two different places... I will have a more detailed look on that.

marionb:
Yes I understand I thought about that as well but I did not see any good way to do that and not take.

sbrunner:
And you nearly do tow times the same thing this should be re factored ...

ger-benjamin:
What about to create only one static function (`sort_element_by_language` ?) in https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/lib/renderer/__init__.py as discussed together ?

vvmruder:
@ger-benjamin : As you can see. I totally support your idea 👍

pyramid_oereb/lib/renderer/__init__.py
```diff
@@ -149,13 +149,41 @@ def get_multilingual_text(self, values):
...
+            text (str): The text value.
 
         Returns:
-            new_text (str): The text value converted to lower case and striped of special characters
+            new_text (str): The text value converted to lower case and striped of special characters.
         """
         new_text = unicode(text.lower())
         return unicodedata.normalize('NFD', new_text)
+
+    def sort_elements_by_language(self, element_list):
```
sbrunner: something like sort_by_localized_text

pyramid_oereb/lib/renderer/__init__.py
```diff
@@ -149,13 +149,41 @@ def get_multilingual_text(self, values):
...
+        try:
+            sorted_list = sorted(
+                element_list,
+                key=lambda text_element: self.unaccent_lower(
+                    self.get_localized_text(text_element.legend_text)['Text']
+                )
+            )
+
+        except AttributeError as ex:
+            log.warn('Other legend can not be sorted: {0}'.format(ex))
```
sbrunner: return element_list

pyramid_oereb/lib/renderer/__init__.py
```diff
@@ -149,13 +149,41 @@ def get_multilingual_text(self, values):
...
+        Args:
+            element_list (pyramid_oereb.lib.records.view_service): The list of map.legends to sort accodring.
+
+        Returns:
+            slegend_entry (pyramid_oereb.lib.records.view_service): Alphapetically and language
+                speciffic sorted elements.
+            (null): nothing is returned if the sort failed
+
+        """
+        sorted_list = None
```
sbrunner: unneded

pyramid_oereb/lib/renderer/__init__.py
```diff
@@ -149,13 +149,41 @@ def get_multilingual_text(self, values):
...
+
+        try:
+            sorted_list = sorted(
```
sbrunner: return sorted(

pyramid_oereb/lib/renderer/extract/json_.py
```diff
@@ -469,18 +469,13 @@ def format_map(self, map_):
...
-                self.format_legend_entry(legend_entry) for legend_entry in map_.legends]
+            other_legend = self.sort_elements_by_language(map_.legends)
 
-            try:
-                map_dict['OtherLegend'] = sorted(
-                    other_legend,
-                    key=lambda legend: self.unaccent_lower(legend['LegendText'][0]['Text'])
+            map_dict['OtherLegend'] = [
+                self.format_legend_entry(legend_entry) for legend_entry in (
+                    other_legend if other_legend else map_.legends
```
sbrunner: other_legend

pyramid_oereb/lib/renderer/extract/templates/xml/view_service.xml
```diff
@@ -10,10 +10,7 @@
 %endif
 %if map.legends:
 <%
-    sorted_legends = sorted(
-        map.legends,
-        key=lambda legend: unaccent_lower(multilingual(legend.legend_text)[0]['Text'])
-    )
+    sorted_legends = sort_elements_by_language(map.legends) if sort_elements_by_language(map.legends) else map.legends
```
sbrunner: sorted_legends = sort_elements_by_language(map.legends)

marionb:
@vvmruder I added the suggested refactoring, can you let me know if this works for you? thanks

pyramid_oereb/lib/renderer/__init__.py
```diff
@@ -143,3 +145,44 @@ def get_multilingual_text(self, values):
...
+
+    def sort_by_localized_text(self, element_list):
+        """
+        Sort a list of translated text elements alphabetically.
+
+        Args:
+            element_list (pyramid_oereb.lib.records.view_service): The list of map.legends to sort.
+
+        Returns:
+            element_list (pyramid_oereb.lib.records.view_service): Alphabetically and language
```
vvmruder: Combine doc in in one section like: "Alphabetically and language specific sorted elements if translations exist or the list of unsorted map.legends if sorting was not possible."

add type like: "element_list (list of pyramid_oereb.lib.records.view_service)"

pyramid_oereb/lib/renderer/extract/templates/xml/multilingual_text.xml
```diff
@@ -1,5 +1,7 @@
 <%page args="text"/>
 <%
+    ## TODO:this code gives an error when no Text attribute is available i.e when there is no translation
```
vvmruder: In which circumstances can it be, that no text attribute is available at this point of code?

vvmruder:
Sorry... Still have some questions/rfc's 😬

pyramid_oereb/lib/renderer/extract/templates/xml/multilingual_text.xml
marionb: If translation text is missing the xml gives an error if called in the browser (I think the function which tries to create the dictionary with the local text has a problem if there is no text value). I opened a new issue (#455) for this because for me this is not in the scope of this PR. However it is something that should be keep in mind.

vvmruder: But from the model it cant be that there is no text I think? @kdeininger What do you think?

vvmruder: Ok But You can close this. The other thing is more an enhancement

vvmruder:
Thanks for your work :)

pyramid_oereb/lib/renderer/extract/templates/xml/multilingual_text.xml
marionb: That I do not know, I just saw that there were errors when I was trying to test in French for which there were no translations available. I cloud build everything with out error but when calling the xml it crashed. For the json there are already some warnings built in:
https://github.com/camptocamp/pyramid_oereb/blob/77a5962c20d0e5c2630bb9abdc72c5bd3de7c837/pyramid_oereb/lib/renderer/extract/json_.py#L147
so there is no problem for the json.

PR #433, opened 2018-02-01
fixing graphic issues

voisardf:
this PR addresses various layout issues as alignements, spacing, display and overlapping of different elements.

It will cover problems declared in issues : 
- https://github.com/camptocamp/pyramid_oereb/issues/398
- https://github.com/camptocamp/pyramid_oereb/issues/416
- (https://github.com/camptocamp/pyramid_oereb/issues/413)

Issue #398, opened 2018-01-25
Mapfish Print: PDF extract analysis and critics


sbuettler:
We have analyced our PDF extracts so far, and i would like to share some critics regarding the template of the extract:

1. Position of the **legal base** information is placed under the wrong topic label.

![legal_base information](https://user-images.githubusercontent.com/29145003/35400519-e4f879d8-01ee-11e8-8971-c97dd073a142.jpg)

2. In the table of content are added under attachments the **legal bases** when you call for a full extract, and the title of the topic is not alligned with the page number:

![table of content_legale bases](https://user-images.githubusercontent.com/29145003/35400442-b613f0ca-01ee-11e8-8d48-e965d2b0d42b.jpg)

3. In the table of content are added under attachment the **legal prohibitions**, which aren't PDFs and therfore can't be attached:

![legal_prohibitions](https://user-images.githubusercontent.com/29145003/35400772-98b0320e-01ef-11e8-8f8c-1ae001f7e052.jpg)

4. Page number is wrong, when the table of content gets too long. If the table of content takes 2 pages the page number remains as if the the table of content takes only 1 page.

![page numbering](https://user-images.githubusercontent.com/29145003/35401105-880caa58-01f0-11e8-9fe1-1f83f5ebe900.jpg)

5. Timestamp in the footer is still written in the informatic kind of way :-)

![time stamp](https://user-images.githubusercontent.com/29145003/35401343-207fd760-01f1-11e8-822a-d4d41f4dd64d.jpg)

6. Symbol dimensions not 6x3mm (see instruction of CH). Is it possible to add automatically the black frame of 0.2pt bigness?

![symbol dimensions](https://user-images.githubusercontent.com/29145003/35401561-caedde18-01f1-11e8-8605-32b19a5fa542.jpg)

7. Missing stroke when there are no information in the information table (detail)

![stroke](https://user-images.githubusercontent.com/29145003/35401786-645b694e-01f2-11e8-9882-252b63f6aba2.jpg)

Nevertheless these critics very good work! Thanks!

voisardf:
Hi Sebastian, thanks for the feedback!

I'm aware of some of the issues others I thought fixed... What I'll gone do:

[ ] 1. For this point, in my opinion we have to change the DB structure a little and ad a table legal_bases the same as we have it for legal_provisions 
@vvmruder @sbrunner @jwkaltz Do you agree? Or do you see an other way to distinguish easely between legal bases and other reference document?
[...]

PR #437, opened 2018-02-05
set hints (other related documents) to a static "-" because its not p…

### LINKED ISSUE #407: Documents are presented under "Further informations and hints"
grafik

But they should be under "Rechtsgrundlagen".

"Weitere Infromationen und Hinweise" should be empty with a "-"

vvmruder:
…ossible to read it from extract version 1.0, switch articles and legal bases to have legal bases as "Rechtsgrundlagen"

fixes #407


vvmruder:
may I ask someone for approve here 😬

voisardf:
sorry, was on the phone... :(

print/print-apps/oereb/topicpage.jrxml
```diff
@@ -397,7 +396,7 @@
 				<subreportParameter name="Theme_Text">
 					<subreportParameterExpression><![CDATA[$F{Theme_Text}]]></subreportParameterExpression>
 				</subreportParameter>
-				<dataSourceExpression><![CDATA[$F{ArticleDataSource}]]></dataSourceExpression>
+				<dataSourceExpression><![CDATA[$F{ReferenceDataSource}]]></dataSourceExpression>
```
voisardf: Why does that change from ArticleDataSource to ReferenceDataSource? Did you have an old version?

vvmruder: Because we need to switch it. I think we have a misleading naming of variables here. As far as I can see this, it solves the problem which was adressed in #407 and #398 first point as well. This was not solved in your pull request #433

vvmruder: Is this answering your question? Or should I change something here?

voisardf: ok, I get it. thanks 👍

PR #444, opened 2018-02-06
update mfp version

jwkaltz:
Update to a mapfish-print version that integrates the fix for https://github.com/camptocamp/pyramid_oereb/issues/355


jwkaltz:
@ger-benjamin there is nothing more to do. But after this PR we can do stand-alone tests with the newest mfp out-of-the-box.

PR #453, opened 2018-02-09
Update Report_de.properties

### LINKED ISSUE #448: Change "Kasterführende Stelle" to "Katasterverantwortliche Stelle"

vvmruder:
fixes #448


vvmruder:
We decided to rely more to the definitions :) cause they are better described then the simple layout guidelines

PR #464, opened 2018-02-15
Update extract schema

kdeininger:
Downloaded current schema definitions from [cadastre.ch](https://www.cadastre.ch/de/manual-oereb.html) and moved them into a folder named by specification date.


kdeininger:
@sbrunner, can we disable code quality checks for the schema folders?

sbrunner:
Technically it's possible but this one looks to be a real error...

kdeininger:
OK, you're right. :upside_down_face: 
We'll fix this error and report it to swisstopo.

Issue #379, opened 2018-01-22
Rounding calculated values like area or percentage


vvmruder:
There is inconsistent part in specification. It is about the precision. For the extracts in JSON/XML format the specification clearly prescribes integer values. For the PDF format it prescribes 1 digit precision float for that values. This is one point to discuss.

In addition rounding values always introduce errors in summarized total amount. So it easily can happen that we have a parcel which points out 100.0 m2 and the elements which are on this parcel are summing up to 100.5 m2.

We need to find a way to do it in the central server part. And in my opinion we need to take these values "as they are delivered" in the rendering part AND in the clients (also in PDF). Otherwise we will get some chaos soon.

vvmruder:
Some suggestions on that?

sbrunner:
What do you mean by saying "as they are delivered"?
In fact I don't understand your last sentence...

For me to have coherent things the rounding should be in the rendering part, not in the core...

vvmruder:
thats true. We should do it in the rendering part. The problem we are facing here is the difference in the specification of the static extract which shows clearly a float with precision 1:
![grafik](https://user-images.githubusercontent.com/3714179/35510462-758a6e50-04f8-11e8-806a-a7390860b395.png)

But the specification for XML DataExtract says that fields which contains this are of type INTEGER...

To bring this to even more confusion everyone knows that javascript (and therefore JSON too) is not supporting pure integer values but the so called number type which is more a float then an integer.

My issue with this is what we wanna do to bring this in a practicable solution

vvmruder:
See #350 too for another branch of discussion

jwkaltz:
@vvmruder do we have a channel where we could validate our interpretation that the % should be float with precision 1 in the static extract?

The screenshot leaves various questions open (number precision respectively its inconsistency with the dataextract; sorting order). It would be good to get some clarification from the specification leaders.

vvmruder:
fixed with #466

Issue #458, opened 2018-02-13
About all the rounding of values


vvmruder:
I want to summarize what my idea/opinion is about what we have in the moment: Its an chaos...

After I had a detailed look on what goes on in the different places of the application I came to the point that we need to fix this centrally.

What am I talking about?
The calculated values in the extract for every PLR. We calculate the values inside the application. We rely on the configured settings for precision. We pass the values through the renderers to produce the DataExtract in whether JSON or XML. Later in clients like PDF/dynamic client we sum up the different values and ROUND again. This shouldn't be. We need to clean this up. The clients shouldnt change anything on the values. Only summarize them is ok.
Otherwise Iam afraid, we wont get any straight logic in this system :(

I reference on the issues: #379, #350 an PR #445

@voisardf @kdeininger @jwkaltz @sbrunner @ger-benjamin 
What are the other opinions?

- [x] fix int parsing in https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L354
- [x] Pass land registry area as integer
- [x] Add a comment that this is silly but will be fixed on new extract schema: https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/lib/renderer/extract/templates/xml/public_law_restriction.xml#L39

vvmruder:
Ok nobody seems to be interested here to solve this ;)

So I will start with some (in my mind) silly examples:

This line uses `int(<float>)` to parse a calculated float to an integer:
https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L354

REALLY BAD: int(99.9) parses to 99 😲

kdeininger:
The land registry area is passed as string to Mapfish Print:
https://github.com/camptocamp/pyramid_oereb/blob/d7b6dacd605b9979afc22950c4ebfbc217e4a9f8/print/print-apps/oereb/config.yaml#L68
https://github.com/camptocamp/pyramid_oereb/blob/d7b6dacd605b9979afc22950c4ebfbc217e4a9f8/print/print-apps/oereb/mainpage.jrxml#L33

This should be changed to integer.

ger-benjamin:
That probably simplistic, but from my point of view,
 1. Only the final renderer (JSON / XML) must round a value. Even if that cost a little bit more along the process.
[...]

PR #466, opened 2018-02-15
Fix rounding of length, area and part in percent

kdeininger:
- Round values in processor according to specification
- Remove unused precision configuration

Refs #458

### DIFF
--- a/print/print-apps/oereb/topiclegend.jrxml
+++ b/print/print-apps/oereb/topiclegend.jrxml
@@ -44,7 +44,7 @@
 				<textElement textAlignment="Right" verticalAlignment="Top" markup="html">
 					<font fontName="Cadastra" size="9"/>
 				</textElement>
-				<textFieldExpression><![CDATA[java.text.MessageFormat.format("{0} m<sup>2</sup>", java.text.NumberFormat.getInstance($P{REPORT_LOCALE}).format($F{Area}))]]></textFieldExpression>
+				<textFieldExpression><![CDATA[java.text.MessageFormat.format("{0} m<sup>2</sup>", String.valueOf($F{Area}))]]></textFieldExpression>
 			</textField>
 			<textField isBlankWhenNull="true">
 				<reportElement x="101" y="0" width="150" height="17" uuid="0f8b4f0d-278c-4906-b608-181ff2add08e">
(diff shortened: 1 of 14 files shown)


pyramid_oereb/lib/renderer/extract/templates/xml/public_law_restriction.xml
```diff
@@ -62,7 +64,7 @@
         <AreaUnit>${public_law_restriction.area_unit}</AreaUnit>
 %endif
 %if public_law_restriction.length:
-        <Length>${int(round(public_law_restriction.length))}</Length>
+        <Length>public_law_restriction.length}</Length>
```
sbrunner: missing `${`

pyramid_oereb/lib/records/plr.py
```diff
@@ -172,14 +161,21 @@ def calculate(self, real_estate):
...
         area = self._sum_area()
-        self._area = None if area is None else round(area, self.area_precision)
-        self.part_in_percent = None if area is None else \
-            round(((area / real_estate.limit.area) * 100), self.percentage_precision)
+        if area is None:
+            self._area = None
+            self.part_in_percent = None
+        else:
+            self._area = int(round(area, 0))
+            self.part_in_percent = round(((area / real_estate.land_registry_area) * 100), 1)
```
sbrunner: no `int()` ?

pyramid_oereb/lib/renderer/extract/templates/xml/public_law_restriction.xml
vvmruder: fixed it

pyramid_oereb/lib/records/plr.py
kdeininger: No, the specification defines "PartInPercent" as decimal number with one decimal place.
https://github.com/camptocamp/pyramid_oereb/issues/458#issuecomment-365928946

print/print-apps/oereb/topiclegend.jrxml
```diff
@@ -44,7 +44,7 @@
 				<textElement textAlignment="Right" verticalAlignment="Top" markup="html">
 					<font fontName="Cadastra" size="9"/>
 				</textElement>
-				<textFieldExpression><![CDATA[java.text.MessageFormat.format("{0} m<sup>2</sup>", java.text.NumberFormat.getInstance($P{REPORT_LOCALE}).format($F{Area}))]]></textFieldExpression>
+				<textFieldExpression><![CDATA[java.text.MessageFormat.format("{0} m<sup>2</sup>", String.valueOf($F{Area}))]]></textFieldExpression>
```
sbrunner: If we decide to do the rounding (the rendering) out of the template they should already be as string, here we have a special case with integer where we have no artifacts, but if we move to one decimal we can have some artifacts because number are coded in binary, not in decimal...

kdeininger:
@sbrunner, I have moved the formatting of area, length and percentage values into the print proxy. They are now passed to Mapfish Print as strings. Could you have another look at the changes, please?

vvmruder:
fixes #379

PR #487, opened 2018-03-14
implement correct sorting of themes on static extract

vvmruder:


jwkaltz:
I am not sure I understand. The themes are sorted correctly when given to the print proxy, right? Then they lose their sorting somewhere?

vvmruder:
This is exactly it. I think we should clean a little the print proxy code to a more readable piece later. For the moment this little bugfix is solving the problem we had with it (federal officers are really looking for correct sorting of themes...).

PR #497, opened 2018-03-29
Fix versions response

### LINKED ISSUE #494: Adapt getVersion specification version to 0.4
As discovered by JU, our getversion call specifies an 1.0.0 version support, but in reality we are delivering 0.4. @kdeininger will fix that. Thanks! :)

kdeininger:
Return the correct specification version (0.4) which is supported.

Fixes #494

### DIFF
--- a/pyramid_oereb/views/webservice.py
+++ b/pyramid_oereb/views/webservice.py
@@ -62,7 +62,7 @@ class PlrWebservice(object):
             u'GetVersionsResponse': {
                 u'supportedVersion': [
                     {
-                        u'version': u'1.0.0',
+                        u'version': u'0.4',
                         u'serviceEndpointBase': endpoint
                     }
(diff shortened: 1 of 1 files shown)


jwkaltz:
For my understanding: what is the understanding of version for versions response? 
I see that 0.4 is the version of the extract data model, but the extract definition per se is in version 0.8.

Is there yet another version number somewhere? Or is it the expectation that getVersions shall return the version of the extract data model (which I personally would find strange, because what we are really returning is the extract in version 0.8)?

voisardf:
@jwkaltz : You are right, the data-extract and the getVersions response should be identical. I didn't get we use the version 0.8 now 🙈

kdeininger:
Ok, it's 0.8 now :wink: Thanks!

PR #500, opened 2018-04-10

sdebayle:


ger-benjamin:
I've remove the commit that want to update "pycodestyle" to 2.4.0. It doesn't work with flake8 3.5.0 for now.
It should in the next version as a commit to fix the issue as been merged: https://gitlab.com/pycqa/flake8/merge_requests/230/commits

PR #503, opened 2018-04-18
add oereb lex model creation method for more convenient use

vvmruder:


pyramid_oereb/contrib/scripts/__init__.py
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
+import optparse
```
sbrunner: optparse is deprecated, you should use argparse...
see: https://docs.python.org/3/library/optparse.html

vvmruder:
I will add it when merged in the documentation branch.

vvmruder:
@sbrunner: I think we use this optparse for all console methods. Shouldn't we do an allround refactoring in a separate branch to clean this?

sbrunner:
> Shouldn't we do an allround refactoring in a separate branch to clean this?

Looks good :-)

PR #515, opened 2018-05-22

sdebayle:


jwkaltz:
@sbrunner any idea why my new requires filter is not working? Is the syntax not OK?

PR #522, opened 2018-06-11
fix bug for legend entries

### LINKED ISSUE #521: Legend entries not filtered by connected ViewService
When multiple view services are configured in a theme and connected to PLR's, the "OtherLegend" attribute in both XML and JSON contains all LegendEntries of the theme filtered correctly by BBOX. This is not correct. The LegendEntries in "OtherLegend" should always be only the ones "connected" to the dedicated ViewService.
pg_admin_screenshot

[CH228888020781_Giubiasco.json.txt](https://github.com/camptocamp/pyramid_oereb/files/2090282/CH228888020781_Giubiasco.json.txt)

vvmruder:
fixes #521

### DIFF
--- a/.codacy.yaml
+++ b/.codacy.yaml
@@ -1,0 +1,2 @@
+exclude_paths:
+  - 'tests/**'
--- a/pyramid_oereb/lib/processor.py
+++ b/pyramid_oereb/lib/processor.py
@@ -185,26 +185,35 @@ class Processor(object):
         # the legend entries.
         type_codes_to_remove = {}
         for inside_plr in inside_plrs:
+            view_service_id = str(inside_plr.view_service_id)
             theme_code = inside_plr.theme.code
             type_code = inside_plr.type_code
             if not type_codes_to_remove.get(theme_code):
-                type_codes_to_remove[theme_code] = {
-                    'codes_to_remove': [],
+                type_codes_to_remove[theme_code] = {}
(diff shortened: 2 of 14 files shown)


vvmruder:
Hey everybody. I need to introduce you to my (in some way) odd changes:

The cantons TI/BS pushed my nose a little deeper in the investigation why we should implement a filter for other legends. While thinking and talking I looked at the data model and it pointed out: No! There is no need for a filter. The model clearly shows, it must be possible to have only such legend entries in the "OtherLegend" of the extract which are belonging to specified PLR and therefore to one (and only one) ViewService. This fact was lost in our implementation completely. Why ever... I don't know. Maybe because we do not have this case in our data so far. Anyway.

Now I did a fix and it turned out: There is a lot of dependency to solve this problem in its completeness. You can clearly see in the changed files.

What I want to mention specially is: It was necessary to introduce the view_service_id to both record classes LegendEntryRecord/PLRRecord. This means the source must deal with it. I implemented it in the standard database source. So it should work without any problem. It does in our infrastructure. But please: Can someone test it too? There are a lot of changes even in central behaviour. And I simply want to be sure that this is not breaking anyones solution by now.

lopo977:
fix for TI issue works!

jwkaltz:
Thank you @vvmruder for the fix and @lopo977 for the testing

PR #523, opened 2018-06-12
Update plr.py

lopo977:
Semantic error legend_entries_from_db has no view_service_id attribute because is a list.
Do you mean legend_entry_from_db instead?


vvmruder:
Thanks for this PR. I saw this change before and fixed it already here #522 . So I think we can close this one. Do you agree?

lopo977:
I agree.
No problem.

PR #542, opened 2018-06-28
Add document_type to DocumentBaseRecord [GSOREB-232]

pfirpfel:
TODO:
 - [x] create a local json/xml/print export which includes the new attribute
 - [x] test xml output again
 - [x] check test/init_db.py

### DIFF
--- a/pyramid_oereb/standard/sources/plr.py
+++ b/pyramid_oereb/standard/sources/plr.py
@@ -271,6 +271,8 @@ class DatabaseSource(BaseDatabaseSource, PlrBaseSource):
                 )
             )
             document_records.append(self._documents_record_class(
+                # FIXME figure out which value to use
+                'LegalProvision',
                 law_status,
                 legal_provision.published_from,
                 legal_provision.title,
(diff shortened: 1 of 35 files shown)


pyramid_oereb/standard/sources/plr.py
```diff
@@ -271,6 +271,8 @@ def from_db_to_document_records(self, legal_provisions_from_db, article_numbers=
                 )
             )
             document_records.append(self._documents_record_class(
+                # FIXME figure out which value to use
```
pfirpfel: Check in print on how to figure this out.

pfirpfel: See: https://github.com/camptocamp/pyramid_oereb/blob/bc1101b4e100bdd764f9a4b83023eee41f0ec5c4/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L237-L266

I have the feeling that references and articles are always empty in the current print implementation...

pfirpfel:
@jwkaltz It should be fine now, the new attribute appears in the json extract and the json specs that generate the print.

PR #544, opened 2018-06-28
New restriction on landownership

marionb:
Fixes 1 and 2 of GSOREB-235


jwkaltz:
Print part will be https://jira.camptocamp.com/browse/GSOREB-243.
But I think it makes sense to do all model changes first before attacking the print stuff. The mapfish_print test (json conversion) will probably fail, though, until we do that too. This will be an expected error (until resolution of the print tickets).

PR #555, opened 2018-07-09
fix wrong scales (push fixes found for SZ to project default)

jwkaltz:
For https://github.com/camptocamp/pyramid_oereb/issues/535


vvmruder:
can be merged from my point of view

PR #556, opened 2018-07-10
Restriction on landownership part in percent nr of points

marionb:
Add points 3 and 4 of Jira issue: GSOREB-235

@vvmruder can you have a short look if the logic is what is expected for the new NrOfPoints attribute? Thanks

### DIFF
--- a/pyramid_oereb/lib/records/plr.py
+++ b/pyramid_oereb/lib/records/plr.py
@@ -105,7 +105,6 @@ class PlrRecord(EmptyPlrRecord):
             self.geometries = geometries
         self.info = info
         self.has_data = True
-        self.min_points = min_length
         self.min_length = min_length
         self.min_area = min_area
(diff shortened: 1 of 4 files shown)


pyramid_oereb/lib/records/plr.py
```diff
@@ -169,11 +190,17 @@ def calculate(self, real_estate):
                 tested_geometries.append(geometry)
                 inside = True
         self.geometries = tested_geometries
+
+        # Points
+        nrOfPoints = self._sum_points()
+        self._nrOfPoints = nrOfPoints if nrOfPoints else None
```
ger-benjamin: This if duplicate the one at line 155, no ?
Perhaps it's better to remove the one at line 155 (= `return points_to_sum`, so it returns 0 instead of "None", and this test give you "None")

PR #560, opened 2018-07-16
Add PlanForLandRegisterMainPage [GSOREB-236]

pfirpfel:
Allows a different map configuration for the main page.

### DIFF
--- a/pyramid_oereb/lib/readers/real_estate.py
+++ b/pyramid_oereb/lib/readers/real_estate.py
@@ -55,7 +55,8 @@ class RealEstateReader(object):
             list of pyramid_oereb.lib.records.real_estate.RealEstateRecord:
                 The list of all found records filtered by the passed criteria.
         """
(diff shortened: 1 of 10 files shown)


pyramid_oereb/lib/readers/real_estate.py
```diff
@@ -74,10 +74,22 @@ def read(self, nb_ident=None, number=None, egrid=None, geometry=None):
             max_NS03=max_NS03
         )
 
+        real_estate_main_page_view_service = ViewServiceRecord(
+            reference_wms=reference_wms,
+            legend_at_web=Config.get_real_estate_main_page_config().get('view_service').get('legend_at_web'),
```
ger-benjamin: Can you call this method (get_real_estate_main_page_config) only once ?

pfirpfel: Yes, that would make more sense!

PR #562, opened 2018-07-17
Use different configuration for main page [GSOREB-244]

pfirpfel:
Use different configuration for main page, opacity defaults as in directive

TODO: 
- [x] fix tests


jwkaltz:
what about the print template?

pfirpfel:
No need, this is the map parametrisation. If you run it out of the box now, the main page has 0.5 opactiy (as defined [here](https://github.com/camptocamp/pyramid_oereb/blob/a96e60bff5b660fe61399a57e31ac840e71f1633/docker/config.yaml#L194)), all other maps have 0.25 (as defined [here](https://github.com/camptocamp/pyramid_oereb/blob/a96e60bff5b660fe61399a57e31ac840e71f1633/docker/config.yaml#L187))

pfirpfel:
Are there more values that could be different besides the map?

PR #566, opened 2018-07-18
adapted requirements for windows

voisardf:
Changing shapely version is still an issue with Windows for now...
So please notify if you do so :)


requirements.txt
```diff
@@ -4,7 +4,7 @@ psycopg2==2.7.5
 pyramid==1.9.2
 pyramid-debugtoolbar==4.4
 PyYAML==3.13
-shapely==1.6.2  # rq.filter: ==1.6.2
+# shapely==1.6.4  # rq.filter: ==1.6.4
```
pfirpfel: Doesn't this break all the non-Windows versions?

voisardf: ups, yes it probably does, as I forgot to uncomment the new line. I'll fix that.

PR #581, opened 2018-07-24
Major reconstruction of behaviour around ViewServiceRecord

vvmruder:
While testing I had a more detailed look on the behaviour around the ViewService.

It turned out, that sanitize methods at ViewServiceRecord were setting standard values silently. This is a little bit of bad behaviour. I added a dedicated configuration for every theme in the standard_yaml and in addition for the real estate I changed the attribute names and structure of config to a more logic way (also using the correct attribute names for this from specification). To make the code more straight forward I moved all calculation methods from RealEstateReader to ViewServiceRecord.

Well this is my opinion. But I think its a little less scrambled now.

### DIFF
--- a/pyramid_oereb/lib/records/view_service.py
+++ b/pyramid_oereb/lib/records/view_service.py
@@ -1,11 +1,10 @@
 # -*- coding: utf-8 -*-
-import re
 import warnings
 import logging
 import requests
 
 from pyramid_oereb.lib.records.image import ImageRecord
-from pyramid_oereb.lib.url import add_url_params
+from pyramid_oereb.lib.url import add_url_params, parse_url
 from pyramid_oereb.lib.url import uri_validator
 from pyramid_oereb.lib.config import Config
 from shapely.geometry.point import Point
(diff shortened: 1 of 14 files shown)


vvmruder:
Just to clarify why...

AVOID hidden configuration. Things are complex enough.

vvmruder:
In addition: I changed behaviour of configuration for get_crs method. Since the federal specification wants to have an array of the available SRS's (nobody knows why) we formerly decided to let the method return an array. This is not really helpful because you cant configure more then one SRS in config. So I changed the method to return simply the SRS configured as unicode string and also changed the way it is used in the capabilities method to wrap it in a list there. This fixes also a little bug why I was stumbling over this all: In original code in RealEstateReader the comparison of the SRS's was done by `srs == u'...'` This was never true since srs was a list.

vvmruder:
And: No legend needed for plan_for_land_register NOR plan_for_land_register_main_page :grimacing:

vvmruder:
Sorry! It was necessary to fix some more bugs. But now it is ready to check :wink:

pyramid_oereb/lib/records/view_service.py
```diff
@@ -298,3 +284,27 @@ def unique_update_legends(self, legend):
...
+    def get_bbox_from_url(wms_url):
+        """
+        Parses wms url for BBOX parameter an returns these points as suitable values for ViewServiceRecord.
+        Args:
+            wms_url (str): wms url which includes a BBOX parameter to parse.
+
+        Returns:
+            set of two shapely.geometry.point.Point: min and max coordinates of bounding box.
+        """
+        url, params = parse_url(wms_url)
```
pfirpfel: Codacy complains about the unused variable 'url'.

What would be the nice way to solve this?

`__, params = parse_url(wms_url)`

or

`params = parse_url(wms_url)[1]`

PR #582, opened 2018-07-25
[wip] Validate extract against schema

pfirpfel:


vvmruder:
Be careful... Validating extract against schema produces always valid with every library we were trying until now. Even if extract is obviously wrong. This is because schema is not correct at all. So do not spend too much time on that. In our opinion first the schema should be fixed.

PR #584, opened 2018-07-26
PDF - Reassign documents to Laws, Hints, LegalProvision

### LINKED ISSUE #583: No correct determination of different document types in static extract
Since we have now the determination of Law, Hint and LegalProvision working fine in the DataExtract (JSON/XML) we need to transport this to the static extract as well. For the moment only LegalProvision's are recognized correctly. All others are gathered under the Law's in static extract. The Hint section is left blank for the moment.

See page 5/6 of this PDF:
[19b19c96-e7ee-4c24-b9b5-370ccd0427b4.pdf](https://github.com/camptocamp/pyramid_oereb/files/2227877/19b19c96-e7ee-4c24-b9b5-370ccd0427b4.pdf)

I would expect, that document "Wichtige Hinweise" will be delivered under "Weitere Informationen und Hinweise" since it is classified as "Hint" correctly in the corresponding JSON DataExtract:
grafik

ger-benjamin:
GSOREB-232
Fix #583


jwkaltz:
I am currently trying to extend the test data for the print test case (all documents must contain a document type, now), I will push when I got it working

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -431,6 +430,26 @@ def _flatten_object(parent, name):
...
+        """
+        uid = Renderer._get_element_of_legal_provision_maybe_uid(document)
+        documentType = document['DocumentType']
+        if documentType == 'LegalProvision':
+            legal_provisions[uid] = document
+        elif documentType == 'Law':
+            laws[uid] = document
+        else:
+            hints[uid] = document
+
```
pfirpfel: Should cover the case if there is no DocumentType and not default to hint.

jwkaltz: if there is no DocumentType, we now get an exception. The default case is only if type is present, but not LegalProvision or Law

pfirpfel:
@jwkaltz  I propose this variant of your "update expected mapfish-print json result"-commit without collateral damage to the unicode string:
https://github.com/camptocamp/pyramid_oereb/commit/8a8fb22c59113c537fe532fcc73d0b930f2a88a9

(exactly the same, but no unecessary replacements; passes tests)

vvmruder:
After all: From point of view this is OK now.

PR #597, opened 2018-08-09
Validate xml output against schema 

pfirpfel:


pfirpfel:
I extracted the sample data and extract fixes in case we don't want this test: #598

Issue #602, opened 2018-08-17
pyramid_oereb no longer buildable on Debian 8


jwkaltz:
pyramid_oereb is no longer buildable on old Debian systems (Debian 8 or older), due to library requirements:

`Downloading/unpacking cryptography>=1.3.4 (from urllib3[secure]==1.23->-r requirements.txt (line 12))
  Downloading cryptography-2.3.1.tar.gz (449kB): 449kB downloaded
  Running setup.py (path:/tmp/pip-build-Y4rmSm/cryptography/setup.py) egg_info for package cryptography
    Traceback (most recent call last):
      File "<string>", line 17, in <module>
      File "/tmp/pip-build-Y4rmSm/cryptography/setup.py", line 28, in <module>
        "cryptography requires setuptools 18.5 or newer, please upgrade to a "
    RuntimeError: cryptography requires setuptools 18.5 or newer, please upgrade to a newer version of setuptools
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):

  File "<string>", line 17, in <module>

  File "/tmp/pip-build-Y4rmSm/cryptography/setup.py", line 28, in <module>

    "cryptography requires setuptools 18.5 or newer, please upgrade to a "

RuntimeError: cryptography requires setuptools 18.5 or newer, please upgrade to a newer version of setuptools`

jwkaltz:
Fixed in #603 . Note that the fix concerns pyramid_oereb itself. Any pyramid applications including pyramid_oereb as a plugin may need some changes as well, if running on older environments.

PR #603, opened 2018-08-17
fix build for Debian 8 (and other old environments)

jwkaltz:
Since the usage of urllib3 introduced between versions 1.2.0 and 1.2.1, pyramid_oereb no longer builds on Debian 8 and other systems that have an old version of setuptools.
The cryptography lib version (required from urllib3) requires a recent version of setuptools.

This PR changes the build process, to force the usage of recent version of pip and setuptools.

See also issue #602 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,8 @@ else
     TESTS_SETUP_DB = tests-docker-setup-db
     TESTS_DROP_DB = tests-docker-drop-db
 endif
-
+PIP_UPDATE = $(VENV_BIN)pip install --upgrade pip setuptools
+ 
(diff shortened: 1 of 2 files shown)


vvmruder:
I tried it in a quick approach and had an error after clean-all + build. In the moment i have no more time to look deeper into it. Can it wait for 1-2 days?

vvmruder:
Sorry I clicked the wrong button. This shouldn't be closed. :grimacing:

jwkaltz:
@vvmruder it is not urgent, I will keep it open. I did try it on Debian 8, and did not get an error, so please do report your error when you have time.

PR #608, opened 2018-08-22
make it easier to build with python3

jwkaltz:


jwkaltz:
@ger-benjamin it depends on what is installed on the machine. If we say nothing, virtualenv help says "The default is the python2 interpreter on your path", at least in the current version of virtualenv, sometime in the future it will become python3 I guess.
If we say python2, it will use 2.7 if that is the version of python2 on the machine.

ger-benjamin:
Ok, thanks for the information. LGTM

Issue #585, opened 2018-07-27
shapely install on windows


voisardf:
We still have this issue that the shapely lib is not installed as it should under windows.
We always have to install it manually from the wheel...

Is there a simple way to fix this (less than an hour of work)? If so, could you do that? Otherwise we'll continue to do it manually as it does not take that long.

thanks!

jwkaltz:
@voisardf I am sorry, I am not aware of a simple way to fix this. Maybe some else in the user group, who uses Windows?

svareg:
Hi there,

I'm having exactly the same problem here.
It seems that there isn't any precompiled version of shapely for Windows on the pipy repository : https://pypi.org/simple/shapely/.

Therefore, when installing on Windows, it fallbacks to the tar.gz file, but this needs a C compilation.
Perhaps installing Microsoft Visual C on the build machine can help solving this problem ?

Personnaly, I also installed the Shapely wheel manually, it seems simpler.
But I couldn't find any wheel for the referenced version 1.6.4. I only found one for the 1.6.4.post1 version here : https://www.lfd.uci.edu/~gohlke/pythonlibs/#shapely, but this is not the right one and I have then an error when starting the pyramid application.

@voisardf Did you find a wheel for the 1.6.4 Version ? Or how did you do, so that the dependencies are met? Thanks!

voisardf:
@svareg : as far as I remember I "cheated": I installed the 1.6.4post1 version manually from the wheel and commented the line in the requirements file so it would not complain.

svareg:
Ow... I actually cheated too...
I've modified the 1.6.4.post1 wheel to create a fake 1.6.4 wheel...
But I'm not very proud of this solution :)

I discuss yesterday with @jwkaltz, and we saw that the version of shapely in the requirements.txt has been manually forced to 1.6.4. He'll investiguate the reason annd keep me informed. It would be nice to have it to 1.6.4.post1, so that we don't need to do such a trick to install pyramid_oereb.

Furthermore, during the installation I add problems to install another package on windows : pyproj.
I add to manually install it from the wheel too.
Did you have problem with this one as well?

jwkaltz:
Fixed in #615

PR #615, opened 2018-08-27
update shapely (use a version for which a windows wheel is available)

jwkaltz:
Addresses https://github.com/camptocamp/pyramid_oereb/issues/585


jwkaltz:
@voisardf If I understand the explanations of @svareg in #585 correctly, this should make building a pyramid_oereb application on Windows easier?

voisardf:
Hi, yes that would be helpful as for now I have to edit the requirements file to be able to install shapely manually afterwards without complaints from the app.

PR #622, opened 2018-08-30
Fix view service properties in JSON extract

kdeininger:
According to [extractdata.json](http://schemas.geo.admin.ch/V_D/OeREB/1.0/extractdata.json), the properties `layerIndex` and `layerOpacity` should start with lower case letters.


kdeininger:
@jwkaltz, good point!
I've applied the changed the opacity property in mapfish_print.py.

PR #623, opened 2018-08-30
Render real estate geometry only if requested

kdeininger:
This fixes the check, if geometries should be rendered for the real estate. As `self._params.geometry` is a boolean flag, it's never `None` and the geometry has always been rendered.


jwkaltz:
@kdeininger I just noticed that this breaks print conversion logic (if no geometry is output).
I will open an issue for that.

PR #627, opened 2018-09-04
Fix real estate limit error at mapfish print proxy

### LINKED ISSUE #626: [print] only use real estate geometry if available
Since https://github.com/camptocamp/pyramid_oereb/pull/623, a real estate geometry is not always available in the data.
The print conversion logic must be adapted to expect this (otherwise the print fails with  KeyError: 'RealEstate_Limit')

kdeininger:
Don't remove real estate limit as it doesn't get rendered anymore.

Fixes #626


kdeininger:
Successfully tested on my dev instance.

jwkaltz:
I guess we need to remove the RealEstate_Limit from the expected result (for the tests)

kdeininger:
Can we merge this PR to fix the next daily build?

PR #629, opened 2018-09-04
fix id

### LINKED ISSUE #628: Invalid XML
The XML is not completely valid on the min/max NS values because it's lacking the gml:id for the GML point and the `xlink:type="simple"` for the attributes itself.

vvmruder:
fixes #628


pfirpfel:
The format seems to be [valid](http://www.datypic.com/sc/niem21/e-gml32_Point.html) now.

Issue #459, opened 2018-02-14
transfer print into separate project


jwkaltz:
As discussed with @vvmruder, 
We should think about transferring the mapfish-print specific logic out of pyramid_oereb,
into an own project, pyramid_oereb_mfp

This would concern the mapfish-print templates and the webapp handling (the whole directory https://github.com/camptocamp/pyramid_oereb/tree/master/print)

The contrib part would stay in pyramid_oereb (https://github.com/camptocamp/pyramid_oereb/tree/master/pyramid_oereb/contrib)

vvmruder:
👍 ❤️ 

This I think will be a great effort. What will be the target solution? Simple git repo?

jwkaltz:
New repository for "print" files will be https://github.com/camptocamp/pyramid_oereb_mfp

sbrunner:
I think that a wrong approach to separate the print templates and the print proxy, there is too many interaction between the tow parts...

jwkaltz:
@vvmruder we discussed the different opinions with @sbrunner and @ger-benjamin . Our new proposal is:
- long-term goal is indeed to move mapfish-print templates to a separate project
- but currently, the logic in https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/contrib/print_proxy/mapfish_print.py is too close to the print template logic to be able to develop them independently one from another. To achieve the long-term goal without having this problem, we should wait on a mapfish-print release that includes the new jsonDataSource (https://github.com/mapfish/mapfish-print/pull/669) and refactor the print-proxy code to use this.
- so short-term handling of this issue would simply be to move the contents of the print directory to contrib/print_proxy, but not yet to an external project.

What do you think?

vvmruder:
I was never talking about moving the print proxy because this clearly belongs to the contribution part of the project. What we should move is the part about the templates and the jar file and the docker part for testing. The print proxy should only be tested about valid output.

In short: I agree with the offered approach. If the complete jasper, MapFishPrint, docker part is removed from this project.

PR #630, opened 2018-09-04
relocate print app

jwkaltz:
partially addresses #459


jwkaltz:
Yes but we do not have a consensus on that yet. So least common denominator right now would seem to at least move the templates into the contribs, because having a top-level "print" directory is confusing (?).

Just a suggestion :)

vvmruder:
Then I would leave it as it was. More confusing is to mix up things more. I still stay on the point that the jasper templates does not have anything to do with the pyramid_project. And therefore I do not want this stuff even deeper in the project.

@voisardf @kdeininger : What's your opinion on this?

jwkaltz:
OK, so I will close this (instead of "verschlimmbessering" things)

voisardf:
I'm late but I do agree :)

PR #631, opened 2018-09-05
provide an XML2PDF print service proxy

vvmruder:
This pull request contains changes to XML which are necessary to satisfy XML2PDF print service. This is only a first approach. If we will see too heavy changes we will implement a dedicated XML2PDF renderer to produce such XML as second step.


vvmruder:
with the fix https://github.com/camptocamp/pyramid_oereb/issues/628 and this changes I was able to generate a PDF
[0fc65a05-153e-4de7-806c-7742af465d84.pdf](https://github.com/camptocamp/pyramid_oereb/files/2352282/0fc65a05-153e-4de7-806c-7742af465d84.pdf)
There is still some problems to solve:

- GML is not valid in the opinion of XML2PDF service (therefore real estate highlight is not shown)
- The xlink thing is not necessary in my point of view
- XML2PDF handles our approach of 'mixable' languages in an extract not properly (that's why some content is lacking).

I'am in contact with GISDATEN AG to find a solution on that.

pramisters:
- According to GIS Daten AG (20.03.2019)  they will solve the xlink issue with the next issue of XML2PDF.

- As XML2PDF expects decoded URLs pyramid_oereb should not encode URLs when a XML dataextract is requested #820

vvmruder:
@jwkaltz This sounds good. I planned to work on that today. So I will rework it and push again. Than it can be merged.

vvmruder:
@jwkaltz It was coming into my mind that we had an issue with encoded URL's in XML. Is this solved too?

jwkaltz:
> @jwkaltz It was coming into my mind that we had an issue with encoded URL's in XML. Is this solved too?

It seems so, yes. Since we used your code without changes, but did not change the XML templates, I am assuming XML2PDF changed something and the encoding is no longer a problem.

pramisters:
No, XML2PDF did NOT change anything concerning encoding. I just tested it again with a unchanged XML from pyramid_oereb. It still doesnt't work.
But, anyway, if the XML2PDF proxy works together with XML2PDF, it's fine for me.

vvmruder:
> > @jwkaltz It was coming into my mind that we had an issue with encoded URL's in XML. Is this solved too?
> 
> It seems so, yes. Since we used your code without changes, but did not change the XML templates, I am assuming XML2PDF changed something and the encoding is no longer a problem.

So I will remove changes on XML. And commit again.

pramisters:
> No, XML2PDF did NOT change anything concerning encoding. I just tested it again with a unchanged XML from pyramid_oereb. It still doesnt't work.
> But, anyway, if the XML2PDF proxy works together with XML2PDF, it's fine for me.



> > @jwkaltz It was coming into my mind that we had an issue with encoded URL's in XML. Is this solved too?
> 
> It seems so, yes. Since we used your code without changes, but did not change the XML templates, I am assuming XML2PDF changed something and the encoding is no longer a problem.

No, XML2PDF did NOT change anything concerning encoding. I just tested it again with a unchanged XML from pyramid_oereb. It still doesnt't work.
But, anyway, if the XML2PDF proxy works together with XML2PDF, it's fine for me.

vvmruder:
Now I'am confused.

pramisters:
Try it:

https://oereb-dev.gis-daten.ch/oereb/report/form

Unchanged XML from pyramid_oereb --> Error:
[dataextract_6.txt](https://github.com/openoereb/pyramid_oereb/files/3405805/dataextract_6.txt)

Manually decoded XML from pyramid_oereb --> OK:
[dataextract_6_decoded.txt](https://github.com/openoereb/pyramid_oereb/files/3405812/dataextract_6_decoded.txt)

So, the XML2PDF proxy does some decoding somewhere, at least yesterday when we tested it, it worked fine. The only problem was, that the real estate geometry was missing.

jwkaltz:
I think we should merge this PR (without the xml templates adaptation), because it already provides a good first step for using the xml2pdf service.
There might be some additional things we will need to get it working 100% in every situation, but this is already a good first step.
(Regarding the geometry issue, I am working on a PR specifically for that).

vvmruder:
@jwkaltz Why you are having problems with changes in XML? They should be valid against schema. Right?

jwkaltz:
> @jwkaltz Why you are having problems with changes in XML? They should be valid against schema. Right?

Some tests are failing with these changes, for example:
E       AssertionError: assert '<data:Legend...:LegendAtWeb>' == '<data:LegendA...:LegendAtWeb>'
E         - <data:LegendAtWeb>http://example-legend.com</data:LegendAtWeb>
E         ?                       ^^^
E         + <data:LegendAtWeb>http%3A%2F%2Fexample-legend.com</data:LegendAtWeb>
E         ?                       ^^^^^^^^^

I was thinking that, if changes in the XML templates are not needed for the XML2PDF proxy, then we should not change the templates (at least not in this PR).

vvmruder:
Since this is a actual bug I think we should resolve it soon. The URL's are encoded twice. This leads to problems with the XML handling. But I added a Bug #884

pyramid_oereb/standard/pyramid_oereb.yml.mako
```diff
@@ -64,9 +64,10 @@ pyramid_oereb:
     # The pyramid renderer which is used as proxy pass through to the desired service for printable static
     # extract. Here you can define the path to the logic which prepares the output as payload for print
     # service and returns the result to the user.
-    renderer: pyramid_oereb.contrib.print_proxy.mapfish_print.Renderer
-    # The minimum buffer in pixel at 72 DPI between the real estate and the map's border.
-    buffer: 30
+    # renderer: pyramid_oereb.contrib.print_proxy.mapfish_print.Renderer
+    renderer: pyramid_oereb.contrib.print_proxy.xml_2_pdf.Renderer
```
jwkaltz: Can we please change this to a variable instead?
Something like 
render: ${print_renderer}

vvmruder: Feel free to do it.

jwkaltz: Exactly, that's why this should be a variable. But I can do that in a follow-up PR, to avoid mixing too many things.

PR #632, opened 2018-09-05
Configurable type mapping for real estate

### LINKED ISSUE #343: Missing mapping for RealEstate.type
The federal specification defines following output for RealEstate.type:

- 'RealEstate' => e.g. Liegenschaft in BL
- 'Distinct_and_permanent_rights.BuildingRight' => e.g. 'Baurecht' in BL
- 'Distinct_and_permanent_rights.right_to_spring_water' => e.g. 'Quellenrecht' in BL
- 'Distinct_and_permanent_rights.concession' => e.g. 'Konzessionsrecht' in BL
- 'Distinct_and_permanent_rights.other' => e.g. 'weitere' in BL
- 'Mineral_rights' => e.g. 'Bergwerk' in BL

vvmruder:
fixes #343

### DIFF
--- a/pyramid_oereb/standard/pyramid_oereb.yml.mako
+++ b/pyramid_oereb/standard/pyramid_oereb.yml.mako
@@ -216,6 +216,19 @@ pyramid_oereb:
         stroke_opacity: 0.6
         stroke_color: '#e60000'
         stroke_width: 5
+    type_mapping:
+      - mapping: RealEstate
+        type: Liegenschaft
+      - mapping: Distinct_and_permanent_rights.BuildingRight
+        type: Baurecht
+      - mapping: Distinct_and_permanent_rights.right_to_spring_water
+        type: Quellenrecht
+      - mapping: Distinct_and_permanent_rights.concession
+        type: Konzessionsrecht
+      - mapping: Distinct_and_permanent_rights.other
+        type: weitere
+      - mapping: Mineral_rights
+        type: Bergwerk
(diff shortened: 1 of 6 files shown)


pyramid_oereb/standard/pyramid_oereb.yml.mako
```diff
@@ -216,6 +216,19 @@ pyramid_oereb:
         stroke_opacity: 0.6
         stroke_color: '#e60000'
         stroke_width: 5
+    type_mapping:
```
voisardf: Am I right that this change means we have to adapt the configuration with our french translations (or do the mapping in fme beforehand)?

vvmruder: Yes exactly. You need to map the type inside your data to the federal defined types. The system allows multi mapping.

An Example. In your data both "Grundstück" and "Liegenschaft" means "RealEstate" than you can configure it like this:
```
type_mapping:
  - mapping: RealEstate
    type: Liegenschaft
  - mapping: RealEstate
    type: Grundstück
```

This way everyone can leave its AV data like it is and configure correct mapping instead :grimacing:

vvmruder: And you don't *must* configure this. If it is not configured it uses the value from data.

vvmruder:
@jwkaltz Is this mergable from your point of view?

jwkaltz:
Yes, thanks

PR #635, opened 2018-09-06
Improve intersection

vvmruder:
We encountered a strange behaviour when a area was intersected with the real estate and the result was a point. Then pyramid_oereb delivered a point. This is obviously incorrect.

As far as I think this should solve the problem. Of course we do not want to have changed geometries.

### DIFF
--- a/pyramid_oereb/lib/records/geometry.py
+++ b/pyramid_oereb/lib/records/geometry.py
@@ -103,8 +103,8 @@ class GeometryRecord(object):
                     self._test_passed = True
             else:
                 log.debug(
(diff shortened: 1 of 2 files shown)


vvmruder:
To be sure for the future I added some tests and the one which should produce the case area=>point fails. Maybe you can have a look on it? When I try the second scenario by hand I get a point after intersection. As I expect so far. But testing the `_test_passed` variable afterwards shows that it is true?!?!?!

vvmruder:
cheers! Almost sleeping-still... :sleeping:

kdeininger:
Good additional tests!

PR #645, opened 2018-09-10
WIP: Fix further information

### LINKED ISSUE #641: Additional information
The aditional information in extract is not necessary any longer. The actual federal specification shows the content now directly in the general information. So we can remove it from the config, the pdf and out of the extracts.

### LINKED ISSUE #644: Additional information on second page
This should be removed because of updated federal specification.

[Weisung-statischer-Auszug-de.pdf](https://github.com/camptocamp/pyramid_oereb/files/2366756/Weisung-statischer-Auszug-de.pdf)
auswahl_054

vvmruder:
fixes #644 fixes #641


vvmruder:
No I don't think so. We already have texts for general infomration. The part with furtherInfromation is deleted completely.

Or do I got something wrong here?

jwkaltz:
I was wondering whether the sentence "Weitere Informationen zum ÖREB finden Sie unter: ....", that I saw in the screenshot to #644 , should be added to our default config (in the attribute general_information).

vvmruder:
I don't think so. Maybe we should add explanation to version change information that this attribute was removed. But the text for general information is depending on cantonal interpretation. So, what we have is only a suggestion of what can be there...

PR #646, opened 2018-09-10
fix styling for document entities per PLR

### LINKED ISSUE #636: Draw SR number of documents not in separate Line
In PDF we do not wan't to have extra line for SR number. Instead it should be concatinated into the title like this:

{title} {sr_number}

vvmruder:
It turns out, that the numbers and abbreviations are not correctly shown in actual print template. In the moment we have:
auswahl_055

Specification defines:
auswahl_056

fixes #636

### DIFF
--- a/print/print-apps/oereb/legalprovision.jrxml
+++ b/print/print-apps/oereb/legalprovision.jrxml
@@ -26,7 +26,7 @@
 	<field name="Title" class="java.lang.String"/>
 	<field name="OfficialNumber" class="java.lang.String"/>
 	<field name="OfficialTitle" class="java.lang.String"/>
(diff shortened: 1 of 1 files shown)


vvmruder:
@ralphschulerSZ: Do you have an idea how to solve this? We need to concatenate title/original title, abbreviation and number to a complete title. And everything only if it is not null or empty string.

vvmruder:
Now it is in the title. We have the following logic of construction:

{title} {(abbreviation)}, {official number}

*  abbreviation is omitted if it is null
*  official number is omitted if it is null

vvmruder:
Any RFC's?

PR #652, opened 2018-09-13
Optional referer for GeoAdmin Search API

kdeininger:
The referer for the GeoAdmin Search API isn't mandatory anymore. The `ConfigurationError` on missing referer can be removed.

Additionally, the address geometry should be stored as shapely `Point` instead of WKT according to the documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ install: $(PYTHON_VENV)
 	touch $@
 
 $(SPHINXBUILD): .venv/requirements-timestamp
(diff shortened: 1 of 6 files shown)


kdeininger:
The docs build seems to fail with new Sphinx version, released yesterday, and Python 3.
https://pypi.org/project/Sphinx/#history

This can be solved by specifying the version `Sphinx<1.8`.

kdeininger:
@jwkaltz 
I've tested the changes on my dev instance and everything worked fine. Feel free to merge this PR before the release of version 1.2.2.

PR #653, opened 2018-09-13
WIP: deflatten responsible offices

### LINKED ISSUE #651: Multiple responsible office not rendered correctly
We saw a strange issue on responsible office per theme page. It is not correctly shown in PDF. It is possible that we have several, responsible offices per theme. In the moment it is concatenated in a strange way as you can see in this PDF on page 4
[PDF-Auszug-Augst_235_DiP.pdf](https://github.com/camptocamp/pyramid_oereb/files/2371399/PDF-Auszug-Augst_235_DiP.pdf)

A more detailed look on the PDF call with getspec parameter points out that name and Office at web is joined for the values by a `\n`...
auswahl_058

This is not correct. Responsible office should be an array of elements with *Name* and *OfficeAtWeb* attributes. As far as I understood the concatination happens inside the MapFishPrint Proxy:
https://github.com/camptocamp/pyramid_oereb/blob/master/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L341

I expect per responsible office an entry in PDF with name as title followed by URL.

My problem is, that I do not understand the logic of this... Maybe someone can support me here? @ger-benjamin or @jwkaltz or @voisardf ?

vvmruder:
fixes #651

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -30,7 +30,6 @@ class Renderer(JsonRenderer):
         for item in items:
             self._flatten_object(item, 'Lawstatus')
             self._localised_text(item, 'Lawstatus_Text')
-            self._flatten_object(item, 'ResponsibleOffice')
             self._multilingual_text(item, 'ResponsibleOffice_Name')
             self._multilingual_text(item, 'TextAtWeb')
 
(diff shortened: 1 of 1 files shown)


vvmruder:
@ger-benjamin: We decided to fix this in a later release. So for the moment it stays as it is. And I will work on this later on.

But really thank you for the hints.

pfirpfel:
Fixed with  #865 instead

PR #665, opened 2018-09-21
Provide import script for federal data

kdeininger:
Provide a Python script that downloads the data for a specified federal topic and imports it into the standard database if data has been modified.

The download URL can be defined for each topic in the `pyramid_oereb.yml`. The data is downloaded and extracted into a temporary directory. A comparison of the checksum within the download and the database is performed to evaluate if the content has been modified. If this is the case, elements are parsed from the XML and XTF file and written to the database. A `--force` parameter can be used to start the import anyway. All tables get truncated before the import is started (except `availability`).

Available parameters:
```
.venv/bin/import_federal_topic --help
```

To Do:
- [x] Add import script
- [x] Add tests
- [x] Update documentation


pyramid_oereb/standard/xtf_import/geometry.py
```diff
@@ -0,0 +1,142 @@
...
+    TAG_COORD = 'COORD'
+    TAG_ARC = 'ARC'
+
+    def __init__(self, session, model, geometry_type, arc_max_diff=0.001, arc_precision=3):
+        self._session = session
+        self._model = model
+        self._geometry_type = geometry_type
+        self._arc_max_diff = arc_max_diff
+        self._arc_precision = arc_precision
+        self._reproject = Reprojector().get_transformation_function(from_srs=21781, to_srs=2056)
```
voisardf: With the model is the reprojection function still used?

kdeininger: Good point! I'll read the target projection from the configuration file.

kdeininger:
@jwkaltz 

Which release should we consider for this PR?

There's a change in the standard database (added column "checksum" in table "data_integration"), which needs a little data migration to be done. So, maybe the next minor release should be preferred.

jwkaltz:
> @jwkaltz
> 
> Which release should we consider for this PR?
> 
> There's a change in the standard database (added column "checksum" in table "data_integration"), which needs a little data migration to be done. So, maybe the next minor release should be preferred.

@kdeininger so you mean not in 1.2.4, but in a 1.3.0 version?

PR #669, opened 2018-10-08
Update geolink-formatter to support new schema version 1.1.1

kdeininger:


pfirpfel:
Are there no changes to /tests/resources/geolink_v1.1.1.xml because of the new schema?

kdeininger:
No, the resulting XML output is the same, but the API should be called using the new version.

jwkaltz:
@vvmruder can we simply merge this, or is there any migration / compatibility issue to consider?

jwkaltz:
Done via PR #680

PR #675, opened 2018-10-17

sdebayle:


pfirpfel:
I reverted the simplejson update. simplejson 3.16.1 will [not be released on PyPI](https://github.com/simplejson/simplejson/issues/240#issuecomment-431014458).

jwkaltz:
Thanks, but that is just a fix for today; tomorrow, requires.io and simplejson will have their next groundhound day and propose once again to update to a non-existing version.
I will do a PR to fix the version to the existing one.

PR #677, opened 2018-10-19
Add halo effect to north arrow

ralphschulerSZ:
In order to see the north arrow on dark backgrounds a helo effect on the north arrow could be helpful.


voisardf:
I think the halo is even a requirement for the PDF. So 👍 from me

jwkaltz:
Thanks for the contribution, @ralphschulerSZ . To me, it looks good as well (tested locally with the standard test data).

PR #714, opened 2018-11-19
Implement Glossary sorting

### LINKED ISSUE #590: Glossary display order
(info and question from @ralphschulerSZ ):

The display order for the glossary items cannot be defined. The glossary currently is sorted by the order the items have in the table.

Should the pyramid_oereb standard be extended? For example, with a display_order_nr attribute?

vvmruder:
I want to share logic already implemented in BL for sorting JSON extract glossary elements alphabetically. This should fix #590 (to be tested)

### DIFF
--- a/pyramid_oereb/lib/renderer/extract/json_.py
+++ b/pyramid_oereb/lib/renderer/extract/json_.py
@@ -158,7 +158,9 @@ class Renderer(Base):
                     log.warning("glossary entry in requested language missing for title {}".format(gls.title))
 
             extract_dict['Glossary'] = glossaries
-
+        
+        # Sort glossary by requested language alphabetically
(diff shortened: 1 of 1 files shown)


peterschaer:
the import statement for the locale module is missing (`import locale`). 

Otherwise sorting works well with our data in french. In german there is still a small issue concerning items beginning with an umlaut (e.g. ÖREB-Kataster). They will be sorted at the end (after Z) and not under O.

I was able to enforce this sorting (Ö under O etc) by setting the defaultlocale explicitly:

`locale.setlocale(locale.LC_ALL, '')`

I'm running this on Windows. setlocale is `German_Switzerland.1252`

jwkaltz:
@vvmruder Benjamin has done a new PR for this issue, #750 .
Can we close this one?

PR #735, opened 2019-01-07
refine docs

vvmruder:
Added some more hints for the migration script and some linking.

### DIFF
--- a/doc/source/changes.rst
+++ b/doc/source/changes.rst
@@ -28,8 +28,7 @@ using the standard scripts. This will solve it for you.
 
 Database
 ^^^^^^^^
-New column ``checksum`` in all **federal** oereb theme *DataIntegration* tables you want to use the new
-import script feature for.
+New column ``checksum`` in all oereb theme *DataIntegration* tables.
 
 .. _changes-version-1.2.3:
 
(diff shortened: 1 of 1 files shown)


doc/source/changes.rst
```diff
@@ -10,23 +10,26 @@ version. Often a version upgrade changes or adds parameters which are used.
...
-If you have custom models (for example, for oereblex), you need to add an attribute ``checksum`` of type String to these.
-Alternatively, you can recreate your models using the standard scripts.
+If you have custom models (for example, for oereblex), you need to add an attribute ``checksum`` of type
+String to these (in class definition of *DataIntegration* model). Alternatively, you can recreate your models
+using the standard scripts. This will solve it for you.
 
 Database
 ^^^^^^^^
-New column ``checksum`` in all oereb theme tables.
+New column ``checksum`` in all **federal** oereb theme *DataIntegration* tables you want to use the new
```
jwkaltz: This is a bit confusing, because all oereb theme tables now have a checksum column, not just the federal themes. Even if (in the standard logic) only the checksum columns from the federal themes will really be used.

vvmruder: But "...all oereb theme tables" is even more. Do you not think so? This way I would think each table of a schema need a checksum attribute.

jwkaltz: Ah yes, so we should say 
"... in all oereb theme *DataIntegration* tables""
right?

vvmruder: Yes

jwkaltz:
@vvmruder feel free to merge :)

Issue #638, opened 2018-09-07
Wrong PDF format


vvmruder:
PDF must be delivered PDF/A-1a!

vvmruder:
Low priority

pfirpfel:
To test: https://community.jaspersoft.com/wiki/pdf-support-jasperreports-library

pfirpfel:
Implemented with #852

PR #737, opened 2019-01-09
[work in progress] PDF/A-1a conformance

pfirpfel:
Addresses #638


pfirpfel:
mapfish/mapfish-print#939 will allow disabling tranpsarency in mapfish-print

pfirpfel:
Continued in #852

Issue #704, opened 2018-11-14
map.legendatweb (view_service.legend_at_web) not multilingual


peterschaer:
`map.legendatweb` supports only a single link. this single link is used in all the languages the extract can be produced in. if we define a link to a german legend this link is also displayed in a french extract.

![legendatweb](https://user-images.githubusercontent.com/3659322/48470834-d49d9880-e7f2-11e8-9889-acd3e145e9f6.PNG)

I know that in the _weisung_ this information is not multilingual. Therefore pyramid_oereb is compliant. But in multilingual cases an extension of this behaviour would be valuable.

vvmruder:
This breaks the actual federal specification of DataExtract. Thats why we need first the OK from federal offices to update specification. After this we can update behaviour.

Somebody does have other suggestions?

vvmruder:
After thinking a little about that, it would be possible to change only the behaviour in the standard DB scheme for this attribute to be multilingual and deliver the result in DataExtract as today. So we 'shadow' the behaviour and are free to adapt to the maybe new specification somewhere in the future...

pfirpfel:
Implemented with  #739

PR #739, opened 2019-01-10
Multilingual map.legend_at_web

pfirpfel:
Implements #704

TODO:

- [x] add more tests:
  - [x] test renderer by requesting unavailable lang, non-default lang
-  [x] add documentation
-  [x] add ~db migration script~ *migration* documentation
-  [x] check print templates
-  [x] make the xml rendering work and add a test for it

### DIFF
--- a/tests/records/test_extract.py
+++ b/tests/records/test_extract.py
@@ -35,7 +35,7 @@ def test_init():
         'test_link',
         1,
         1.0,
-        'test_legend'
+        {'lang': 'test_legend'}
     ))
     plr_office = OfficeRecord({u'en': u'PLR Authority'})
     resolver = DottedNameResolver()
--- a/tests/records/test_extract.py
+++ b/tests/records/test_extract.py
@@ -35,7 +35,7 @@ def test_init():
         'test_link',
         1,
         1.0,
-        {'lang': 'test_legend'}
+        {'de': 'test_legend'}
     ))
     plr_office = OfficeRecord({u'en': u'PLR Authority'})
     resolver = DottedNameResolver()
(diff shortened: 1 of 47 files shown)


jwkaltz:
Because this will involve a database change, it should not be included in release 1.3.1. Instead we should release 1.3.1 just before this PR is ready, then start a 1.4 version and finish this PR.

pfirpfel:
Move tasks of adding more tests for XML rendering to #779

pfirpfel:
@peterschaer Could you test the modified field? The JSON extract and the mapfish print PDF should be fine from my testing.

pfirpfel:
Cherry-picked #782 and fixed XML rendering

tests/records/test_extract.py
```diff
@@ -35,7 +35,7 @@ def test_init():
         'test_link',
         1,
         1.0,
-        'test_legend'
+        {'lang': 'test_legend'}
```
ger-benjamin: It will not fail with `lang` as lang ?

tests/webservice/test_getegrid.py
```diff
@@ -160,7 +160,7 @@ def test_get_egrid_response():
         view_service = ViewServiceRecord('test',
                                          1,
                                          1.0,
-                                         'test',
+                                         {'lang': 'test'},
```
ger-benjamin: same as upper

tests/records/test_extract.py
pfirpfel: No, it's not getting checked here. But I agree, it is confusing. I change it.

PR #746, opened 2019-01-11
add some debug messages

jwkaltz:


pyramid_oereb/lib/records/documents.py
```diff
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
+import logging
 import warnings
 import datetime
 
+LOG = logging.getLogger(__name__)
```
pfirpfel: Nitpicking: can we keep the log variable lowercase?

jwkaltz: OK

pfirpfel: thanks

PR #756, opened 2019-01-22
font size in topicpage.jrxml

ralphschulerSZ:
Same font size of the footer and the page counter text field as in other templates. In our instance it has the effect, that the information is not visible when we use 6.5 instead of 6.

### DIFF
--- a/print/print-apps/oereb/topicpage.jrxml
+++ b/print/print-apps/oereb/topicpage.jrxml
@@ -543,7 +543,7 @@
 					<property name="com.jaspersoft.studio.unit.width" value="pixel"/>
 				</reportElement>
 				<textElement>
-					<font fontName="Cadastra" size="6.5"/>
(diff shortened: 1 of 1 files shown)


pfirpfel:
I cannot reproduce your issue with 6.5pt

I not sure changing the font size to 6pt is a good idea, since the official template demands 6.5pt:
selection_153

vvmruder:
About the font sizes: We had this issue before. After checking it with federal offices it is ok to round it to the next full lower size. So 6.5 would be 6.

No problem with specification. This is valid for all font sizes all over the PDF.

jwkaltz:
@ralphschulerSZ @vvmruder actually, these changes have already been done in https://github.com/openoereb/pyramid_oereb/pull/814. When testing with the new version of Jasper Reports, the problem occurred in other circumstances as well.
So we can close this PR.

PR #773, opened 2019-02-04
Also consider the path-part of a WMS service for creating legend entries

andreasneumann:
Without this change, WMS requests failed, if there was a /path/to/wms after the server name.


jwkaltz:
@andreasneumann feel free to merge the change

PR #777, opened 2019-02-12

pfirpfel:


jwkaltz:
@pfirpfel it looks like we are missing codacy checks since moving the organization, can you please have a look?

PR #785, opened 2019-02-15
Small changes from codacy feedback

pfirpfel:


ger-benjamin:
You also have removed the "part_in_percent". So now I'm not sure it's 100% correct/coherent (even if it works):
Perhaps the best we can do is to keep this "part_in_percent" removed, use a private "_part_in_percent" value in the code instead of "part_in_percent" and add a 
```
  @property
    def part_in_percent(self):
         """decimal or None: (decimal): Part of the property area touched by the restriction in percent."""
        return self._part_in_percent
```

As for area_share

pfirpfel:
> You also have removed the "part_in_percent". So now I'm not sure it's 100% correct/coherent (even if it works):
> Perhaps the best we can do is to keep this "part_in_percent" removed, use a private "_part_in_percent" value in the code instead of "part_in_percent" and add a
> 
> ```
>   @property
>     def part_in_percent(self):
>          """decimal or None: (decimal): Part of the property area touched by the restriction in percent."""
>         return self._part_in_percent
> ```
> 
> As for area_share

I had to add a setter as well, because other parts of the code write to it. Not sure if it is nicer that way though...

PR #810, opened 2019-03-11
Fix version of pyproj to avoid problematic >=2.0.0 releases

pfirpfel:
The new major release of pyproj returns different (wrong?) results for coordinate transformations.
Setting the version to be lower than the new 2.0.0 major release lets our tests pass again.

[Related merge request for python-reprojector](https://gitlab.com/gf-bl/python-reprojector/merge_requests/17)


pfirpfel:
@kdeininger: FYI,  I'm "fixing" the version here for now.

PR #821, opened 2019-03-26
Fix import script for federal topics

### LINKED ISSUE #811: Broken import script for federal topics
The import script for the federal topics seems to be broken. It throws an AssertionError on `assert Config._config is not None`.

Furthermore, the values of the new multilingual attributes (`sub_theme` and `legend_at_web`) should be wrapped in dictionaries using a default language:

https://github.com/openoereb/pyramid_oereb/blob/7c6bb1b5d738ca0de4a19b83f9f184c9de81b7bf/pyramid_oereb/standard/xtf_import/public_law_restriction.py#L27

https://github.com/openoereb/pyramid_oereb/blob/7c6bb1b5d738ca0de4a19b83f9f184c9de81b7bf/pyramid_oereb/standard/xtf_import/view_service.py#L28

https://github.com/openoereb/pyramid_oereb/blob/7c6bb1b5d738ca0de4a19b83f9f184c9de81b7bf/pyramid_oereb/standard/xtf_import/legend_entry.py#L41

For example:

```python
sub_theme={
    default_language: parse_string(legend_entry, self.TAG_SUB_THEME)
}
```

The default language could be set using an additional command line argument.

pfirpfel:
Fixes  #811

### DIFF
--- a/pyramid_oereb/standard/xtf_import/__init__.py
+++ b/pyramid_oereb/standard/xtf_import/__init__.py
@@ -17,6 +17,7 @@ from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 from tqdm import tqdm
 
+from pyramid_oereb.lib.config import Config
 from pyramid_oereb.standard.xtf_import.article import Article
 from pyramid_oereb.standard.xtf_import.base_refinement import BaseRefinement
 from pyramid_oereb.standard.xtf_import.document import Document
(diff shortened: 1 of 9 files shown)


pfirpfel:
I was able to import AirportsSecurityZonePlans with these changes.

pfirpfel:
This fix is now available as a beta release via PyPi:
`pyramid-oereb==1.4.3b0`

Issue #824, opened 2019-03-28
PartInPercent floating precision is too long in PDF extract


svareg:
Hello,

In some PDF extract, we have the following output :

![grafik](https://user-images.githubusercontent.com/36037102/55147699-26621200-5147-11e9-974d-4b45459fece7.png)

This is due to the folowing code in [mapfish_proxy ](https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/contrib/print_proxy/mapfish_print.py) (line 367): 

```
for item in ['AreaShare', 'LengthShare', 'PartInPercent']:
      if item in legend:
          if item in legends[type_]:
              legends[type_][item] += legend[item]
          else:
              legends[type_][item] = legend[item]
```

Here we should make a rounding of `legends[type_][item] += legend[item]`, because we are working with float numbers :
![grafik](https://user-images.githubusercontent.com/36037102/55148031-c6b83680-5147-11e9-895a-401df9e5e2e4.png)

Thanks !

pfirpfel:
Fixed with  #826, will be in the next release

PR #826, opened 2019-04-02
Round PartInPercent before formatting it

pfirpfel:
Probably solves #824


svareg:
Yeah, problem solved !

grafik

Thanks!

PR #828, opened 2019-04-04
Fix encoding issue in __str__ method

kdeininger:
The PLR record has a string formatting method, which is used for debugging. This method needs to encode non-ASCII content, especially for the attribute `information`.


vvmruder:
@pfirpfel This can be merged if ok for you.

PR #831, opened 2019-04-10
make customParams (for mapfish print WMS url setup) parametrizeable

jwkaltz:
make customParams (for mapfish print WMS url setup) parametrizeable, instead of hard-coded as they are now.

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -147,7 +147,7 @@ class Renderer(JsonRenderer):
                 result = wms_url_params
             else:
                 log.warning("get_wms_url_params() ignoring unaccepted configuration value {}"
-                         .format(wms_url_params))
+                            .format(wms_url_params))
         else:
             log.info("no wms_url_params configuration detected; using default value")
             result = {'TRANSPARENT': 'true'}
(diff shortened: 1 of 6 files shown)


ralphschulerSZ:
Hi Wolfgang, I tested your code and it works fine here.

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -131,6 +131,29 @@ def __call__(self, value, system):
...
+        :rtype: list
+        """
+        result = {}
+        wms_url_params = Config.get('print', {}).get('wms_url_params', False)
+        if wms_url_params:
+            log.debug("get_wms_url_params() read configuration {}".format(wms_url_params))
+            if isinstance(wms_url_params, dict):
+                result = wms_url_params
+            else:
+                log.warning("get_wms_url_params() ignoring unaccepted configuration value {}"
```
pfirpfel: Shoudln't it still return the default values in this case?

jwkaltz: IMO in that case it is better to return nothing, this way it is more likely that the user (the bad configurator) will notice something is wrong.

pfirpfel: Ok for me

PR #841, opened 2019-04-25
Make Section 'Certification' in PDF configurable

### LINKED ISSUE #765: Make Section 'Certification' in PDF configurable
We currently don't need the section 'Certtification' (Beglaubigung) on the front page of the PDF-extract. The visibility of this section should be configurable

marionb:
This adds a new config variable to make the Certification section in the print template configurable.
The variable defaults to `true` for which value the certification section is shown in the pdf extract.

Fixes #765

### DIFF
--- a/docker/config.yaml
+++ b/docker/config.yaml
@@ -86,7 +86,7 @@ vars:
       # Whether to display the RealEstate_SubunitOfLandRegister (Grundbuchkreis) in the pdf extract or not.
       # Default to true.
       display_real_estate_subunit_of_land_register: true
-      # Wheder to display the Certification section in the pdf extract or not.
+      # Whether to display the Certification section in the pdf extract or not.
       # Default to true
       display_certification: false
       # Split themes up, so that each sub theme gets its own page
--- a/docker/config.yaml
+++ b/docker/config.yaml
@@ -86,6 +86,9 @@ vars:
       # Whether to display the RealEstate_SubunitOfLandRegister (Grundbuchkreis) in the pdf extract or not.
       # Default to true.
       display_real_estate_subunit_of_land_register: true
+      # Wheder to display the Certification section in the pdf extract or not.
+      # Default to true
+      display_certification: false
       # Split themes up, so that each sub theme gets its own page
       # Disabled by default.
       split_sub_themes: true
--- a/docker/config.yaml
+++ b/docker/config.yaml
@@ -88,7 +88,7 @@ vars:
       display_real_estate_subunit_of_land_register: true
       # Whether to display the Certification section in the pdf extract or not.
       # Default to true
-      display_certification: false
+      display_certification: true
       # Split themes up, so that each sub theme gets its own page
       # Disabled by default.
       split_sub_themes: true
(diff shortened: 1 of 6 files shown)


docker/config.yaml
```diff
@@ -86,6 +86,9 @@ vars:
       # Whether to display the RealEstate_SubunitOfLandRegister (Grundbuchkreis) in the pdf extract or not.
       # Default to true.
       display_real_estate_subunit_of_land_register: true
+      # Wheder to display the Certification section in the pdf extract or not.
```
jwkaltz: Whether

pyramid_oereb/standard/pyramid_oereb.yml.mako
```diff
@@ -85,6 +85,9 @@ pyramid_oereb:
     # Whether to display the RealEstate_SubunitOfLandRegister (Grundbuchkreis) in the pdf extract or not.
     # Default to true.
     display_real_estate_subunit_of_land_register: true
+    # whether to display the "Certification" section in the pdf extract or not.
+    # Default to true
+    display_certification: false
```
pfirpfel: If you write `Default to true`, the default value should also be true :)

pfirpfel: It's `true`, if the value is not set. But if your default config defines it as false by default, this will cause confusion.
Do it either like the value above (display_real_estate_subunit_of_land_register) and set it to true or explain that if undefined, it will use true, and comment the `display_certification: false` out. That way people can uncomment it to disable the feature.

print/print-apps/oereb/mainpage.jrxml
```diff
@@ -155,6 +159,7 @@
 			<textField>
 				<reportElement isPrintRepeatedValues="false" x="0" y="709" width="490" height="12" uuid="5090c0e3-b8c0-4df8-9447-d203a7d6a664">
 					<property name="com.jaspersoft.studio.unit.y" value="pixel"/>
+					<printWhenExpression><![CDATA[$P{Display_Certification}==true]]></printWhenExpression>
```
pfirpfel: You can remove the `==true`

print/print-apps/oereb/mainpage.jrxml
```diff
@@ -146,6 +149,7 @@
 			<textField isBlankWhenNull="true">
 				<reportElement x="0" y="725" width="493" height="10" uuid="f8b5d9a5-d842-43c8-b8d7-a8b68f697303">
 					<property name="com.jaspersoft.studio.unit.y" value="mm"/>
+					<printWhenExpression><![CDATA[$P{Display_Certification}==true]]></printWhenExpression>
```
pfirpfel: You can remove the `==true`

print/print-apps/oereb/mainpage.jrxml
```diff
@@ -137,6 +139,7 @@
 			<textField>
 				<reportElement x="0" y="697" width="100" height="12" uuid="e0c1f775-61b3-41c4-a7ec-b8e68adeeb11">
 					<property name="com.jaspersoft.studio.unit.y" value="mm"/>
+					<printWhenExpression><![CDATA[true]]></printWhenExpression>
```
pfirpfel: This should also be `<![CDATA[$P{Display_Certification}]]>`, otherwise there is still the title there...

marionb:
Thanks for the reviews.

PR #842, opened 2019-04-26
Fix long titles in pdf

### LINKED ISSUE #703: Exclusion of liability with multiline title overlapping the content
An exclusion of liability with a title running on multiple lines is overlapping the content:

mutlipleexl

marionb:
This PR:
- fixes titles overlapping the contents in the exclusion of liability section
- adds data which is shown in the exclusion of liability section

Fixes #703

Please note that the `Grundlagendaten` are pushed down to the end of the exclusion of liability (see screenshot). @peterschaer can you please check if this is correct.

grundlagenDaten

### DIFF
--- a/.codacy.yaml
+++ b/.codacy.yaml
@@ -1,2 +1,3 @@
 exclude_paths:
   - 'tests/**'
+  - 'sample_data/*.json'
--- a/.codacy.yaml
+++ b/.codacy.yaml
@@ -1,3 +1,3 @@
 exclude_paths:
   - 'tests/**'
-  - 'sample_data/**/*.json'
+  - 'sample_data/**.json'
--- a/.codacy.yaml
+++ b/.codacy.yaml
@@ -1,3 +1,3 @@
 exclude_paths:
   - 'tests/**'
-  - 'sample_data/*.json'
+  - 'sample_data/**/*.json'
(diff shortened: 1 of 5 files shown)


pfirpfel:
Do you think it is possible to make the spacing in between title and text of a single exclusion and in between different exclusions nicer?

For example a single-line title has a margin to its text, but this margin gets lost if the title uses more than one line. I did not test titles with more than one line.
Also, it would looks nicer if there would be some space between the text and the next title.

What do you think, @peterschaer ?

My example:

Selection_175

pfirpfel:
Code looks good to me. Thanks for adding the additional sample data!

.codacy.yaml
```diff
@@ -1,2 +1,3 @@
 exclude_paths:
   - 'tests/**'
+  - 'sample_data/*.json'
```
pfirpfel: I did not test this, but if I understand the [syntax](https://support.codacy.com/hc/en-us/articles/115002130625-Codacy-Configuration-File) correctly, this would also cover the json files in the sub folders:
`- 'sample_data/**/*.json'`

marionb: You are right, thinks. I did not pay attention to the sub-folders

Issue #774, opened 2019-02-05
PDF extract without PLR


peterschaer:
We have cases where there is a real estate without any public law restriction. In these cases the pdf extract contains a page with an empty map and an empty table (s. p.4 in [1_DE_PYRAMID.pdf](https://github.com/openoereb/pyramid_oereb/files/2832244/1_DE_PYRAMID.pdf)). This page is confusing and should be omitted.

svareg:
Hello, 

At BS we've just spotted the same problem.
And I just looked at OEREB-Kataster in BL, it can be reproduced here too.
Is there any simple correction or workaround ?

Thanks !

vvmruder:
For me there are 2 possible solutions to be discussed:

1. We return a valid HTML code 404 (HTTP not Found) if there was no PLR matching.
2. We return a correctly styled empty PDF (this would need more clarification how it should look).

Solution 1 is obviously the more easy one.

Discussion opened... :+1:

peterschaer:
I don't think Solution 1 is a good one.  I think that the user expects a PDF like this one ([CH373540414688.pdf](https://github.com/openoereb/pyramid_oereb/files/2914827/CH373540414688.pdf)) if there is no PLR on a real estate. In the TOC on page 2 the user gets all the relevant information. This is much better than a 404 code. Why should we not be able to produce a PDF in those cases?

peterschaer:
The PDF we get from pyramid_oereb and mapfish_print is actually perfect for me except for the superfluous page 3 in the example I added at the start of this issue.

voisardf:
Well, - normally there should be no property without restrictions I guess as there should be at least some data regarding the LandUse. Now, I see that the case can appear if not all data is already available.

Personnally I think we have the legal obligation to produce an PDF, even if there are no restrictions (available). Otherwise I'd agree that an 404 would a handy solution...

my 2 cents

svareg:
A 404 error can be a quick temporary solution, to prevent the generation of a dirty PDF. 
But I agree with @peterschaer, I feel we should be able to generate a correct PDF file even if there isn't any concerned PLR, so that the user gets what he is waiting for : a document that indicates that there is no concerned PLR.

@voisardf : I'm not 100% sure, but I think there can be cases where there isn't any concerned PLR, even with landuse. For example is we generate a PDF for a street parcel.

vvmruder:
Indeed. I agree on solution 2. Solution one would be a temporary fix only.

[...]

PR #846, opened 2019-04-30
Only show map and table if a PLR is available for a real estate

marionb:
Addresses issue #774.
This only prints the PLR section of the PDF if at least one PLR is available for a real estate.

@peterschaer and @svareg can please you check if this works for you.

**Developers note:**
To reconstruct a real estate with out a PLR do the following:

in the print_proxy make the function [convert_to_printable_extract](https://github.com/openoereb/pyramid_oereb/blob/af0262a654a540fa1fe4d311077a23d7ee0fa291/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L161) return an empty `RealEstate_RestrictionOnLandownership`in the `extract_dict` list by adding the following lines at the end of the function

        extract_dict['RealEstate_RestrictionOnLandownership'] = []
        extract_dict['NotConcernedTheme'] += extract_dict['ConcernedTheme']
        extract_dict['ConcernedTheme'] = []

### DIFF
--- a/print/print-apps/oereb/pdfextract.jrxml
+++ b/print/print-apps/oereb/pdfextract.jrxml
@@ -310,7 +310,11 @@
 	</group>
 	<detail>
 		<part uuid="dbf7b612-de21-4609-8537-6538c611480c">
-			<!-- IMPORTANT NOTE: This report part is only printed if there is a PLR for this real estate to avoid empty pages and tables. -->
+			<!--
+			IMPORTANT NOTE:
+			The RealEstate_RestrictionOnLandownershipDataSource is only of type JRMapCollectionDataSource
+			if there is at least one PLR for the real estate. If there is no PLR it's empty and of type JREmptyDataSource.
+			 -->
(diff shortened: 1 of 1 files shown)


print/print-apps/oereb/pdfextract.jrxml
```diff
@@ -310,6 +310,8 @@
 	</group>
 	<detail>
 		<part uuid="dbf7b612-de21-4609-8537-6538c611480c">
+			<!-- IMPORTANT NOTE: This report part is only printed if there is a PLR for this real estate to avoid empty pages and tables. -->
+			<printWhenExpression><![CDATA[$P{RealEstate_RestrictionOnLandownershipDataSource} instanceof net.sf.jasperreports.engine.data.JRMapCollectionDataSource]]></printWhenExpression>
```
pfirpfel: Can we add something like:
`Only if there is at least one PLR in RealEstate_RestrictionOnLandownershipDataSource, it has the type JRMapCollectionDataSource. If it's empty, it is a JREmptyDataSource.`

Otherwise it may be hard to understand, why we check for a type instead of just counting elements.

PR #851, opened 2019-05-15
Sorting legend

### LINKED ISSUE #380: Mapfish Print: Sorting of legend
The legend in static extract should be sorted like follows:
1. subtheme
2. geometry_type
3. part_in_percent

marionb:
Sorts the legends according to:
1. geometry Type
2. length or pars in percent

Fixes #380

**NOTE**
for testing you can set the sub-theme flag to false in [docker/config.yaml](https://github.com/openoereb/pyramid_oereb/blob/master/docker/config.yaml#L94) to have a better example with more legends.

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -23,6 +23,8 @@ else:
 
 log = logging.getLogger(__name__)
 
+LEGEND_ELEMENT_SORT_ORDER = {'AreaShare': 1, 'LengthShare': 2, 'NrOfPoints': 3}
+
 
 class Renderer(JsonRenderer):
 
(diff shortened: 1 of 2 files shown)


pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -601,3 +601,34 @@ def _get_sorter(theme_code):
...
+            elem: one legend entry
+
+        Returns:
+            geom_type, int value
+        """
+        geom_type = elem['Geom_Type']
+        if geom_type == 'NrOfPoints':
+            return (geom_type, 0)
+        else:
+            return geom_type, elem[geom_type]
```
pfirpfel: Descending order might be better, because the larger values are more relevant?

`return geom_type, -elem[geom_type]`

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -601,3 +601,34 @@ def _get_sorter(theme_code):
...
+        """
+        return sorted(legend_list, key=self._sort_legend_elem)
+
+    @staticmethod
+    def _sort_legend_elem(elem):
+        """
+        Returns a tuple of the geometry type of the element and the value of the geometry
+        type if it is available. Else it returns the geometry type and 0 as value
+
+        Args:
```
pfirpfel: Using a tuple as sorted-key is pretty advanced. I needed to look that up.
Since we know the data that gets sorted, I would go more into details here.

It sorts by geom_type, they are sorted alphabetically: AreaShare, LengthShare, NrOfPoints 
This turns out to be the desired order. If the data changes, alphabetically might not be enough anymore.

Second priority in ordering here is the area/length. I would also mention that in a comment. 
As explained in another comment, I think ascending order is better here.

pfirpfel: May be an even more explicit version:

```
#global variable 
LEGEND_ELEMENT_SORT_ORDER = { 'AreaShare': 1, 'LengthShare': 2, 'NrOfPoints': 3}

category = LEGEND_ELEMENT_SORT_ORDER.get(geom_type, 4) # everything else later
value = elem.get(geom_type, 0)

return (category, value)
```

marionb: @pfirpfel do you think it makes sens to put the LEGEND_ELEMENT_SORT_ORDER in the project config? (https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/pyramid_oereb.yml.mako)

jwkaltz: project config is not necessary, there are currently no requirements for that.

marionb: OK @jwkaltz thanks for the info.

marionb:
This looks good I'll merge. Thanks for your work @pfirpfel

PR #857, opened 2019-05-23
remove (mapfish) print, was moved to pyramid_oereb_mfp

### LINKED ISSUE #459: transfer print into separate project

jwkaltz:
Fixes #459

### DIFF
--- a/docker/config.yaml
+++ b/docker/config.yaml
@@ -92,6 +92,10 @@ vars:
       # Split themes up, so that each sub theme gets its own page
       # Disabled by default.
       split_sub_themes: true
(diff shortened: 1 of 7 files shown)


pfirpfel:
I want to do another release before merging this.

PR #858, opened 2019-05-23
optimize implementation sorting of themes for toc

### LINKED ISSUE #443: optimize implementation sorting of themes for toc
Currently, when the themes are separated into concerned_theme and not_concerned_theme lists, they do keep their relative sorting order within these lists. But the overall sorting order of all themes as defined in the configuration is then no longer directly known. When a theme has to get moved from concerned_theme to not_concerned_theme, we can not easily know in which place of the list to put it.

In https://github.com/camptocamp/pyramid_oereb/pull/429, a sorting "from scratch" has been implemented, to resort the not_concerned_theme in the right order. But this code could be more efficient and clearer if we keep more knowledge about the original sort order somewhere in the data structures.

jwkaltz:
Fixes #443

### DIFF
--- a/pyramid_oereb/lib/config.py
+++ b/pyramid_oereb/lib/config.py
@@ -63,14 +63,12 @@ class Config(object):
         result = []
         plrs = Config._config.get('plrs')
         if plrs and isinstance(plrs, list):
-            position = 1
-            for theme in plrs:
+            for position, theme in enumerate(plrs, start=1):
                 result.append(ThemeRecord(
                     theme.get('code'),
                     theme.get('text'),
                     position
                 ))
-                position += 1
         return result
 
     @staticmethod
--- a/pyramid_oereb/lib/config.py
+++ b/pyramid_oereb/lib/config.py
@@ -89,15 +89,13 @@ class Config(object):
 
         plrs = Config._config.get('plrs')
         if plrs and isinstance(plrs, list):
-            position = 1
-            for theme in plrs:
+            for position, theme in enumerate(plrs, start=1):
                 if theme.get('code') == code:
                     return ThemeRecord(
                         theme.get('code'),
                         theme.get('text'),
                         position
                     )
-                position += 1
         return None
 
     @staticmethod
--- a/pyramid_oereb/lib/config.py
+++ b/pyramid_oereb/lib/config.py
@@ -63,7 +63,7 @@ class Config(object):
         result = []
         plrs = Config._config.get('plrs')
         if plrs and isinstance(plrs, list):
-            position = 0
+            position = 1
             for theme in plrs:
                 result.append(ThemeRecord(
                     theme.get('code'),
--- a/pyramid_oereb/lib/config.py
+++ b/pyramid_oereb/lib/config.py
@@ -63,11 +63,14 @@ class Config(object):
         result = []
         plrs = Config._config.get('plrs')
         if plrs and isinstance(plrs, list):
+            position = 0
             for theme in plrs:
                 result.append(ThemeRecord(
                     theme.get('code'),
-                    theme.get('text')
+                    theme.get('text'),
+                    position
                 ))
+                position += 1
         return result
 
     @staticmethod
@@ -86,12 +89,15 @@ class Config(object):
 
         plrs = Config._config.get('plrs')
         if plrs and isinstance(plrs, list):
+            position = 1
             for theme in plrs:
                 if theme.get('code') == code:
                     return ThemeRecord(
                         theme.get('code'),
-                        theme.get('text')
+                        theme.get('text'),
+                        position
                     )
+                position += 1
         return None
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


pyramid_oereb/lib/processor.py
```diff
@@ -121,28 +122,14 @@ def is_inside_plr(theme_code):
...
-
-        for theme_code in theme_codes:
-            theme = get_not_concerned_theme(extract, theme_code)
-            if theme is not None:
-                sorted_not_concerned_theme.append(theme)
-
-        extract.not_concerned_theme = sorted_not_concerned_theme
-        return extract
+    def sortfunc(elem):
+        return elem[len(elem - 1)]
```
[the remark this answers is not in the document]
jwkaltz: Thanks, fixed

pyramid_oereb/lib/config.py
```diff
@@ -63,11 +63,14 @@ def get_themes():
         result = []
         plrs = Config._config.get('plrs')
         if plrs and isinstance(plrs, list):
+            position = 0
```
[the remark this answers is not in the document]
jwkaltz: Changing it to start at 1 here, also, so it is more consistent.
Note however that in the end, it does not matter, only the relative number of the themes is relevant (for ordering), not the absolute number.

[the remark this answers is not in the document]
jwkaltz: By the way, this is also why it does matter for the tests

pyramid_oereb/lib/config.py
```diff
@@ -86,12 +89,15 @@ def get_theme(code):
 
         plrs = Config._config.get('plrs')
         if plrs and isinstance(plrs, list):
+            position = 1
             for theme in plrs:
```
pfirpfel: Because plrs is a list, you can just use [enumerate](http://book.pythontips.com/en/latest/enumerate.html) and get rid of the position:

`for position, theme in enumerate(plrs, start=1):`

The start property allows the ongodly choice of starting to count with 1...

pyramid_oereb/lib/sources/plr.py
```diff
@@ -83,5 +83,6 @@ def read(self, real_estate, bbox):
                 all public law restrictions. This is related to the fact that we need to provide not only the
                 public law restrictions that are related to the real estate but also the ones which are in
                 the visible extent of the map.
+            position: relative position of the plr (within a list of plrs)
```
pfirpfel: Can you add the type?
`(int or None)`

pyramid_oereb/standard/sources/plr.py
```diff
@@ -484,16 +483,18 @@ def collect_legend_entries_by_bbox(self, session, bbox):
...
-    def read(self, real_estate, bbox):
+    def read(self, real_estate, bbox, position=None):
         """
         The read point which creates a extract, depending on a passed real estate.
 
         Args:
             real_estate (pyramid_oereb.lib.records.real_estate.RealEstateRecord): The real
                 estate in its record representation.
             bbox (shapely.geometry.base.BaseGeometry): The bbox to search the records.
+            position: relative position of the plr (within a list of plrs)
```
pfirpfel: Can you add the type?
(int or None)

pyramid_oereb/lib/config.py
jwkaltz: Thanks

pyramid_oereb/lib/sources/plr.py
jwkaltz: Done

pyramid_oereb/standard/sources/plr.py
jwkaltz: Done

pyramid_oereb/lib/config.py
pfirpfel: There are two other instances in your PR where you can do the same.

pyramid_oereb/lib/readers/extract.py
```diff
@@ -141,15 +141,17 @@ def read(self, real_estate, municipality, params):
 
         if municipality.published:
 
+            position = 1
             for plr_source in self._plr_sources_:
```
pfirpfel: `for position, plr_source in enumerate(self._plr_sources_, start=1):`

pyramid_oereb/lib/config.py
```diff
@@ -63,11 +63,14 @@ def get_themes():
         result = []
         plrs = Config._config.get('plrs')
         if plrs and isinstance(plrs, list):
+            position = 1
             for theme in plrs:
```
pfirpfel: `for position, theme in enumerate(plrs, start=1):`

jwkaltz: Done

pyramid_oereb/lib/readers/extract.py
jwkaltz: Done

PR #859, opened 2019-05-23
Fix page numbers in longer TOC

marionb:
This is a suggestion of how to correctly show the page numbers for more complex page numbers.
@jwkaltz and @pfirpfel can you take a look at this option and give me some input of whether this can be used.

TODO:

- ~~Add a proper variable in the Docker config.yaml to decide which TOC to use the simple or the complex~~
- ~~Check that the the spacing of the elements in the new TOC are kept as closely to the intended template.~~
- ~~Add the `Full` version of the TOC which allows attachments~~

TO TEST:
Enable this feature here: https://github.com/openoereb/pyramid_oereb/blob/76229720f3393f418de91f6534c136396d2551c5/docker/config.yaml#L98

### DIFF
--- a/docker/config.yaml
+++ b/docker/config.yaml
@@ -92,6 +92,10 @@ vars:
       # Split themes up, so that each sub theme gets its own page
       # Disabled by default.
       split_sub_themes: true
+      # Determine if a multiple page table of content (TOC) is used with a slightly different layout but
+      # better page numbering. If it is known that the TOC is very long and runs over more than one page it
+      # is preferred to set this to true.
+      multi_page_TOC: false
       # Specify any additional URL parameters that the print shall use for WMS calls
       wms_url_params:
         TRANSPARENT: 'true'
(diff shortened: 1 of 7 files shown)


pfirpfel:
Feature should be disabled by default.

I agree, there is still some work to do for the spacing on the new template.

pfirpfel:
Can you add a example output for the full extract for both versions when you're done?

marionb:
Some Examples using the multipage PDF (one reduced and one full):
[statischer_OEREB_Auszug_20190524025255.pdf](https://github.com/openoereb/pyramid_oereb/files/3217075/statischer_OEREB_Auszug_20190524025255.pdf)
[statischer_OEREB_Auszug_20190524025133.pdf](https://github.com/openoereb/pyramid_oereb/files/3217076/statischer_OEREB_Auszug_20190524025133.pdf)

PR #864, opened 2019-05-28
Multiple responsible office

marionb:
WIP for Multiple responsible office

TODO:
- ~~Do not flatten responsible offices to create a list~~
- check that the used example is correct ? 
- fix print proxy to create a list of multiple responsible offices
- Adapt print templates
- Add/correct tests


pfirpfel:
Will use #865 instead

PR #872, opened 2019-06-13
'in'valid XML fields LengthShare, NrOfPoints

lopo977:
To be valid, LengthShare and NrOfPoints elements, have to be prefixed with ns data


pfirpfel:
I removed my commit again. AreaUnit / LengthUnit are part of the `extensions`, were you can put basically everything. The renaming would be more consistent, but it is likely to break existing installations.

PR #885, opened 2019-07-19
flexible print configuration

jwkaltz:
Follow-up to PR #631:
Be able to flexibly choose print configuration for XML2PDF or for MapFish-Print.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,5 +37,6 @@ __pycache__/
 /test-db/12-create.sql
 /test-db/13-fill.sql
 /docker/logo_*.png
+/docker/config.yaml
 .pytest_cache
 .idea
--- a/CONST_vars.yml
+++ b/CONST_vars.yml
@@ -2,7 +2,9 @@ vars:
 
   sqlalchemy_url: '{SQLALCHEMY_URL}'
   png_root_dir: '{PNG_ROOT_DIR}'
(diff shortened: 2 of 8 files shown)


vvmruder:
For me this adds ways too much layers of complexity. Why not simply document the selectable print proxy and thats it.

Every user should be perfectly able to setup this in his pyramid oereb YAML. I don't think this helps anybody.

jwkaltz:
> For me this adds ways too much layers of complexity. Why not simply document the selectable print proxy and thats it.
> 
> Every user should be perfectly able to setup this in his pyramid oereb YAML. I don't think this helps anybody. Since user names and tokens need to be specified anyhow.

This PR is the follow-up to https://github.com/openoereb/pyramid_oereb/pull/631#discussion_r304872016

I am not sure to understand your concern. For the user, nothing changes, s/he simply has an additional option when using the scripts to define which print backend s/he wants to use. If s/he does not use this option, there is a default. I don't see how this adds complexity, isn't this analog to the other configuration options?

vvmruder:
In this comment I clearly point out my suggestions to leave it as integration problem.

I think adding all the possible options a config yaml could look like we will end up in a fairly complex environment. And I want to keep it simple as possible. As I said: At least the credential need to be filled in later by hand.

And the standard configuration should only act as a starting point. All customisation will be done by hand later.

PR #914, opened 2019-10-10
Fix XML issues

### LINKED ISSUE #911: Ampersand not escaped in XML extract
We have a few public law restrictions with an ampersand (&) in the information attribute. in my opinion the ampersand should be escaped by pyramid_oereb (with `&amp;`). This does not happen, therefore we have invalid xml output in these cases. e.g.
https://www.oereb.apps.be.ch/extract/reduced/xml/CH313516468953

vvmruder:
Collects all fixes regarding to the XML issues.


pyramid_oereb/lib/renderer/extract/templates/xml/extract.xml
```diff
@@ -69,10 +69,10 @@
         <data:CantonalLogo>${extract.cantonal_logo.encode()}</data:CantonalLogo>
         <data:MunicipalityLogo>${extract.municipality_logo.encode()}</data:MunicipalityLogo>
     %else:
-        <data:LogoPLRCadastreRef>${request.route_url('{0}/image/logo'.format(route_prefix), logo='oereb')}</data:LogoPLRCadastreRef>
-        <data:FederalLogoRef>${request.route_url('{0}/image/logo'.format(route_prefix), logo='confederation')}</data:FederalLogoRef>
-        <data:CantonalLogoRef>${request.route_url('{0}/image/logo'.format(route_prefix), logo='canton')}</data:CantonalLogoRef>
-        <data:MunicipalityLogoRef>${request.route_url('{0}/image/municipality'.format(route_prefix), fosnr=extract.real_estate.fosnr)}</data:MunicipalityLogoRef>
+        <data:LogoPLRCadastreRef>${request.route_url('{0}/image/logo'.format(route_prefix), logo='oereb') | x}</data:LogoPLRCadastreRef>
+        <data:FederalLogoRef>${request.route_url('{0}/image/logo'.format(route_prefix), logo='confederation') | x}</data:FederalLogoRef>
+        <data:CantonalLogoRef>${request.route_url('{0}/image/logo'.format(route_prefix), logo='canton') | x}</data:CantonalLogoRef>
```
pfirpfel: Where does the value for `x` come from?

vvmruder: Its a mako mechanism to handle strings which need to comply with the desired target format of output:
https://docs.makotemplates.org/en/latest/filtering.html

pfirpfel: Ah ok, thanks a lot!

pyramid_oereb/lib/renderer/extract/templates/xml/extract.xml
```diff
@@ -69,10 +69,10 @@
...
+        <data:MunicipalityLogoRef>${request.route_url('{0}/image/municipality'.format(route_prefix), fosnr=extract.real_estate.fosnr) | x}</data:MunicipalityLogoRef>
```
voisardf: why do we need escaping for fosnr?

tests/renderer/xml/test_view_service.py
```diff
@@ -26,23 +26,38 @@ def test_empty():
 
 def test_reference_wms():
     map = ViewServiceRecord(
-        reference_wms='http://example.com',
+        reference_wms='http://example.com?SERVICE=WMS&REQUEST=GetMap&FORMAT=image/png&SRS=epsg:2056',
```
voisardf: example.com is a registered  domain. 
Wouldn't it be better to use an URI to a server that does not exist, or is under our control? You never know what the owner of this domain decides to send back as an error message besides the fact, that it is probably pretty anoying for them to get all this fake requests and mails... unless they analyze them for marketing purposes. 😉

pyramid_oereb/lib/renderer/extract/templates/xml/extract.xml
vvmruder: Its actually not escaping the fosnr but an url containing the fosnr.

tests/renderer/xml/test_view_service.py
vvmruder: ok ok, I agree... Even when we never call it in the end. But I will change it to a better one.

voisardf: thanks - my bad! not awake enough 🙈

pyramid_oereb/lib/renderer/extract/templates/xml/extract.xml
voisardf: thanks - my bad! not awake enough 🙈

tests/renderer/xml/test_view_service.py
kdeininger: That not necessary. As written on http://example.com/:

> **Example Domain**
>
> This domain is established to be used for illustrative examples in documents. You may use this domain in examples without prior coordination or asking for permission.

https://www.iana.org/domains/reserved

vvmruder: Puhh saved a lot of work :wink:

vvmruder:
If no other requests exists on this PR I will merge.

PR #915, opened 2019-10-10
Provide multilingual OEREB logo

### LINKED ISSUE #913: ÖREB logo not multilingual
the ÖREB logo is - unlike the swisstopo logo - not multilingual. In the pyramid_oereb_standard.yml we can reference only one instance of the logo. In our case we reference the german version. But the german logo is also displayed in extracts in french. So we get the following situation - french extract and a german ÖREB logo:

logo

pyramid_oereb should enable us to reference language dependent versions of the ÖREB logo.

kdeininger:
Closes #913


kdeininger:
**Note:** Old configuration (non-multilingual) still works.

PR #917, opened 2019-10-11
Add file extension in symbol and logo URLs

### LINKED ISSUE #908: Links to Symbols: ".png" missing
Legend Symbols are provided as PNG files (i.e. stored as Base64 images in the DB). However, any file name extension (`.png`) is missing in Xml extracts wich is an error IMHO.

Example:
`
<data:SymbolRef>
http%3A%2F%2Fiap-map.gl.ch%2Foereb%2Fwsgi%2Foereb%2Fimage%2Fsymbol%2FLandUsePlans%2F3%2F1632.71-01-01
</data:SymbolRef>
`

fix:
`
<data:SymbolRef>
http%3A%2F%2Fiap-map.gl.ch%2Foereb%2Fwsgi%2Foereb%2Fimage%2Fsymbol%2FLandUsePlans%2F3%2F1632.71-01-01.png
</data:SymbolRef>
`

kdeininger:
Closes #908


pyramid_oereb/lib/renderer/extract/json_.py
```diff
@@ -111,21 +112,25 @@ def _render(self, extract, param):
                 'LogoPLRCadastreRef': self._request.route_url(
                     '{0}/image/logo'.format(route_prefix),
                     logo='oereb',
-                    language=self._language
+                    language=self._language,
+                    extension=filetype.guess_extension(bytearray(extract.logo_plr_cadastre.content)) or 'png'
```
vvmruder: Why we use default png here?

kdeininger: I could remove the default type and wrap it into a small file type validator. Validation only succeeds for allowed types (png, jpg and svg). Therefore, a matcher for svg files would have to be added.

What do you think?

vvmruder: Sounds more proof. How would you design a svg matcher?

Dockerfile
```diff
@@ -12,6 +12,12 @@ WORKDIR /app
 #    apt-get clean && \
 #    rm --recursive --force /var/lib/apt/lists/*
 
+RUN \
+    apt-get update && \
+    apt-get install --assume-yes --no-install-recommends libffi-dev && \
```
jwkaltz: Why this way, and not via requirements?

Makefile
```diff
@@ -92,7 +92,7 @@ endif
 export SQLALCHEMY_URL = "postgresql://$(PG_CREDENTIALS)@$(@_POSTGIS_IP):5432/pyramid_oereb_test"
 export PNG_ROOT_DIR = pyramid_oereb/standard/
 
-.coverage: $(PYTHON_VENV) $(TESTS_DROP_DB) $(TESTS_SETUP_DB) pyramid_oereb/standard/pyramid_oereb.yml .coveragerc $(shell find -name "*.py" -print)
+.coverage: $(PYTHON_VENV) $(TESTS_DROP_DB) $(TESTS_SETUP_DB) pyramid_oereb/standard/pyramid_oereb.yml .coveragerc $(shell find -name "*.py" -print | fgrep -v /.venv)
```
jwkaltz: Is this change for the point in your comment "exclude directory .venv as dependency in Makefile target .coverage"?
Why is this being excluded, in fact?

pyramid_oereb/lib/records/image.py
```diff
@@ -20,4 +33,113 @@ def encode(self):
...
+        Raises:
+            TypeError: Raised if type is invalid or could not be recognized.
+        """
+
+        try:
+            ft = filetype.guess(obj)
+        except IOError:
+            ft = None
+
+        # Check for SVG content if filetype.guess fails
```
jwkaltz: A question regarding backwards compatibility:
could this change behaviour, if png is not detected for some reason in an installation (whereas the file is really png)?
If so, we should add a note of caution in the migration notes.

Dockerfile
kdeininger: This is not a Python package. It is related to this error:
https://travis-ci.org/openoereb/pyramid_oereb/jobs/598100771#L1284

Makefile
kdeininger: Somehow, running `make clean-all checks` the target for `.coverage` failed because of a Python file of an installed package, that has been registered by the `find` call, but did not exist anymore.

My theory:
Running make, it first evaluates the dependencies between the single targets, also calling `find` in this case. This generates a list with all Python files of the installed packages at each start of a run. Then the target `clean-all` removes all this packages. The packages are installed  again as a dependency of the `checks` target, but maybe some versions of dependent packages could have changed, which may result in different Python files as registered at start. This leads to errors, like
```
No rule to make target '.venv/lib/python3.6/site-packages/some-package/some-file.py', needed by '.coverage'.  Stop.
```

That is why I think, the folder `.venv` has to be excluded.

pyramid_oereb/lib/records/image.py
kdeininger: Yes, you are right. It is a kind of restriction that affects backwards compatibility, if someone uses other image formats than PNG or SVG. We should add something in the migration notes.

Dockerfile
jwkaltz: But it seems this error has nothing to do with your changes? I saw this error also in PR#918, and I also had an error with libffi when building master locally, independently of these PRs. But they seem to be random. When I simply restarted the Travis Job on #918, the error disappeared, and locally, when I clean up some docker images, the error also disappears.

I think it would be better to do a separate issue & PR for this libffi problem, so that your PR contains only the changes related to the functional change. I assume your PR will pass Travis independently if this (if necessary, after 1-2 restarts).

pyramid_oereb/lib/records/image.py
kdeininger: I have added some migration notes with a warning concerning the image formats:
https://github.com/openoereb/pyramid_oereb/pull/917/commits/9b1c910d671fbc1fb1a9d40dc3155b8f82c63f4f#diff-f86adfaa9b2056d4a7b2d6624899961dR10-R23

Dockerfile
kdeininger: Okay, I will undo this change and try it again. :+1:

pyramid_oereb/lib/records/image.py
jwkaltz: Thanks

Makefile
vvmruder: A proof would be if `make clean-all` followed by a separate `make checks` produces the same result as @kdeininger 's  change. Since the problem is coming up in really rare situations (like changing sub-sub package dependencies) we probably are not able to reproduce the case easily. Anyway... Excluding `.venv` directory from coverage report generation looks like a good idea.

PR #918, opened 2019-10-15
added CertificationAtWeb to the multilingual params to parse

voisardf:
As the title says, the CertificationAtWeb is added to the parameters to be parsed as multilingual dict for the mfp print proxy.
This is needed to fix issue https://github.com/openoereb/pyramid_oereb_mfp/issues/29

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -232,6 +232,7 @@ class Renderer(JsonRenderer):
         self._multilingual_m_text(extract_dict, 'GeneralInformation')
         self._multilingual_m_text(extract_dict, 'BaseData')
         self._multilingual_m_text(extract_dict, 'Certification')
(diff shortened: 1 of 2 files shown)


jwkaltz:
> NOTE:
> I could not reproduce the test failure locally because of the python version.

I will try to remove that evil error-hiding workaround in a separate PR -> PR#919. 

@voisardf so, to finish this PR, as Marion stated, we think the expected json should be changed (flat structure for certificationatweb in the print proxy json, because it now gets flattened)

voisardf:
yes, I forgot about the tests and then I was in a meeting with the boss 😉 
It should be ok now. Unless there are other test cases I have to adapt...

jwkaltz:
> It should be ok now

If the content change from "https://oereb.bl.ch/certification/de" to "Beglaubigung nach: https://oereb.bl.ch/certification/de" is intentional, you need to change the test data, too.

voisardf:
thanks, merging

PR #932, opened 2019-10-31
synchronize configuration files of docker and non-docker world

jwkaltz:
synchronize standard/default configuration files of docker and non-docker world.
It should make future evolutions and understanding of configurations a bit easier

(there are still some differences regarding Mako variable usage that can probably be harmonized, but I did not have time within this PR to do that)


jwkaltz:
> LGTM 
> 
> But wouldn't it be better to create a make target to render the docker configuration using the standard configuration mako on every `make serve`?

Good idea @kdeininger ! I will do that in a later PR 
(this PR then is a preparation for that goal, since the configuration is almost identical now)

PR #939, opened 2019-11-11
fix xml encoding, fixes #926

### LINKED ISSUE #926: Encoding of & in XML Extract
The presence of the character **&** in the columns **type_code** or **type_code_list** on table **LegendEntry** and **PublicLawRestriction** prevents the creation of a valid XML-Extract.

This character should be encoded as **&amp;** in the whole XML-Extract.

Thanks.

vvmruder:
See Issue #926 for further details.


vvmruder:
I tried to encode almost all textual information in XML since it could contain possibly bad characters. Does anyone see a problem with this?

PR #950, opened 2019-11-22
Add fallback language for pdf extract

### LINKED ISSUE #924: The defined default language is not taken into account by mfp
The generation of the PDF-extract with MFP fails if not all elements are present in the requestet language, whereas the generation of json an xml works as expected in these cases.  
If a multilingual element (exp. Titel) of an extract is not present in the requested language, the json and xml are generated taking the default language into account.  The pdf generation fails in these cases.

marionb:
This PR fixes #924. It adds a fallback language for the pdf extract when requesting the extract in another language that what is set to default in the config. For the fallback language the default is taken.
You can test by adding for example `lang=fr`in the GET parameters of the URL.

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -67,6 +67,7 @@ class Renderer(JsonRenderer):
         # If language present in request, use that. Otherwise, keep language from base class
         if 'lang' in self._request.GET:
             self._language = self._request.GET.get('lang')
+            self._fallback_language = Config.get('default_language')
 
         # Based on extract record and webservice parameter, render the extract data as JSON
         extract_record = value[0]
@@ -576,7 +577,10 @@ class Renderer(JsonRenderer):
     def _multilingual_text(self, parent, name):
         if name in parent:
             lang_obj = dict([(e['Language'], e['Text']) for e in parent[name]])
-            parent[name] = lang_obj[self._language]
+            if self._language in lang_obj.keys():
+                parent[name] = lang_obj[self._language]
+            else:
+                parent[name] = lang_obj[self._fallback_language]
 
     def _sort_sub_themes(self, restrictions):
         # split restrictions by theme codes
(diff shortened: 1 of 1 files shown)


pfirpfel:


```diff

Clones removed
==============
+ pyramid_oereb/contrib/print_proxy/mapfish_print.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/pyramid_oereb/pyramid_oereb/pullRequest?prid=4547450&bid=15299115)

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -65,8 +65,9 @@ def __call__(self, value, system):
 
         self._request = self.get_request(system)
         # If language present in request, use that. Otherwise, keep language from base class
-        if 'lang' in self._request.GET:
-            self._language = self._request.GET.get('lang')
+        if 'lang' in self._request.GET or 'LANG' in self._request.GET:
```
jwkaltz: Please use same logic as in pyramid_oereb/views/webservice,py line 751
language = str(self._params.get('LANG')).lower()

Maybe we should even isolate that logic to a function that can be used in both places, webservice and print_proxy?

pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -574,9 +575,13 @@ def _multilingual_m_text(self, parent, name):
         self._multilingual_text(parent, name)
 
     def _multilingual_text(self, parent, name):
+        fallback_language = Config.get('default_language')
```
jwkaltz: We should not do it like that, because this way the Configuration will be read for every element of text (it should be read only once).

pyramid_oereb/contrib/print_proxy/mapfish_print.py
danduk82: I think it is more a problem of esacping the case of the parameter in the request. [This solution](https://stackoverflow.com/questions/764235/dictionary-to-lowercase-in-python) might work, you must check all the keys to lowercase : `dict((k.lower(), v) for k,v in {'My Key':'My Value'}.iteritems())`

PR #954, opened 2019-11-25
Avoid connection error

### LINKED ISSUE #953: Occasional error "psycopg2.OperationalError: server closed the connection unexpectedly"
Sometimes, a user request can fail with the following error: "psycopg2.OperationalError: server closed the connection unexpectedly"

Reproduced on a local test setup (with the dummy database setup of pyramid_oereb) as follows:
* use Postgres 10
* do some requests
* restart db container: docker restart pyramidoereb_oereb-db_1
* do several new requests (the one using the connection which was already used before the db restart will fail)

jwkaltz:
Fixes #953 

If a connection has not been used in the last 30 seconds, recycle it before using it for a new request to the database.

Tested successfully locally (see instructions in #953)


jwkaltz:
Will redo this PR - including unintended changes by accident

PR #962, opened 2019-12-05
Complete comment for buffer setting.

ger-benjamin:
To understand better what this value means regarding the print system you use.

And as "10px" it's the default value of MapFishPrint, I come back again from 30 to 10. But let me know if you want to set 30px as default. (Is there an official specification about the buffer ? I haven't seen one)

### DIFF
--- a/docker/config.yaml.mako
+++ b/docker/config.yaml.mako
@@ -74,15 +74,17 @@ vars:
       use_wms: "true"
       validate: "false"
       # The following parameters are not used by xml2pdf, but might be in the fure (see issue #873)
-      buffer: 30
+      buffer: 10
       basic_map_size: [493, 280]
       pdf_dpi: 300
       pdf_map_size_millimeters: [174, 99]
   % else:
       # Configuration for MapFish-Print print service
       renderer: pyramid_oereb.contrib.print_proxy.mapfish_print.Renderer
-      # The minimum buffer in pixel at 72 DPI between the real estate and the map's border.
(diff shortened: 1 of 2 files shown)


docker/config.yaml.mako
```diff
@@ -74,15 +74,17 @@ vars:
...
       basic_map_size: [493, 280]
       pdf_dpi: 300
       pdf_map_size_millimeters: [174, 99]
   % else:
       # Configuration for MapFish-Print print service
       renderer: pyramid_oereb.contrib.print_proxy.mapfish_print.Renderer
-      # The minimum buffer in pixel at 72 DPI between the real estate and the map's border.
-      buffer: 30
+      # The minimum buffer in pixel at 72 DPI between the real estate and the map's border. If your print
+      # system draw a margin around the feature (the real estate), you have to set your buffer
```
jwkaltz: draw -> draws

docker/config.yaml.mako
```diff
@@ -74,15 +74,17 @@ vars:
...
+      # here accordingly. MapFishPrint set per default a (Map -> ZoomToFeatures ->) MinMargin to 10px.
```
jwkaltz: set -> sets

But I think we should not put this here, but instead add an explicit configuration line in the pyramid_oereb_mfp.

ger-benjamin: Ok. I'll move this part in pyramid_oereb_mfp

PR #979, opened 2020-01-03
Add plr sorting hook

jwkaltz:
This PR introduces a new hook, to be able to define a sorting of the plrs within the themes in the extract data.
It also proposes a contrib method for one specific sorting implementation, by type_code, as requested by @romefi . This contrib method can be tested by defining in the project configuration file the following:

sort_within_themes_method: pyramid_oereb.contrib.plr_sort_within_themes_by_type_code

(see example in new standard configuration file)

C2C reference ticket: GEO-2844


pyramid_oereb/standard/pyramid_oereb.yml.mako
```diff
@@ -385,6 +385,9 @@ pyramid_oereb:
...
         it: Il contenuto del Catasto RDPP si considera noto. Il Canton ---NOME DEL CANTON--- non può essere ritenuto responsabile per la precisione e l'affidabilità dei documenti legislativi in formato elettronico. L'estratto ha carattere puramente informativo e non è in particolare costituti-vo di diritti e obblighi. Sono considerati giuridicamente vincolanti i documenti approvati o pubblicati passati in giudicato. Con l'autenticazione dell'estratto viene confermata la conformità dell'estratto rispetto al Catasto RDPP al momento della sua redazione.
         rm: ...
+    sort_within_themes_method: pyramid_oereb.standard.hook_methods.plr_sort_within_themes
+    # Example of a specific sorting method:
+    #sort_within_themes_method: pyramid_oereb.contrib.plr_sort_within_themes_by_type_code
```
ger-benjamin:
```suggestion
    # sort_within_themes_method: pyramid_oereb.contrib.plr_sort_within_themes_by_type_code
```

pyramid_oereb/standard/hook_methods.py
```diff
@@ -14,6 +15,9 @@
 from pyramid_oereb.lib.records.office import OfficeRecord
 
 
+log = logging.getLogger(__name__)
```
ger-benjamin: you add it but it's not used

pyramid_oereb/contrib/__init__.py
```diff
@@ -4,3 +4,54 @@
 This package is meant for contributions to the project, that are not part of the core libraries but may be
 useful in general.
 """
+import logging
+
```
ger-benjamin: Can you remove this empty line ?

pyramid_oereb/contrib/__init__.py
```diff
@@ -4,3 +4,54 @@
...
+from functools import cmp_to_key
+
```
ger-benjamin: Can you remove this empty line ?

docker/config.yaml.mako
```diff
@@ -386,6 +386,9 @@ vars:
...
           it: Il contenuto del Catasto RDPP si considera noto. Il Canton ---NOME DEL CANTON--- non può essere ritenuto responsabile per la precisione e l'affidabilità dei documenti legislativi in formato elettronico. L'estratto ha carattere puramente informativo e non è in particolare costituti-vo di diritti e obblighi. Sono considerati giuridicamente vincolanti i documenti approvati o pubblicati passati in giudicato. Con l'autenticazione dell'estratto viene confermata la conformità dell'estratto rispetto al Catasto RDPP al momento della sua redazione.
           rm: ...
+      sort_within_themes_method: pyramid_oereb.standard.hook_methods.plr_sort_within_themes
+      # Example of a specific sorting method:
+      #sort_within_themes_method: pyramid_oereb.contrib.plr_sort_within_themes_by_type_code
```
ger-benjamin:
```suggestion
      # sort_within_themes_method: pyramid_oereb.contrib.plr_sort_within_themes_by_type_code
```

pyramid_oereb/standard/hook_methods.py
ger-benjamin: it is not the same as (use lamda function)
`real_estate.public_law_restrictions = sorted(real_estate.public_law_restrictions, key=lambda x: x.get('type_code', ''))` (whithout the whole `cmp` function ?)

In all cases, with this code and these names: `['Arboretum', 'Ärboretum', 'Belize']` you will get this result:
``['Arboretum', 'Belize', 'Ärboretum']``
Because Ä > Z. It is the wanted result ?

pyramid_oereb/contrib/__init__.py
```diff
@@ -4,3 +4,54 @@
...
+                    ret = 0
+                else:
+                    ret = 1
+            else:
+                ret = 0
+        else:
+            ret = 0
+        return ret
+
+    real_estate.public_law_restrictions = sorted(real_estate.public_law_restrictions, key=cmp_to_key(cmp))
```
ger-benjamin: it is not the same as (use lamda function):
`real_estate.public_law_restrictions = sorted(real_estate.public_law_restrictions, key=lambda x: x.get('type_code', ''))`(whithout the whole `cmp` function ?)

In all cases, with this code and these names: `['Arboretum', 'Ärboretum', 'Belize']` you will get this result:
`['Arboretum', 'Belize', 'Ärboretum']`
Because Ä > Z. It is the wanted result ?

jwkaltz: I don't think we can use the simple solution in this case, because the goal is not to sort by type_code, but to keep the ordering of themes, and within the themes, sort by type_code. I couldn't find a way to do that without a custom comparator function - WDYT?

pyramid_oereb/standard/hook_methods.py
jwkaltz: removed

pyramid_oereb/contrib/__init__.py
ger-benjamin: Oh, I see. Ok. So I've not immediat better solution.
And what about umlaut ? it is correct like that ?

pyramid_oereb/contrib/__init__.py
jwkaltz: removed

pyramid_oereb/contrib/__init__.py
jwkaltz: removed

pyramid_oereb/contrib/__init__.py
jwkaltz: There are different standards for that... for me, the way Python does it (consider Ä, Ö, ... as separate letters and therefore sort them after Z) is OK.

PR #982, opened 2020-01-07
Add possibility to archive generated pdf (by MFP)

ger-benjamin:
https://jira.camptocamp.com/browse/GEO-2823

The archived (stored, there is no compression or anything) file will have this name for instance:
`07-01-2020_10-11-17_CH113928077734.pdf`
(`day-month-year_hour-minute-seconds_egrid.pdf`)

As demanded, the timezone used for the filename is 'UTC+2'. That means the summer CH time.


ger-benjamin:
@jwkaltz done

PR #984, opened 2020-01-08
Optional cache for Oereblex integration

jwkaltz:
One pyramid_oereb user, who uses Oereblex, has analyzed that most of the time spent in generating the data extracts is spent in requesting data from Oereblex, whereas the same data is requested many times, even within the same data extract.

This pull request proposes a cache for the oereblex records, which can be used optionally. This has been tested for one (real) pyramid_oereb project, on an average data extract:
* without cache, the extract is consistently generated within 5 seconds,
* with cache activated, the extract is generated in 3 seconds on the first call, and 1.5 seconds on subsequent calls.

The more (repeating) geolinks an oereb service has, the more the performance improvement will be relevant.

Note: if the cache approach is validated and found useful, more subtle cache mechanisms might be implemented in the future (notably, an expiry date for cache entries, because otherwise the Oereblex data will always stay the same as long as the oereb server is not restarted).

C2C reference: GEO-2686

### DIFF
--- a/pyramid_oereb/contrib/sources/document.py
+++ b/pyramid_oereb/contrib/sources/document.py
@@ -76,8 +76,6 @@ class OEREBlexSource(Base):
             params (pyramid_oereb.views.webservice.Parameter): The parameters of the extract request.
             geolink_id (int): The geoLink ID.
         """
-        log.debug("read() start")
-
         # Request documents
         url = '{host}/api/{version}geolinks/{id}.xml'.format(
             host=self._parser.host_url,
@@ -88,7 +86,7 @@ class OEREBlexSource(Base):
         request_params = {
             'locale': language
         }
-        log.debug("read() getting documents, url: {}, parser: {}".format(url, self._parser))
+        log.debug("read() getting documents for geolink_id {}, url: {}, parser: {}".format(geolink_id, url, self._parser))
         documents = self._parser.from_url(url, request_params, proxies=self._proxies, auth=self._auth)
         log.debug("read() got documents")
 
--- a/pyramid_oereb/contrib/sources/document.py
+++ b/pyramid_oereb/contrib/sources/document.py
@@ -86,7 +86,8 @@ class OEREBlexSource(Base):
         request_params = {
             'locale': language
         }
-        log.debug("read() getting documents for geolink_id {}, url: {}, parser: {}".format(geolink_id, url, self._parser))
+        log.debug("read() getting documents for geolink_id {}, url: {}, parser: {}"
+                  .format(geolink_id, url, self._parser))
         documents = self._parser.from_url(url, request_params, proxies=self._proxies, auth=self._auth)
         log.debug("read() got documents")
 
--- a/pyramid_oereb/contrib/sources/oereblex_cache.py
+++ b/pyramid_oereb/contrib/sources/oereblex_cache.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+import logging
+
+log = logging.getLogger(__name__)
+
+
+class OEREBlexCache(object):
+    """
+    A cache for Oereblex records
+    """
+    _cache_dict = {}
+
+    @staticmethod
+    def get_records(lexlink):
+        """
+        Returns the records for a specified lexlink, if they are in the cache
+        """
+        records = OEREBlexCache._cache_dict.get(lexlink)
+        if records:
+            log.debug("CACHE HIT on lexlink {} return {} records from cache".format(lexlink, len(records)))
+        else:
+            log.debug("CACHE MISS on lexlink {}".format(lexlink))
+        return records
+
+    @staticmethod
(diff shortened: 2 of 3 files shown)


kdeininger:
This caching solution might be dangerous as the cache grows until the server is restarted. There is no expiration time set for the cached values.

We already thought about caching too. Maybe a more generic method, e.g. using pyramid_beaker, could be used for caching (OEREBlex, other sources or even a whole extract).

vvmruder:
I think instead shifting the real problem a bit we should think about a general change. The standard model supports documents stored in database. So probably the most sustainable solution would be a dataintegrationstep (like the one with the federal data) to download all documents on data deploy moment and store them in the database. This way we also ship around the problem of diverging states between oereb lex and pyramid_oereb instance. It happens from time to time that documents in oereb lex are deleted while they are still 'lex linked' in pyramid_oereb.

This could be done by implementing an integration step similar the federal one which utilizes the lexlink formatter library.

jwkaltz:
> (...)
> We already thought about caching too. Maybe a more generic method, e.g. using pyramid_beaker, could be used for caching (OEREBlex, other sources or even a whole extract).

I have thought about that, I see an advantage of an application cache in that all the processing needs to be done only once, XML parsing etc. - I did not benchmark the difference regarding caching only the XML data and caching the processed records, though. However, we have a similar approach in GMF, and the performance improvement is huge.

MaxHoeferGEOINFO:
Hi all - Max here from Geoinfo, we host the OEREB Portal for Kanton AI and AR.
The refresh idea we would have to discuss with the Kanton, especially in regards of refresh rate. No refresh at all is not good, file size wise and also we completely loose the agility that the OEREBLex offers. 
On the other hand, I'm not sure how big the benefit is, if we refresh the cache too often. Also, I believe the OEREB Portal is probably used once per user, for one specific property. So geolinks will/might be different for each usage, and then the cache will help for duplicated geolinks only. 
I will discuss with the Kanton.

I think the idea mentioned from vvmruder is actually a workaround around the problem itself.
OEREBLex would be redundant then. 


Changing pyramid_oereb towards a parallel querry of geolinks is too much effort? 
Cheers
Max

MaxHoeferGEOINFO:
Hi all

I talked with Kanton AI - since the most use cases will be that one user requests information for one property, as mentioned above, the cache as I understood it here will not help a lot unfortunately, since the first request is the slow one.  
What would help though is, if we would have an option to cache all geolinks, during the night for example? 

Also, documents stored in the oereblex will only have 15 changes per year approximately, so the agility I mentioned beforehand would not be necessary anymore.

Cheers
Max

jwkaltz:
> I talked with Kanton AI - since the most use cases will be that one user requests information for one property, as mentioned above, the cache as I understood it here will not help a lot unfortunately, since the first request is the slow one.
(...)

In my opinion, the cache is a very significant performance improvement: since the typical data extract uses several times the same geolinks, even for the very first call after server start, the performance is much better than without a cache.
Plus, even if when new parcels are selected, I assume they will typically have common Geolinks with parcels that have been selected before (since the plr / geolinks often cover a wide surface).
I recommend you try it out in your project and see what the impact is.

MaxHoeferGEOINFO:
Hi Wolfgang

since this does not influence us only, I want to talk to the User-Group first and see what they say. 
I will come back once I know more.

Cheers
Max

jwkaltz:
Hi Max,

it influences only those who use the new class, this is completely optional. All others are not influenced.

MaxHoeferGEOINFO:
Hi Wolfgang

yes, but from the "bad performance" all are affected. So if we find a solution, this solution should be looked at from more users I believe.

Cheers
Max

pfirpfel:


```diff

Complexity increasing per file
==============================
- pyramid_oereb/contrib/sources/oereblex_cache.py  2
- pyramid_oereb/contrib/sources/plr_oereblex.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/pyramid_oereb/pyramid_oereb/pullRequest?prid=4784748&bid=15936069)

jwkaltz:
Hi @MaxHoeferGEOINFO and all,

for your information, with @ralphschulerSZ , we have done a benchmark using the 100 parcels in canton Schwyz that Ralph typically uses for automated tests. In the attached list, you see the impact on performance at Schwyz; this measures the time that a client waits for a JSON data extract (full time, including network transfer), once with the new cache activated and once without:
[oereb_cache_perfs_results.xlsx](https://github.com/openoereb/pyramid_oereb/files/4076903/oereb_cache_perfs_results.xlsx)

At the bottom of the list some statistics are computed: 
- the average data extract is 2 times faster with this PR. 
- if we look specifically at extracts that currently take a long time (> 8 seconds), the average for these is now more than 2,5 times faster, going down from 10.7 to 4 seconds.

jwkaltz:
See PR #993 instead

PR #993, opened 2020-01-24
use per topic -store- to verify already queried documents

### LINKED ISSUE #992: Poor performance for OEREBlex queries
It was reported that OEREBlex API calls are slow. This means that time from request to response lasts regularly more then 5 seconds which is indeed not satisfying.

@MaxHoeferGEOINFO & @jwkaltz could you please add some details of your statistics here?

vvmruder:
fixes #992

### DIFF
--- a/pyramid_oereb/contrib/sources/plr_oereblex.py
+++ b/pyramid_oereb/contrib/sources/plr_oereblex.py
@@ -35,6 +35,7 @@ class DatabaseOEREBlexSource(DatabaseSource):
         """
         super(DatabaseOEREBlexSource, self).__init__(**kwargs)
         self._oereblex_source = OEREBlexSource(**Config.get_oereblex_config())
+        self._queried_lexlinks = {}
 
     def get_document_records(self, params, public_law_restriction_from_db):
         """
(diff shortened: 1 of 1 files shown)


vvmruder:
@MaxHoeferGEOINFO & @jwkaltz & @kdeininger could you please test if this is working?

pfirpfel:


```diff

Complexity increasing per file
==============================
- pyramid_oereb/contrib/sources/plr_oereblex.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/pyramid_oereb/pyramid_oereb/pullRequest?prid=4880381&bid=16171273)

vvmruder:
This could be optimized to a global store of documents. But I think it is quite unusual that law documents are shared between topics.

jwkaltz:
Here are some results of performance tests, focusing on the most problematic extracts currently at Schwyz: 
* For 8 extracts currently taking on average 10 seconds, the average goes down to 3 seconds
* The maximum time goes down from 14 seconds to 4 seconds.

There is no significant difference with the approach proposed in #984, in fact, in most cases, this approach here is a bit faster than the one #984. I assume it depends on how close the tested extracts are (I assume neighbouring parcels will tend to share some geolinks).

In any case, because this PR has no drawbacks at all and has a huge performance benefit, I propose to merge this one as standard behavior now.

MaxHoeferGEOINFO:
Hi all

I have also done some tests yesterday, I just wanted to confirm that I included the changes correctly in my custom code, before I share the results here. The results can be find attached. 
[OEREB Test 20200127.xlsx](https://github.com/openoereb/pyramid_oereb/files/4120909/OEREB.Test.20200127.xlsx)

I also see a very good improvement - requests which took 9 seconds are now down to 3, and our biggest property in AI went from 44 seconds down to 6. 

Thanks a lot for the efforts here, I think this is very good already. 
Cheers
Max

PR #1005, opened 2020-02-11
getegrid: set output_format in case of exception handling

jwkaltz:


jwkaltz:
@danduk82 this fixes a small regression introduced with the stats refactoring, please check this PR

PR #1006, opened 2020-02-14
Make geometry inclusion for pdf calls configurable

jwkaltz:
Make geometry inclusion for pdf calls configurable.

Follow-up to PR #883. For complex/big parcels, the print payload becomes very large when all geometry data is included; but in fact, for the MapFish Print this is not necessary. 

This PR has been tested at Schwyz, using pyramid_oereb v1.7.1.

### DIFF
--- a/pyramid_oereb/views/webservice.py
+++ b/pyramid_oereb/views/webservice.py
@@ -339,10 +339,11 @@ class PlrWebservice(object):
         if extract_format == 'pdf' and user_requested_geometry:
             raise HTTPBadRequest('Geometry is not available for format PDF.')
 
-        # If PDF is to be produced, always include geometry
-        # (even though the URL shall not contain geometry parameter)
+        # If PDF is to be produced, check if geometry should be included
+        # (even though the URL is not allowed to not contain geometry parameter)
         if extract_format == 'pdf':
-            with_geometry = True
+            if Config.get('print', {}).get('with_geometry', True):
+                with_geometry = True
 
         # With images?
         with_images = self._params.get('WITHIMAGES') is not None
--- a/pyramid_oereb/views/webservice.py
+++ b/pyramid_oereb/views/webservice.py
@@ -340,10 +340,10 @@ class PlrWebservice(object):
             raise HTTPBadRequest('Geometry is not available for format PDF.')
 
         # If PDF is to be produced, check if geometry should be included
-        # (even though the URL is not allowed to not contain geometry parameter)
+        # (this override can be needed for the print service. Note that, to be compliant to specification,
+        # the URL for a PDF request should not contain the geometry parameter)
         if extract_format == 'pdf':
-            if Config.get('print', {}).get('with_geometry', True):
-                with_geometry = True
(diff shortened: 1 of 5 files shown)


pfirpfel:


```diff

Complexity increasing per file
==============================
- pyramid_oereb/views/webservice.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/pyramid_oereb/pyramid_oereb/pullRequest?prid=5004639&bid=16511107)

pfirpfel:


```diff

Complexity increasing per file
==============================
- pyramid_oereb/views/webservice.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/pyramid_oereb/pyramid_oereb/pullRequest?prid=5004639&bid=16511107)

pyramid_oereb/views/webservice.py
```diff
@@ -339,10 +339,11 @@ def __validate_extract_params__(self):
...
             raise HTTPBadRequest('Geometry is not available for format PDF.')
 
-        # If PDF is to be produced, always include geometry
-        # (even though the URL shall not contain geometry parameter)
+        # If PDF is to be produced, check if geometry should be included
+        # (even though the URL is not allowed to not contain geometry parameter)
         if extract_format == 'pdf':
-            with_geometry = True
+            if Config.get('print', {}).get('with_geometry', True):
+                with_geometry = True
```
marionb: This saves you some writing and is I think easier to read.
```suggestion
            with_geometry = Config.get('print', {}).get('with_geometry', True) or with_geometry
```
(by the way if in the config `with_geometry = False' the value for `with_geometry` here can still be `True` depending on what was set above. I don't know if this is intended?)

jwkaltz:
> I think the geometry is necessary for the PDF extract to draw the real estate highlight. But you are right, this might result in a lot of data overhead as the geometries of the public law restrictions are unused.
> 
> Maybe we could adapt the Mapfish Print renderer to ensure that only the real estate geometry is included to reduce data.

Yes, maybe that would be an option, too -
but if you don't mind, I would like to do this PR first, because this way, each project has the option to go back to the way MapFish Print was called in pyramid_oereb version 1.4. In the case of Schwyz, for one of the parcels, the payload otherwise increases from 80k to 220 MB (280 times bigger)...
Since the behaviour is made configurable, each project can choose whether it wants to way it was done in pyramid_oereb 1.4, or whether it wants the default way since pyramid_oereb 1.5.

pyramid_oereb/views/webservice.py
jwkaltz: Thanks, I have changed it.

Regarding your question: the "geometry" option is somewhat confusing, because it is allowed for XML/JSON calls, but not for PDFs. But for some PDF calls (xml2pdf), geometry data is needed.
I have updated the comment in the code, hopefully it is clearer like this.

pfirpfel:


```diff

Complexity increasing per file
==============================
- pyramid_oereb/views/webservice.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/pyramid_oereb/pyramid_oereb/pullRequest?prid=5004639&bid=16511107)

PR #1008, opened 2020-02-18
Debug coordinate transformations for GEO-2924

marionb:
add debug logging for coordinate transformations for the egrid request.

### DIFF
--- a/pyramid_oereb/views/webservice.py
+++ b/pyramid_oereb/views/webservice.py
@@ -420,10 +420,15 @@ class PlrWebservice(object):
             shapely.geometry.Point or shapely.geometry.Polygon: The transformed coordinates as
             Point.
         """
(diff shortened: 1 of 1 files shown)


jwkaltz:
This would already for useful for debugging, but I would prefer to see the actual input parameters to the external function in the logs

(I realize that, in this case, it will simply be "epsg:" concatenated with what you are proposing to log - but I think this would make it easier to read the logs (and quicker to understand what the external function will do))

marionb:
@jwkaltz I think we can close this PR. From what I saw the issue is solved? Or do we sill need this?

jwkaltz:
> @jwkaltz I think we can close this PR. From what I saw the issue is solved? Or do we sill need this?

these debug statements can be useful in another case (when there is an actual transformation happening), so I suggest we merge the PR

PR #1027, opened 2020-04-22
Doc: user group information

jwkaltz:
Add mechanism to be able to add any files to the doc; add user group short information and slides


jwkaltz:
@vvmruder do you want to add more slides now from the last user group? (it can be done later as well, of course)

vvmruder:
In the moment its ok for me. I think its a good idea to gather this in the docs. Thank you for that.

Issue #1013, opened 2020-03-09
Shapely 1.7.0 error on intersect with invalid geometries


ger-benjamin:
Hi,

It concerns pyramid_oereb master or dev build after: https://github.com/openoereb/pyramid_oereb/pull/1003

We have moved to `Shapely 1.7.0`. I've tried to update a pyramid_oereb with it but it looks crashing on big parcels with this error:

```
TopologicalError: The operation 'GEOSIntersection_r' could not be performed. Likely cause is invalidity of the geometry <shapely.geometry.polygon.Polygon object at 0x7f8007b1f050>
```

Here: `File .../oereb/.venv/lib/python2.7/site-packages/pyramid_oereb/lib/records/geometry.py", line 134.`

Perhaps I get this error because I still use python 2.7 in the project ? **Can someone does some tests on big parcels to see if the error happens too ?**
I wonder if it is always happening or only without python 3

If nobody can reproduce the issue, I'll try to export the polygon to test on my side and maybe open an issue in the Shapely repo.

```
.../oereb/.venv/lib/python2.7/site-packages/pyramid_oereb/lib/records/view_service.py:119: UserWarning: Type of "layer_opacity" should be "float"
  warnings.warn('Type of "layer_opacity" should be "float"')
ERROR shapely.geos TopologyException: Input geom 0 is invalid: Ring Self-intersection at or near point 2699116.8500000001 1208887.851 at 2699116.8500000001 1208887.851
mod_wsgi (pid=9903): Exception occurred processing WSGI script '.../oereb/apache/application.wsgi'.
Traceback (most recent call last):
  File .../oereb/.venv/lib/python2.7/site-packages/pyramid/router.py", line 270, in __call__
    response = self.execution_policy(environ, self)
  File .../oereb/.venv/lib/python2.7/site-packages/pyramid/router.py", line 279, in default_execution_policy
    return request.invoke_exception_view(reraise=True)
  File .../oereb/.venv/lib/python2.7/site-packages/pyramid/view.py", line 778, in invoke_exception_view
    reraise_(*exc_info)
  File .../oereb/.venv/lib/python2.7/site-packages/pyramid/router.py", line 277, in default_execution_policy
    return router.invoke_request(request)
  File .../oereb/.venv/lib/python2.7/site-packages/pyramid/router.py", line 249, in invoke_request
    response = handle_request(request)
[...]

PR #1032, opened 2020-04-23
Add more exception handling for geometry intersection

jwkaltz:
To be able to quickly identify the data which is problematic for Shapely V1.7 (see issue #1013 ), add exception handling and more logging.

### DIFF
--- a/pyramid_oereb/lib/records/plr.py
+++ b/pyramid_oereb/lib/records/plr.py
@@ -180,13 +180,18 @@ class PlrRecord(EmptyPlrRecord):
         tested_geometries = []
         inside = False
         for geometry in self.geometries:
-            if geometry.calculate(
-                    real_estate,
-                    self.min_length, self.min_area,
-                    self.length_unit, self.area_unit
-            ):
-                tested_geometries.append(geometry)
-                inside = True
+            try:
+                if geometry.calculate(
+                        real_estate,
+                        self.min_length, self.min_area,
+                        self.length_unit, self.area_unit
+                ):
+                    tested_geometries.append(geometry)
+                    inside = True
+            except Exception as ex:
+                log.info("PLR geometry check not possible, plr is {}, geometry.geom is {}"
+                         .format(self, geometry.geom))
+                raise ex
         self.geometries = tested_geometries
 
         # Points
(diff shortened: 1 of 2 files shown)


pfirpfel:


```diff

Complexity increasing per file
==============================
- pyramid_oereb/lib/records/geometry.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/pyramid_oereb/pyramid_oereb/pullRequest?prid=5406462&bid=17723306)

pyramid_oereb/lib/records/plr.py
```diff
@@ -180,13 +180,18 @@ def calculate(self, real_estate):
...
+            try:
+                if geometry.calculate(
+                        real_estate,
+                        self.min_length, self.min_area,
+                        self.length_unit, self.area_unit
+                ):
+                    tested_geometries.append(geometry)
+                    inside = True
+            except Exception as ex:
+                log.info("PLR geometry check not possible, plr is {}, geometry.geom is {}"
```
kdeininger, line 192: Why is this logged using `info()` instead of `error()`?

jwkaltz: Thanks for having a look @kdeininger , in fact, I am not really happy with this PR, I was trying to output ids of the involved objects, but did not succeed; the real estate is clear, but I am not sure from where to get a meaningful key regarding the geometry being intersected with the real estate geometry, to make it easier to look it up in the data source. Maybe you have a better idea.

kdeininger: You could try to calculate a point inside the invalid geometry:
https://shapely.readthedocs.io/en/maint-1.6/manual.html#object.representative_point

Using the point and the topic of the invalid PLR, it should be possible to identify the feature.

vvmruder:
As mentioned in #1111 I think we not handle generally broken geometries in calculation. This should be done by simple (to be defined checks) on input into DB. PostGIS offers a lot of geometry health check functionality.

Indeed we should be able to give a more specific info about geometrically errors. Therefor I like @kdeininger suggestion. Instead of calculating a point in a broken geometry (which might fail too) we could simply use a point out of the geometry itself. Maybe a simple serialized linestring. This seems more robust to me.

PR #1033, opened 2020-04-23

jwkaltz:


jwkaltz:
@kdeininger is it worthwhile to do a new maintenance release of pyramid_oereb with the new version of geolink_formatter (in time for the user group)? Or should we wait a bit?

PR #1034, opened 2020-04-24
Configurable geoLink validation

kdeininger:
An optional configuration parameter `validation` is added to the `OEREBlexSource`. The value is passed to the `geolink_formatter` and turns the XML validation on/off.

To ensure backwards compatibility, the default value is `true` which means the validation to be enabled.

Related to https://github.com/openoereb/geolink_formatter/issues/85


jwkaltz:
@kdeininger do you want to merge this? 
(Together with the shapely revert, I would do a maintenance release afterwards)

PR #1061, opened 2020-07-09
Drop python2 support

### LINKED ISSUE #889: Drop python 2.7 support
Python 2.7 [stops receiving bug fixes after January 1, 2020](https://www.python.org/dev/peps/pep-0373/). By then we should drop python 2.7 support and all users should have migrated to python3 by then.

jwkaltz:
Fix #889

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ language: python
 
 matrix:
   include:
-  - python: "2.7"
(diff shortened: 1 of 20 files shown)


jwkaltz:
Already reviewed PR in earlier version; now re-applying after publication of release v1.7.6.

PR #1085, opened 2020-10-19
Oereblex: add parameter for oereblex call

### LINKED ISSUE #1065: OEREBlex: usage of oereb_id
In the OEREBlex API, there is a parameter oereb_id which is not yet used by pyramid_oereb. 

Liechtenstein would like to use this parameter to be able to have different contexts for the same geolink, for example:
- https://oereblex.llv.li/api/geolinks/5.html?oereb_id=8
- https://oereblex.llv.li/api/geolinks/5.html?oereb_id=16

The document used for LandUsePlans is also relevant for the NoiseSensitivityLevels. To avoid entering it twice in Oereblex, the mechanism with the oereb_id can be used in Oereblex (but as this parameter is not implemented yet in pyramid_oereb, it can not be used further).

Is this a need for other OEREBlex users as well?

jwkaltz:
Fix #1065

Proposal to add an (optional) parameter string for oereblex calls, to be able to use "oereb_id" parameter, or other parameters in the future in a generic way. 
In the proposal of this PR, the parameter string would be stored in the model, in addition (separately) to the "geolink". 

Status: DRAFT, for now, looking for feedback whether this approach makes sense. To be tested yet on a real instance.

C2C internal reference: GEO-3711

### DIFF
--- a/pyramid_oereb/contrib/sources/document.py
+++ b/pyramid_oereb/contrib/sources/document.py
@@ -76,21 +76,23 @@ class OEREBlexSource(Base):
         if self._parser.host_url is None:
             raise AssertionError('host_url has to be defined')
 
-    def read(self, params, geolink_id):
+    def read(self, params, geolink_id, oereblex_params):
         """
         Requests the geoLink for the specified ID and returns records for the received documents.
 
         Args:
             params (pyramid_oereb.views.webservice.Parameter): The parameters of the extract request.
             geolink_id (int): The geoLink ID.
+            oereblex_params (string): Any additional parameters to pass to Oereblex
         """
         log.debug("read() start")
 
         # Request documents
-        url = '{host}/api/{version}geolinks/{id}.xml'.format(
+        url = '{host}/api/{version}geolinks/{id}.xml?{url_params}'.format(
             host=self._parser.host_url,
             version=self._version + '/' if self._pass_version else '',
-            id=geolink_id
+            id=geolink_id,
+            url_params=oereblex_params
         )
         language = params.language or self._language
         request_params = {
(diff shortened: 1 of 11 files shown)


pyramid_oereb/contrib/sources/document.py
```diff
@@ -76,21 +76,23 @@ def __init__(self, **kwargs):
         if self._parser.host_url is None:
             raise AssertionError('host_url has to be defined')
 
-    def read(self, params, geolink_id):
+    def read(self, params, geolink_id, oereblex_params):
```
ger-benjamin, line 79:
```suggestion
    def read(self, params, geolink_id, oereblex_params=None):
```
To avoid passing `None` explicitely in tests for instance

ger-benjamin (approved):
Not tested but looks good.

kdeininger (commented):
In my opinion, this is more a "per topic configuration". So my preferred solution would be, to put the OEREBlex parameters into the `pyramid_oereb.yml` instead of every data record.

jwkaltz:
> In my opinion, this is more a "per topic configuration". So my preferred solution would be, to put the OEREBlex parameters into the `pyramid_oereb.yml` instead of every data record.

Thanks for your input @kdeininger - but I think we can not do this cleanly in the theme ("plr") configuration section, because of the shared data types between non-oereblex world and oereblex world.

Or do you mean within the oereblex: configuration section, add additional theme configurations? This would be easy, but could be confusing as the themes would then be configured in two different sections?

jwkaltz:
> Or do you mean within the oereblex: configuration section, add additional theme configurations? This would be easy, but could be confusing as the themes would then be configured in two different sections?

@kdeininger what do you think about this?

kdeininger:
I don't think this would be confusing. It's just a mapping of theme codes with IDs from OEREBlex. There are already other mappings, that can be configured within the `oereblex` section. This way, users without OEREBlex won't be affected in the topic configuration.

jwkaltz:
Superseded by #1117

PR #1089, opened 2020-10-21
For full extracts, add configuration parameter to make additional sld usage optional

### LINKED ISSUE #1077: "full" PDF extract uses WMS SLD
Hi everyone

we want to integrate the full PDF extract for the Canton AR.
While integration, we have found out, that the full PDF extract requires an SLD from the WMS used. 
The SLD is not required when only a reduced extract is requested. 

We have an issue with the SLD, which we can not fix easily.

"LookupError: The image could not be downloaded. URL was: https://www.geoportal.ch/services/wms/ktai?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=ch.geoportal.basiskarten_landschaftsmodelle.1492.0.hk_oereb&mapId=1492&CRS=EPSG%3A2056&WIDTH=2055&HEIGHT=1169&BBOX=2748662.0319265304%2C1243597.057142857%2C2748714.3140734695%2C1243626.750857143&SRS=EPSG%3A2056&sld=https%3A%2F%2Foereb.ai.ch%2Fktai%2Fwsgi%2Foereb%2Fsld%3Fegrid%3DCH214141597761, Response was <?xml version="1.0" encoding="UTF-8"?><ServiceExceptionReport version="1.3.0" xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ogc http://wms.live.geoportal.ch:80/schemas/wms/1.3.0/exceptions_1_3_0.xsd">"

Is the request of the SLD for the full extract required? Or would it be possible that the full extract is implemented in the same way as the reduced? 
If yes, then I will work with camp2camp and let them make a configuration setting in pyramid_oereb which enables to not use the SLD. 

Thanks & Cheers
Max

jwkaltz:
Fix #1077

Tested on a cantonal infrastructure, as full extract can not be tested in local docker test environment.

C2C internal reference: GEO-3534

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,11 @@
 Changelog
 ---------
 
+1.8.0 (DRAFT)
+*************
+- Enhance federal data import script to make it more usable with Docker (#1078)
+- For full extracts, add configuration parameter to make additional sld usage optional (#1077)
+
 1.7.6
 *****
 - Improve federal data import script (#1057)
--- a/docker/config.yaml.mako
+++ b/docker/config.yaml.mako
@@ -172,6 +172,9 @@ vars:
           types:
           - GeometryCollection
 
+    # Configuration options for full extract
+    full_extract_use_sld: true
+
     # Configuration for OEREBlex
     oereblex:
       # OEREBlex host
--- a/docker/config.yaml.mako
+++ b/docker/config.yaml.mako
@@ -172,7 +172,7 @@ vars:
           types:
           - GeometryCollection
 
-    # Configuration options for full extract
+    # Configuration option for full extract: apply SLD on land register WMS (defaults to true)
     full_extract_use_sld: true
 
     # Configuration for OEREBlex
--- a/pyramid_oereb/lib/processor.py
+++ b/pyramid_oereb/lib/processor.py
@@ -305,6 +305,7 @@ class Processor(object):
         extract.glossaries = glossaries
         # obtain the highlight wms url and its content only if the parameter full was requested (PDF)
         if params.flavour == 'full':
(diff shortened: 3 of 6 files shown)


ger-benjamin (commented):
Can you please improve the description of what this option does (in config files) ? Also says that defaults is `True` ?

With this change, we don't have anymore a highlight ?
I'm not sure of what this method does: https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/records/real_estate.py#L92 (a description is also welcome) but perhaps it's better to remove the update_sld_url param if your new config is set to False ?

jwkaltz:
> Can you please improve the description of what this option does (in config files) ? Also says that defaults is `True` ?

OK, done.

> 
> With this change, we don't have anymore a highlight ?

It depends on the WMS use, but the goal is to be able to not retrieve an additional SLD when generating a full extract.

> I'm not sure of what this method does: https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/records/real_estate.py#L92 (a description is also welcome) but perhaps it's better to remove the update_sld_url param if your new config is set to False ?

In fact, due to the change in lib/processor.py, the method set_highlight_url will not be called at all if in the config the parameter is set to false.

ger-benjamin:
I've read again the code and I understand now, you code is correct, we want to avoid using the whole set_highlight_url method.

I've added some comments on this function. It's a little bit easier to understand now.

ger-benjamin (approved):

PR #1093, opened 2020-11-02
oereblex enactment_date: log warning instead of making extract fail

jwkaltz:
oereblex enactment_date: log warning instead of making extract fail when enactment_date is empty

C2C reference: GEO-3710


kdeininger:
I think, `enactment_date` stands for `publiziertAb`, which is a mandatory attribute in the transfer structure according to the federal specification. However, this attribute does not occur in the extract model. So, for me it would be okay to log a warning if it is missing.

kdeininger (approved):

PR #1107, opened 2021-01-07
Add test for empty geometry of intersection results (Python 3.7 support)

vuilleumierc:
With Python 3.7 there is currently a failed test, test_geometry for geometry9: the `Point(2, 2)` is evaluated as in the polygon `MultiPolygon([Polygon(((0, 0), (0, 1), (1, 1), (1, 0)))])`). After trying with my local installations of Python, I saw that the type of the return values for shapely intersection in case of empty intersection appears to be different (in both cases version 1.6.4.post1). In Python 3.6 the return value of the shapely function was an empty GeometryCollection and in 3.7 it is an empty Point. I added a test for empty geometry similarly to fc260fde8dc559390edc4ca5f0fb61cb4f0776bb which solved this issue, but I think further testing of this issue on other installations would be a nice

### DIFF
--- a/pyramid_oereb/lib/records/geometry.py
+++ b/pyramid_oereb/lib/records/geometry.py
@@ -131,49 +131,51 @@ class GeometryRecord(object):
         polygon_types = geometry_types.get('polygon').get('types')
         point_types = geometry_types.get('point').get('types')
         if self.published:
-            result = self._extract_collection(self.geom.intersection(real_estate.limit))
+            intersection = self.geom.intersection(real_estate.limit)
             # TODO upon update to Shapely 1.7, a check for result.is_emtpy will be needed (see PR#1037)
             # differentiate between Points and MultiPoint
-            if self.geom.type not in point_types + line_types + polygon_types:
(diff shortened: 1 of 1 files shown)


vuilleumierc:
Here is what I tested:
Python 3.6
image
Python 3.7
image

The results are then handled differently in pyramid_oereb, see https://github.com/openoereb/pyramid_oereb/blob/353e99837b4f2dd34850e30203639d01f41bdb59/pyramid_oereb/lib/records/geometry.py#L94
and in 3.7 the point is assumed to be in the polygon https://github.com/openoereb/pyramid_oereb/blob/353e99837b4f2dd34850e30203639d01f41bdb59/pyramid_oereb/lib/records/geometry.py#L149-L152

jwkaltz (approved):
The change is correct.
But I wonder if it will impact how pyramid_oereb reacts to invalid geometries (the Ring self-intersections, which are allowed in Interlis but not by OGC).
We should add a test case for that -> I will create a separate issue, this PR can be merged.

PR #1109, opened 2021-01-08
Add tests for invalid geometries

### LINKED ISSUE #1108: Add test case for invalid geometries
pyramid_oereb should have a test case for invalid geometries, to clearly define what the expected behavior is in case of invalid geometries. Furthermore, this will help to notice the impact of library changes on the handling of invalid geometries - notably shapely, see #1013.

vuilleumierc:
Fixes #1108
Add two tests with invalid geometries


jwkaltz (commented):
we should try to be more explicit with the purpose of the file geometry1.txt -
maybe call it "interlis-valid-ogc-invalid-geometry.txt" or something like that.
Plus, maybe we should add an additional unit test somewhere to check that this is really an invalid geometry (according to OGC)?

tests/records/test_geometry.py
```diff
@@ -221,11 +247,33 @@ def test_calculate(geometry, real_estate_geometry, length_limit, area_limit, len
...
         2761,
-        1,
+        real_estate_geometry.area,
         real_estate_geometry
     )
     geometry_record.calculate(real_estate, length_limit, area_limit, 'm', 'm2')
     assert geometry_record._test_passed == test
     assert geometry_record._length_share == length_share
-    assert geometry_record._area_share == area_share
+    if geometry_record._area_share is None:
```
jwkaltz, line 256: I don't understand this change

tests/records/test_geometry.py
```diff
@@ -221,11 +247,33 @@ def test_calculate(geometry, real_estate_geometry, length_limit, area_limit, len
...
+        assert geometry_record._area_share == area_share
+    else:
+        assert round(geometry_record._area_share) == area_share
```
jwkaltz, line 259: Why round? And how is this related to the test of invalid geometries?

jwkaltz (approved):

Issue #1100, opened 2020-12-16
Add python 3.7 and 3.8 support


jwkaltz:
Current pyramid_oereb officially supports Python 3.6. This is the version that is currently tested via CI (continuous integration).

The project should verify that it is also functional with Python 3.7 and 3.8.
* manual tests with dummy data,
* automated tests,
* additional CI runs (for the newer versions of Python) -> this should be done simultaneously to #1103 

Note: Python 3.9 is currently not being addressed, because there are some known issues with pyproj with that version of Python, so we will wait for new releases of the used libraries.

vuilleumierc:
There is currently still a problem regarding intersection operation with OGC invalid (but Interlis valid) geometries in Python 3.7, even when using shapely 1.6.4. This operation is not possible anymore and raises an error. The version of the underlying GEOS library is different when installing shapely using pip in Python 3.6 (GEOS 3.6.2) and in Python 3.7 (GEOS 3.8.0). This change of behavior of GEOS is discussed here: https://github.com/r-spatial/sf/issues/1121

ger-benjamin:
We have versions incoherence between tests and dev:
To have same version in tests and in dev we should have the tests running in a container instead of a local virtual env.
I see three scenario:
 - We move all tests in the main (pyramid_oereb _server) container. But the image will be heavier (files and all dev-requirements).
 - We split tests to run only unit test (that need to access the code itself) in the main container, and have a dedicated container for acceptance test. But it can be long to sort our test are they are all mixed together. 
 - We have a container dedicated for the test (with tests and code). Perhaps more work but probably more clean.

In all case we have to clean a little bit the Makefile (adapt the targets and know why we use two ways to have the same DB ?)
Perhaps we have to clean a little bit the dev-requierements.

With tests and dev in a container, we have the risque to loose some check on the non-docker version.

@kdeininger do you have an opinion on that ?

kdeininger:
I would prefer the third option. We could use the application image as base for the test image and just add the tests and their dependencies. This way, the application image should stay small and clean.

> With tests and dev in a container, we have the risque to loose some check on the non-docker version.

That is right. We would have to build the application and test images with each supported Python version.

The two databases have different purposes. The first one contains sample data that can be used for development or a local demo. The second database is used for testing and contains special test cases. It gets filled and cleaned on every test run. We could combine both databases, but then we would have to adapt the tests to work with the sample data and ensure the database to be restored after each test to be available for development.

[...]

PR #1115, opened 2021-01-29
Tests multi python

ger-benjamin:
For #1100 ( GSPOE-40) 

First, I had trouble to run tests with multiple version of python3 installed (I always have python3.7 installed, even if I only want python 3.6...) Then I've used [tox](https://tox.readthedocs.io/en/latest/index.html) to test multiple version of python (3.6, 3.7 and 3.8).
I've changed that in my last commit. Now the CI run a specific version of python 3 multiple times, And I've removed `tox`. I would like to keep the commit with tox as it's not a bad way to solve this problem.

The tests container part took me some times. I use a specific `tests.Dockerfile` to build env for tests. It doesn't inherit (start from) from the Dockerfile as it's finally useless. The main problem is that I can't avoid to ignore tests-related file in the `.dockerignore`. That means we have some new files (including the tests) in the pyramid_oereb image.
It looks possible to have one specific `.dockerignore` for a specific Dockerfile but I cannot make that works... (ref: https://github.com/moby/moby/issues/12886#issuecomment-480575928)


vuilleumierc (approved):
Looks good to me :)

PR #1121, opened 2021-02-03
Restrict version of numpy to maintain Python 3.6 compatibility

### LINKED ISSUE #1120: Numpy release (now: 1.20.0) no longer supports Python 3.6
The newest release of Numpy (1.20.0) no longer supports Python 3.6. Python 3.7 or higher is required. Attempts to install numpy 1.20.0 in Python 3.6 fail.
See also https://numpy.org/neps/nep-0029-deprecation_policy.html

jwkaltz:
Fix #1120 
(workaround as long as pyramid_oereb officially supports Python 3.6)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,5 +22,5 @@ requests==2.25.1
 geolink-formatter==1.5.3
 pyconizer==0.1.4
 tqdm==4.56.0
-numpy==1.19.5
+numpy==1.19.5 # rq.filter: <1.20.0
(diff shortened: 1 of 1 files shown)


ger-benjamin (approved):
Tests fails with an higher version of numpy ?

jwkaltz:
> Tests fails with an higher version of numpy ?

There is no package of numpy 1.20.0 for Python 3.6.

PR #1126, opened 2021-02-15
Use localtime

jwkaltz:


marionb (approved):
:+1: I like this solution it solves all the issues together.
Tested locally and is working well.

Issue #1103, opened 2021-01-04
Continuous Integration: move away from travis-ci.org


jwkaltz:
Currently, the continuous integration (notably, automated tests) are handled in travis-ci.org

However, in travis-ci.org, the following appears: "Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information."

We need to evaluate whether travis-ci.com should be used instead, and if so, with which account, or if continuous integration should be migrated somewhere else.

jwkaltz:
@kdeininger since oereb_client now has the CI on GitHub actions, I suppose it would be best to use this also for pyramid_oereb, what do you think?

kdeininger:
I think GitHub Actions will fit our needs and it seems to be much faster. Migrating to GitHub Actions should not be very complicated.

vvmruder:
Moving to github actions 👍

How do we plan the step.

jwkaltz:
We plan to do this via maintenance budget. There should not be any side effects.

PR #1134, opened 2021-03-15

jwkaltz:


jwkaltz:
Local checks are successful now, CI fails because of docker pull rate exceeded (see #1103)
I will merge anyway to have the fix for SQLAlchemy now, and monitor next daily builds.

PR #1139, opened 2021-03-19
Add configuration option for print header

vuilleumierc:
Add new flag in printable extract to be able to hide canton logo from PDF header (GEO-4197)

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -544,6 +544,8 @@ class Renderer(JsonRenderer):
                 if 'NrOfPoints' in legend:
                     legend['NrOfPoints'] = '{0}'.format(legend['NrOfPoints'])
 
+        extract_dict['PrintCantonLogo'] = Config.get('print', {}).get('print_canton_logo', True)
+
         log.debug("After transformation, extract_dict is {}".format(extract_dict))
         return extract_dict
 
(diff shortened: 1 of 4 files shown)


pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -544,6 +544,8 @@ def convert_to_printable_extract(self, extract_dict, feature_geometry, pdf_to_jo
                 if 'NrOfPoints' in legend:
                     legend['NrOfPoints'] = '{0}'.format(legend['NrOfPoints'])
 
+        extract_dict['PrintCantonLogo'] = Config.get('print_canton_logo')
```
jwkaltz, line 547: Default must be true (if the Config item is missing)

pyramid_oereb/standard/pyramid_oereb.yml.mako
```diff
@@ -248,6 +248,9 @@ pyramid_oereb:
   # The method used to return the municipality logos.
   get_municipality_method: pyramid_oereb.standard.hook_methods.get_municipality
 
+  # Flag to print or not the canton logo
```
jwkaltz, line 251: This must be in the print configuration section (only for mapfish print configuration options)

voisardf:
sorry to intervene. Who asked for this configuration, what is exactely the reason behind it? As I believe to recall, the logos are mandatory in the specification.

jwkaltz:
> sorry to intervene. Who asked for this configuration, what is exactely the reason behind it? As I believe to recall, the logos are mandatory in the specification. - or is it just dependent on the print method used?

Hi @voisardf , indeed the cantonal logs will still be active by default. We want to add a configuration option for Liechtenstein, who wants to be able to not use a cantonal logo, but still stick to the standard code and print templates.

voisardf:
which makes perfect sense! My apologies to Lichtenstein 😃 👍

jwkaltz (approved):

Issue #1160, opened 2021-04-29
Statistics: DB password hardwired in production.ini


lopo977:
Password or credentials are normally passed via environment variables. (i.e. SQL_ALCHEMY_URL / POSTGRES_SERVICE_PASS in config.yaml.mako / config.yaml)
For the 'statistics' functionality, the database instead password is hardwired in the production.ini file (see [handler_sqlalchemylogger])

Is it possible to use the environment variable also for the connection to the database for statistics? 
Thank you. 
Regards,
Luca

jwkaltz:
@lopo977 Hi Luca, we did the implementation in #1167.
For better consistency, we plan to do similar adaptations for the other variables, so that they are also real environment variables, in a subsequent pull request.

lopo977:
@jwkaltz Hi Wolfgang. 
Thank you for the adaptation. 
I've just seen the implementation in #1167

PR #1167, opened 2021-05-04
Use environment variables for statistics DB credentials

vuilleumierc:
Issue #1160
To try it out:
`cp sample.env .env`
-> insert credentials in .env (see [here](https://github.com/openoereb/pyramid_oereb/blob/2a1d3e2c26d5e396cc5814f3908f01b9d6d9c4f0/docker/production.ini#L78))
`make serve`


sample.env
```diff
@@ -0,0 +1,5 @@
+STATISTICS_DB_USER=<db-user>
+STATISTICS_DB_PASS=<db-password>
+STATISTICS_DB_HOST=<db-host>
+STATISTICS_DB_PORT=<db-post>
```
jwkaltz, line 4: db-port

jwkaltz (changes requested):
Implementation looks good.
Can you please adapt the documentation in doc/source/contrib/stats.rst.mako
And I think we should change the README.rst as well

jwkaltz (approved):

PR #1173, opened 2021-05-05
Database connection as environment variables

vuilleumierc:
For the Docker version: this is to avoid having the database credentials in the docker-compose file. They have to be provided in a .env file at the root of the project (this file is not committed, see sample.env for an example). To do: update doc


jwkaltz (changes requested):
In bin/eval-templates we should avoid listing individual files. Maybe we simply get rid of the /etc/pyramid_oereb path in the container and have the config.yaml in the same directory as the production.ini? The only place the path /etc/pyramid_oereb is used is in the variable pyramid_oereb.cfg.c2ctemplate.file

PR #1233, opened 2021-06-21
Update dependency of psycopg, avoid binary package

jwkaltz:
Update dependency of psycopg, avoid binary package as it is no longer working and it is anyway not recommended to use in production.

Fixes broken update of #1232

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,8 @@ FROM camptocamp/c2cwsgiutils:3
 LABEL maintainer "info@camptocamp.org"
 
 WORKDIR /app
-ARG DEV_PACKAGES="python3.7-dev build-essential libgeos-dev"
+ARG DEV_PACKAGES="libgeos-c1v5 libpq-dev"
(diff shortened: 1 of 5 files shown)


jwkaltz:
@ger-benjamin @danduk82 I wanted to pass the new "DEV_PACKAGES" in the "serve" target, to avoid having the libraries listed in the Dockerfile, but I could not find a way.
Do you have an idea?
(Of course we can always do that simplification in a later PR)

ger-benjamin:
As you use `make serve` to make the `docker-compose up`, you can add this line
DEV_PACKAGES=${DEV_PACKAGES} here https://github.com/openoereb/pyramid_oereb/blob/master/docker-compose.yml#L23 and that should work... maybe.

But that's strange to use environment variable for dockerfile packages to install... I wonder if it's not better to revert this change (sorry) and simply add a commentary in both Dockerfile and test.Dockerfile files

jwkaltz:
@ger-benjamin I updated the PR as discussed

ger-benjamin (commented):
:ok_hand:

Issue #1172, opened 2021-05-05
Changes for class office


voisardf:
- [x] office_at_web is now multilingual URI

### Standard model definitions for db (standard models are created through templates!):
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/templates/plr_integer_primary_keys.py.mako#L65
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/templates/plr_string_primary_keys.py.mako#L65

### PLR sources definition:
https://github.com/openoereb/pyramid_oereb/blob/c5c6cb571027b41bf158f3118ae3bfff4a60be6d/pyramid_oereb/standard/sources/plr.py#L219-L231

### Record definition (new type will be dict):
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/records/office.py#L27

### Import Script:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/xtf_import/document.py

### Renderer:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/renderer/extract/templates/xml/office.xml#L5-L7

### Tests:
https://github.com/openoereb/pyramid_oereb/blob/master/tests/records/test_office.py

PR #1242, opened 2021-06-29
Attribute office_at_web becomes a multilingual dictionary

vuilleumierc:
Issue #1172


vvmruder:
Bildschirmfoto 2021-06-29 um 15 41 02

Proof of working code for a XML extract (JSON is missing for now). I used the following link:
http://localhost:6543/oereb/extract/reduced/xml/CH113928077734

Maybe you need to start with `make clean serve-dev` to newly import updated test dev data on server startup?

vuilleumierc:
Thanks for checking. The XML works for me too, but the structure does not seem to be correct in the JSON extract
image

vvmruder:
Ok. Will do further checks

voisardf:
I'm not sure, but isn't the flattening for the JSON done in the contrib print_proxy? If so, there might be an adaption missing for this attribute?

vuilleumierc:
I found the problem (see commit 476ace9) but I will check also with mapfish_print

vvmruder:
Problem is probably somewhere here: https://github.com/openoereb/pyramid_oereb/blob/b20ebb5ab36b3aa888fbad083e6bc18fb1d04c36/pyramid_oereb/lib/renderer/extract/json_.py#L454-L455

vvmruder:
> I found the problem (see commit [476ace9](https://github.com/openoereb/pyramid_oereb/commit/476ace9a2e96745845de0b53e0fe9341c6acb54f)) but I will check also with mapfish_print

Exactly :D

Good catch

voisardf:
👍

voisardf (approved):
Looks good to me, thanks

vvmruder:
Did you tried `make test`?

vuilleumierc:
Good point, I hadn't done it since my last commits :) it is fixed

vvmruder:
@vuilleumierc I added the multilingual attribute also in some places for the test data. It came to my curiosity why checks are still ok. Especially the checks against the xml schema. After changing to multilingual they should fail. So I updated the used test data as well to cover also this case now.

vvmruder (approved):
I think now its pretty complete. Congrats for this fast and nice work.

PR #1246, opened 2021-07-02
1239-integrate-new-schema

michmuel:
Also here the failed tests:
FAILED tests/renderer/test_xml.py::test_extract_against_schema[parameter0-test_extract0]
FAILED tests/renderer/test_xml.py::test_extract_against_schema[parameter1-test_extract1]
FAILED tests/renderer/test_xml.py::test_extract_against_schema[parameter2-test_extract2]

Are other resources required?


vvmruder (approved):
This looks pretty complete. I think we can remove all the old schemas since we only want to satisfy the new spec.

PR #1248, opened 2021-07-05
Implement model factories for standard

vvmruder:
For a more convenient way to use pyramid_oereb in terms of integration and development factory functions are introduced to use the standard models. In this way redundant code is omitted and it is not necessary to "produce" the models physically before they can be used.


kdeininger (approved):

vvmruder:
Bildschirmfoto 2021-08-02 um 12 38 07

I found this error on running the server.

tests/init_db.py
```diff
@@ -1,18 +1,33 @@
...
 
 file_adapter = FileAdapter()
 
+Config._config = None
+pyramid_oereb_test_yml = 'pyramid_oereb/standard/pyramid_oereb.yml'
+Config.init(pyramid_oereb_test_yml, 'pyramid_oereb')
+
+def create_models():
+    themes = parse_multiple_standard_themes(Config)
+    return themes
```
marionb, line 20: the function is `create_models` but the variable returned is `themes`. This is a bit confusing to read.

tests/init_db.py
```diff
@@ -23,8 +38,11 @@ def init(self):
         self._import_forest_perimeters()
 
     def _truncate(self):
-        from pyramid_oereb.standard.models import main, contaminated_sites, \
-             land_use_plans, motorways_building_lines, forest_perimeters
+        from pyramid_oereb.standard.models import main
```
marionb, line 41: This function looks somehow hard to maintain to me is it not possible to do something similar to what you did in `create_tables.py` but instead of drop truncate the tables?

marionb (changes requested):
Here some feedback. The code in general looks good.
Perhaps you can re-base to the current master so it is easier to test also with the new make targets.

tests/init_db.py
vvmruder: True, I changed it

tests/init_db.py
vvmruder: I understand what you mean-But I think this will go away asap when the testing and sample data is joined. I would like to keep this out of scope for this PR-do you agree?

vvmruder:
@marionb I merged the master into this branch.

voisardf:
I'm ok with merging this PR and addressing the error existing when loading test data by issue https://github.com/openoereb/pyramid_oereb/issues/1308 as it's something touching the current work of @marionb
It makes no sense to me to clean it up here when it risks generating new data and code conflicts with the ongoing work in PR https://github.com/openoereb/pyramid_oereb/pull/1306

so if it's ok for you @jwkaltz @marionb @vvmruder I'd suggest to merge this PR and have a short discussion between Marion and Clemens to clarify how the missing step can be directly integrated in the ongoing work  in https://github.com/openoereb/pyramid_oereb/pull/1306

pyramid_oereb/standard/models/theme.py
```diff
@@ -0,0 +1,366 @@
...
+            published_until (datetime.date): The date from when the geometry should not be available
+                anymore for publishing on extracts. This directly affects the behaviour of extract
+                generation.
+            geo_metadata (str): A link to the metadata which this geometry is based on which delivers
+                machine  readable response format (XML).
+            public_law_restriction_id (str): The foreign key to the public law restriction this geometry
+                is  related to.
+            public_law_restriction (PublicLawRestriction): The dedicated relation to the public law
+                restriction instance from database.
+            office_id (str): The foreign key to the office which is responsible to this public law
```
michmuel, line 285: Is this made intentionally or is it took over from a former model. I think there is no such relation in the "Rahmenmodell".

tests/init_db.py
marionb: Yes that is fine for me.

marionb (approved):
Thanks for the changes @vvmruder
Testing this locally works.
We can take a look at what needs to change for the DB part and if I can add this in my PR.

pyramid_oereb/standard/models/theme.py
vvmruder: You are right. But for the sake of blocking I filed a new issue and we will solve this in a new PR https://github.com/openoereb/pyramid_oereb/issues/1310

PR #1250, opened 2021-07-06
Remove isReduced from extract only - no api

### LINKED ISSUE #1177: Changes for class Extract
Only 'small' changes for this class
- [ ] isReduced - attribute abandonned
- [x] two attributes renamed in german (ThemeWithoutData => **ThemaOhneDaten**, UpdateDateOS => **StandAV**) 
       NOTE: These two attributes are only used in the extract where they are use in English. It is therefore not translated there.
- [x] multilingual logos and other multilingual adaptions already in pyramid_oereb

### Renderer:
https://github.com/openoereb/pyramid_oereb/blob/c5c6cb571027b41bf158f3118ae3bfff4a60be6d/pyramid_oereb/lib/renderer/extract/templates/xml/extract.xml#L72
https://github.com/openoereb/pyramid_oereb/blob/c5c6cb571027b41bf158f3118ae3bfff4a60be6d/pyramid_oereb/lib/renderer/extract/json_.py#L92

### Tests:
https://github.com/openoereb/pyramid_oereb/blob/787ce93db13fa9af2b775b9a80a7b4078fee56a3/tests/resources/schema/20160830/extractdata.json#L313
https://github.com/openoereb/pyramid_oereb/blob/787ce93db13fa9af2b775b9a80a7b4078fee56a3/tests/resources/schema/20170825/extractdata.json#L655
https://github.com/openoereb/pyramid_oereb/blob/787ce93db13fa9af2b775b9a80a7b4078fee56a3/tests/renderer/test_json.py#L134
https://github.com/openoereb/pyramid_oereb/blob/787ce93db13fa9af2b775b9a80a7b4078fee56a3/tests/contrib/print_proxy/resources/expected_getspec_extract.json#L694
https://github.com/openoereb/pyramid_oereb/blob/787ce93db13fa9af2b775b9a80a7b4078fee56a3/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L144

marionb:
fixes #1177 
@vvmruder will this part still be used in the end?
https://github.com/openoereb/pyramid_oereb/blob/787ce93db13fa9af2b775b9a80a7b4078fee56a3/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L144
~~It is only if the extract is reduced.~~ As I understood the aim is to have only the reduced extract so we should be able to remove everything that is in that if. Would you agree?
If not how should it be handled if there is no `isReduced` parameter any more?


vvmruder:
AS I understand it, this check should go away. Only status code should be checked from now

marionb:
So unless the status code is not OK it will always enter this `IF` condition (this will be the standard case).
Until now this was not the case. Only if something other than `reduced` or `embeddable` (`isReduced = False`) was called in the request, this part of the code was called.

vvmruder:
Yes correct. But as far as I see we only have one kind of extract in the future. At least in the document https://github.com/openoereb/pyramid_oereb/files/6442566/OEREB-XML-Aufruf_de.pdf chapter 3.2 there is no URL parameter anymore to request a special flavour from the getExtractById Service.

vuilleumierc:
Maybe you could still adapt these two files, which relate to the extract:
https://github.com/openoereb/pyramid_oereb/blob/1ca7f6bf8803184e09f4dac03e13779ae47ce6fa/pyramid_oereb/lib/renderer/extract/json_.py#L260

https://github.com/openoereb/pyramid_oereb/blob/1ca7f6bf8803184e09f4dac03e13779ae47ce6fa/pyramid_oereb/lib/renderer/extract/templates/xml/public_law_restriction.xml#L6

otherwise I tested and everything looks fine :+1:

marionb:
OK it is removed, I was not sure if that is not more part of the api. But if it is gone we can not forget about it :-)

vvmruder (changes requested):
I did a local `make test` and it threw another failing test:
```
FAILED tests/renderer/test_json.py::test_format_plr[parameter2] - Failed: DID NOT RAISE <class 'ValueError'>
```

Can you confirm this? If yes, I think there must be something to fix in the tests.

vuilleumierc (approved):

marionb:
@vvmruder I just abandoned testing the full flavour - they were failing since I also removed that code part in the rendered. I hope that is OK?

vvmruder:
> @vvmruder I just abandoned testing the full flavour - they were failing since I also removed that code part in the rendered. I hope that is OK?

Yes. That was, what I thought about.

Did you cross checked the whole code to remove the "flavour-behavior"? Like remove all occurencies of Full, Reduced, Embeddable and the logical switches for this? Maybe the Issue was not that clear about it. We only have one extract in the future. No need for this thing for example:
https://github.com/openoereb/pyramid_oereb/blob/abandon_isReduced_%231177/pyramid_oereb/lib/renderer/extract/templates/xml/extract.xml#L35-L45

marionb:
> Did you cross checked the whole code to remove the "flavour-behavior"? L ...

@vvmruder no because it was not clear to me to also do that. I thought I should not change the API yet and it was not so easy to distinguish where the api stars and what. I'll take another look

vvmruder:
You are right. this was not described very well. Since there are some overlapping tasks I would be happy if we can already get rid of the code parts around the flavours. What do you think? Or better doing this in later project.

marionb:
@vvmruder I think it makes more sense to remove that now. I can do that.

vvmruder:
Thank you. I hope I was not impolite :)

marionb:
@vvmruder I think now I removed everything I could find that is not relevant to the API.
there is however this: https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/records/embeddable.py
Can this be removed also what do you think? If I remove it it will have a big effect.

vvmruder:
> @vvmruder I think now I removed everything I could find that is not relevant to the API.
> there is however this: https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/records/embeddable.py
> Can this be removed also what do you think? If I remove it it will have a big effect.

I think we leave it. For my understanding it holds the information about the cadastral data. I will have a more detailed look and open another issue if necessary.

vvmruder (approved):

vvmruder:
Yes. Merge it please.

PR #1251, opened 2021-07-06
Changes for class ViewService

### LINKED ISSUE #1179: Changes for class ViewService
Some attribute changes already existing and some new ones:
- map - multilingual blob => Download WMS image with correct multilingual url 
- reference_wms - multilingual URI
- legend_at_web - attribute abandonned (was complet legend element in PDF extract)
- min, max - renamed attributes NS95_min, NS95_max
- [x] layer_index - new, but already implemented in pyramid_oereb
- [x] layer_opacity - new, but already implemented in pyramid_oereb

### What needs to be done:
- [ ] make sure the map object supports multilingual blob
- [x] transform reference_wms into multilingual URI
- [x] remove occurences of legend_at_web everywhere
- [ ] get rid of min_NS03, max_NS95 elements, maybe rename min_NS95, max_NS95 attributes troughout the code https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/records/view_service.py#L90-L93

### Standard model definitions for db (standard models are created through templates!):
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/templates/plr_integer_primary_keys.py.mako#L244-L259
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/templates/plr_string_primary_keys.py.mako#L244-L259

### PLR sources definition:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/sources/plr.py#L108-L117

### Record definition:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/records/view_service.py#L61-L311

### Import Script:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/xtf_import/view_service.py

### Renderer:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/renderer/extract/templates/xml/view_service.xml
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/renderer/extract/json_.py#L489-L530

vuilleumierc:
Closes #1179
edit: ~~I saw in the Rahmenmodell that the map image needs to be a MultilingualBlob, so this still needs to be done. But if you already have some input feel free to comment~~
To get the extract with multilingual images: 
http://localhost:6543/oereb/extract/reduced/json/CH113928077734?WITHIMAGES=true
To change the language: 
http://localhost:6543/oereb/extract/reduced/json/CH113928077734?WITHIMAGES=true&LANG=fr


pyramid_oereb/lib/readers/real_estate.py
```diff
@@ -57,14 +57,14 @@ def read(self, params, nb_ident=None, number=None, egrid=None, geometry=None):
         plan_for_land_register_config = Config.get_plan_for_land_register_config()
 
         real_estate_view_service = ViewServiceRecord(
-            plan_for_land_register_config.get('reference_wms'),
+            plan_for_land_register_config.get('reference_wms')[Config.get('default_language')],
```
marionb, line 60: perhaps you can get the language from the `params` and then use that language or if it does not exist use the default. Something like what was done for the logo here https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/lib/readers/extract.py#L150.
Like that you are sure to use the correct language and not only the default language. But first check if this is required.

marionb (changes requested):
most looks good.
Just one point for the language.

pyramid_oereb/lib/readers/real_estate.py
vuilleumierc: thanks, I will check that :)

vvmruder: Yes, I would like some solution suggested by @marionb

pyramid_oereb/lib/records/view_service.py
```diff
@@ -216,80 +209,89 @@ def get_bbox(geometry):
...
-        self.reference_wms = add_url_params(self.reference_wms, {
-            "BBOX": ",".join([str(e) for e in bbox]),
-            "SRS": 'EPSG:{0}'.format(Config.get('srid')),
-            "WIDTH": int(map_size[0]),
-            "HEIGHT": int(map_size[1])
-        })
-        self.calculate_ns()
+
+        if language not in self.reference_wms:
+            msg = f"No WMS reference found for the requested language ({language}), using default language"
```
voisardf, line 233: I didn't know about this formatting possibility. Nice, I learned something today 👍

pyramid_oereb/lib/renderer/extract/templates/xml/article.xml
```diff
@@ -4,7 +4,7 @@
     <data:Number>${article.number}</data:Number>
     %if article.text:
     <data:Text>
-        <%include file="multilingual_m_text.xml" args="text=article.text"/>
+        <%include file="multilingual_text.xml" args="text=article.text"/>
```
voisardf, line 7: Should this be in this PR?

voisardf (approved):
Some xml templates in the PR that I don't think related? But if it's just a fast way to get them fixed, fine by me.
Looks good to me otherwise.

pyramid_oereb/lib/renderer/extract/templates/xml/article.xml
vuilleumierc: Thanks for your review, you are right this is not directly linked to the PR - I noticed there were three templates doing the same thing so I did some refactoring (50b988db5bda464a2b6d033a4f63b72c7409931f)

voisardf:
What about this PR. Is there just a second review missing?

vuilleumierc:
> What about this PR. Is there just a second review missing?

Yes, I was waiting for a second review before merging

vvmruder:
Did you checked if the extract is produced correctly with the "WITIMAGES" parameter?

vuilleumierc:
> Did you checked if the extract is produced correctly with the "WITHIMAGES" parameter?

Yes I checked, the images are in the extract with the parameter WITHIMAGES:
image
Is it OK if I merge this now?

vvmruder:
> > Did you checked if the extract is produced correctly with the "WITHIMAGES" parameter?
> 
> Yes I checked, the images are in the extract with the parameter WITHIMAGES:
> image
> Is it OK if I merge this now?

Hmm-The base64 looks really short. Could you try the following with the string:
```
import base64
coded_string = '''Q5YACgA...'''
base64.b64decode(coded_string)
```

vuilleumierc:
The base64 string is shortened in this view (using Firefox - see the three dots in the middle?)
Here is the decoded image:
image

vvmruder:
👍 

No Iam perfectly happy. Thank you! I missed the dots indeed :)

vvmruder (approved):

vuilleumierc:
OK thanks - no problem, it doesn't hurt to check the image properly for once :)

Issue #1247, opened 2021-07-05
Implement new foreignkey binding between PLR and LegendEntry


vvmruder:
Based on definition of the [Rahmenmodell](https://github.com/openoereb/pyramid_oereb/files/6442555/Rahmenmodell-de.pdf) page 15+16 we need to change DataModel and Records relations.

### TODO:

- implement relation in models
- implement relation in records
- implement changed behavior in StandardDB PLR Source
- Add convenient check if data is correct (maybe in the StandardDB PLR Source) in terms of definition on page 16 => only one geometry type per plr
- adapt and fix testing

PR #1258, opened 2021-07-08
1247-implement-new-foreignkey-binding-between-PLR-and-LegEntry

michmuel:
closes https://github.com/openoereb/pyramid_oereb/issues/1247
closes https://github.com/openoereb/pyramid_oereb/issues/1254
closes https://github.com/openoereb/pyramid_oereb/issues/1266


vvmruder:
As I see it:
Bildschirmfoto 2021-07-08 um 16 25 59

We still have simple foreign key in plr table to the view service. In addition we get a new foreign key relation between plr table and legend entry coming from the plr table. Because one PLR has then one legend entry defined be key relation all the attributes about type can be defined in legend entry and can be removed from the PLR table. As you can see on the picture (it is from the Rahmenmodell.pdf - Transferstruktur part) the plr is really a small table then. All is solved in the relation to the legend entry.

Does this help?

michmuel:
Screenshot from 2021-07-15 13-34-55

Diese Tests sind noch zu lösen oder hängen von anderen Issues ab.

michmuel:
Screenshot from 2021-07-15 13-34-55

pyramid_oereb/lib/records/plr.py
```diff
@@ -27,16 +27,17 @@ class PlrRecord(EmptyPlrRecord):
     Public law restriction record.
     """
 
-    def __init__(self, theme, information, law_status, published_from, responsible_office, symbol,
-                 view_service, geometries, sub_theme=None, type_code=None,
-                 type_code_list=None, basis=None, refinements=None, documents=None, info=None, min_length=0.0,
+    def __init__(self, theme, legende_text, law_status, published_from, published_until, responsible_office,
```
vvmruder, line 30: Here we the LegendEntryRecord should be passed instead of the legend_text

vvmruder: To keep it like the other record relations. You can access the legend text of LegendEntryRecord inside the init method to retrieve the content dict.

pyramid_oereb/lib/records/plr.py
```diff
@@ -62,31 +61,24 @@ def __init__(self, theme, information, law_status, published_from, responsible_o
...
+            warnings.warn('Type of "legende_text" should be "dict"')
 
         if sub_theme is not None and not isinstance(sub_theme, dict):
             warnings.warn('Type of "sub_theme" should be "dict"')
 
         assert isinstance(geometries, list)
         assert len(geometries) > 0
 
-        self.information = information
+        self.legende_text = legende_text
```
vvmruder, line 73: Attribute should be legend_text (english). To stay in existing code style.

pyramid_oereb/standard/models/airports_building_lines.py
```diff
@@ -287,7 +287,7 @@ class LegendEntry(Base):
     legend_text = sa.Column(JSONType, nullable=False)
     type_code = sa.Column(sa.String(40), nullable=False)
     type_code_list = sa.Column(sa.String, nullable=False)
-    topic = sa.Column(sa.String, nullable=False)
+    theme = sa.Column(sa.String, nullable=False)
```
vvmruder, line 290: 👍 good catch

pyramid_oereb/lib/records/plr.py
```diff
@@ -62,31 +61,24 @@ def __init__(self, theme, information, law_status, published_from, responsible_o
...
         self.law_status = law_status
         self.published_from = published_from
+        self.published_until = published_until
```
vvmruder, line 76: If you add this you maybe should link and solve also issue #1266 in your pull request

vvmruder (changes requested):
please have a look at the inline comments

voisardf (changes requested):
Just a typo to fix in several places. Looks good to me otherwise

pyramid_oereb/lib/records/plr.py
michmuel: I do not completely understand: is this wrong (also in the master https://github.com/openoereb/pyramid_oereb/blob/f148d34fa8bf29b1504c0d6e34aa5212b93578db/pyramid_oereb/standard/sources/plr.py#L309)?

pyramid_oereb/lib/records/plr.py
michmuel: Ok.

pyramid_oereb/lib/records/plr.py
vvmruder: I canged the code like I mean it. Please have a look.

michmuel:
Dann muss für https://github.com/openoereb/pyramid_oereb/blob/6bdceea3a7274620dd0ca46504869ffe43ed516d/pyramid_oereb/standard/sources/plr.py#L321 aber noch speziell ein LegendEntryRecord erstellt werden, oder?

vvmruder:
> Dann muss für
> 
> https://github.com/openoereb/pyramid_oereb/blob/6bdceea3a7274620dd0ca46504869ffe43ed516d/pyramid_oereb/standard/sources/plr.py#L321
> 
> aber noch speziell ein LegendEntryRecord erstellt werden, oder?

Ja, das muss man machen. Das läuft dann auch über eine (noch zu implementierende) Methode an der PLRSource `from_db_to_legend_entry_record`. Die Funktion die momentan so heisst:
https://github.com/openoereb/pyramid_oereb/blob/6bdceea3a7274620dd0ca46504869ffe43ed516d/pyramid_oereb/standard/sources/plr.py#L91-L103

muss umbenannt werden in `from_db_to_legend_entry_records`, weil sie eine Liste von Records zurück gibt. Das kannst du dann gleich mit lösen.

pyramid_oereb/lib/records/plr.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, theme, legende_text, law_status, published_from, published_un
         assert isinstance(geometries, list)
         assert len(geometries) > 0
 
-        self.legende_text = legende_text
+        self.legende_text = legend_entry.legend_text
```
michmuel, line 73: Do we need both or only the first one?
self.legend_entry = legend_entry
self.legende_text = legend_entry.legend_text

vvmruder: I would prefer first. And in Addition a method/property on the record class like this:

```
@property
def legend_text(self):
    return self.legend_entry.legend_text
```

michmuel:
Screenshot from 2021-07-19 16-14-27

michmuel:
Does it work like I expect that if "published_until" is NULL in the db None is reported in python?
https://github.com/openoereb/pyramid_oereb/blob/313f6aa88c0bb89ec725523d4bf8706c66d891f8/pyramid_oereb/lib/records/plr.py#L111

michmuel:
Screenshot from 2021-07-21 16-16-43

vuilleumierc:
Great work, thanks :+1:
I think there still a `topic` that needs to be changed to `theme` here: https://github.com/openoereb/pyramid_oereb/blob/d4bffd8b1882db3d6e3201a97cf93c8372e7f3fb/pyramid_oereb/standard/xtf_import/legend_entry.py#L44
(should fix the test `tests/xtf_import/test_legend_entry.py::test_parse`)

michmuel:
This one works with the suggested changes: tests/xtf_import/test_legend_entry.py

For pyramid_oereb/pyramid_oereb/standard/xtf_import/public_law_restriction.py I guess the attributes "legend_text", "theme", "sub_theme", "type_code" and "type_code_list" should be removed according to the Rahmenmodell. Right?

vuilleumierc:
I think you're right, I hadn't realized we were testing the transfer structure and not the extract here

vvmruder (approved):
Looks great now. Thank you for that heavy work trough! 👍

PR #1259, opened 2021-07-14
Remove DocumentBase reference from model templates

vuilleumierc:
After regenerating the standard models with `make create-default-models` I noticed there were remaining references to DocumentBase in the templates (and thus in the newly generated models) -> this PR is to correct this

### DIFF
--- a/pyramid_oereb/standard/templates/plr_integer_primary_keys.py.mako
+++ b/pyramid_oereb/standard/templates/plr_integer_primary_keys.py.mako
@@ -403,7 +403,7 @@ class PublicLawRestrictionDocument(Base):
             restriction.
         plr (pyramid_oereb.standard.models.${schema_name}.PublicLawRestriction):
             The dedicated relation to the public law restriction instance from database.
-        document (pyramid_oereb.standard.models.${schema_name}.DocumentBase):
+        document (pyramid_oereb.standard.models.${schema_name}.Document):
             The dedicated relation to the document instance from database.
     """
     __tablename__ = 'public_law_restriction_document'
@@ -416,7 +416,7 @@ class PublicLawRestrictionDocument(Base):
     )
     document_id = sa.Column(
         sa.Integer,
-        sa.ForeignKey(DocumentBase.id),
+        sa.ForeignKey(Document.id),
         nullable=False
     )
     plr = relationship(
@@ -424,6 +424,6 @@ class PublicLawRestrictionDocument(Base):
         backref='legal_provisions'
     )
     document = relationship(
-        DocumentBase
+        Document
     )
     article_numbers = sa.Column(sa.String, nullable=True)
--- a/pyramid_oereb/standard/templates/plr_string_primary_keys.py.mako
+++ b/pyramid_oereb/standard/templates/plr_string_primary_keys.py.mako
@@ -403,7 +403,7 @@ class PublicLawRestrictionDocument(Base):
             restriction.
         plr (pyramid_oereb.standard.models.${schema_name}.PublicLawRestriction):
             The dedicated relation to the public law restriction instance from database.
-        document (pyramid_oereb.standard.models.${schema_name}.DocumentBase):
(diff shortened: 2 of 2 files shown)


pyramid_oereb/standard/templates/plr_integer_primary_keys.py.mako
```diff
@@ -416,14 +416,14 @@ class PublicLawRestrictionDocument(Base):
...
+        sa.ForeignKey(Document.id),
         nullable=False
     )
     plr = relationship(
         PublicLawRestriction,
         backref='legal_provisions'
     )
     document = relationship(
-        DocumentBase
+        Document
```
voisardf, line 427: I think this relation drops too as there is no more link from DocumentBase to Document? Or am I'm wrong and it's the reference of a legal provision to the legal base relation?

pyramid_oereb/standard/templates/plr_string_primary_keys.py.mako
```diff
@@ -416,14 +416,14 @@ class PublicLawRestrictionDocument(Base):
...
+        sa.ForeignKey(Document.id),
         nullable=False
     )
     plr = relationship(
         PublicLawRestriction,
         backref='legal_provisions'
     )
     document = relationship(
-        DocumentBase
+        Document
```
voisardf, line 427: same question here

voisardf (commented):
Nice catch! thanks. Please check my comment.

kdeininger (approved):
Thanks! 👍

pyramid_oereb/standard/templates/plr_integer_primary_keys.py.mako
vuilleumierc: Yes good point !

vvmruder: @kdeininger What do you think? I'am pro to @voisardf the relation from document to document is gone. So relation can be removed. Can you check this in detail?

vvmruder:
@vuilleumierc We need to check against a fixed model from master branch for valid relations.

pyramid_oereb/standard/templates/plr_integer_primary_keys.py.mako
kdeininger: Yes, there are no relationships between documents anymore.

vvmruder:
Can you adpt yout PR regarding to this example:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/models/airports_security_zone_plans.py#L96-L147

pyramid_oereb/standard/templates/plr_integer_primary_keys.py.mako
kdeininger: Sorry, checked the wrong relationship. This relationships is at the PLR and is still needed. The class has changed from `DocumentBase` to `Document`. So everything should be fine.

vuilleumierc:
> Can you adapt your PR regarding to this example:
> https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/models/airports_security_zone_plans.py#L96-L147

Sorry I am not sure what there is to adapt - if I check class PublicLawRestrictionDocument in the models on master, I think the template in this PR is consistent, what do you think?
https://github.com/openoereb/pyramid_oereb/blob/080ae00316f0411598e6fa45cddb26300c8daf42/pyramid_oereb/standard/models/airports_security_zone_plans.py#L392-L428

voisardf:
thanks for checking. Then everything should be allright and ready to be merged 👍

vvmruder:
Eh? I lost track. Sorry...

vvmruder:
Ok. Got it. It was about the class `PublicLawRestrictionDocument`

Yes you can merge. Your suggested code ist correct.

vvmruder (approved):

PR #1267, opened 2021-07-15
Create Theme element

### LINKED ISSUE #1256: Create Theme element in pyramid_oereb ecosystem
Because the subtheme is now of the type Theme (Record) in the extract we need some additional construction to get the data from somewhere.

The decision is to implement it via a table in the main schema and have a semi/soft  Foreignkey relation. This means we need an additional Model in the `pyramid_oereb.standard.models.main.py`

This means:

- implement a model fitting the needs of the already existing theme record (attributes code and text, where text is multilingual so JSONType in model)
- implement a fitting source (get idea how this might work from the glossary database source)
- implement a fitting reader (get idea how this might work from the glossary reader)
- get everything together in the processor => This means before everything else get all records of themes, and pass them to the PLR-Source, they will be needed there => This means to adapt complete behavior of the PLRSource from its real base class to all its child implementations.

### Pseudocode

Model
------

class Theme():
  id = PrimaryKey
  code = string + unique constraint
  text = JSONType

vuilleumierc:
Closes #1256 
New: theme data are moved from `PublicLawRestriction` to a new `Theme` table. The link between the tables goes through the theme code (e.g. `LandUsePlans`), which is used as a soft foreign key (The theme codes are not yet updated to the new specification - it will be done separately after the merging of #1248). The themes are now read from the DB at the application startup, when the configuration is initialized (instead of at each request). 
With this PR the section `theme` is now required in the YAML configuration file for the configuration to be initialized

### DIFF
--- a/pyramid_oereb/lib/readers/extract.py
+++ b/pyramid_oereb/lib/readers/extract.py
@@ -138,7 +138,7 @@ class ExtractReader(object):
 
         else:
             for plr_source in self._plr_sources_:
-                themes_without_data.append(Config.get_theme(plr_source.info.get('code')))
+                themes_without_data.append(Config.get_theme_by_code(plr_source.info.get('code')))
 
         # Load base data form configuration
         resolver = DottedNameResolver()
--- a/pyramid_oereb/lib/config.py
+++ b/pyramid_oereb/lib/config.py
@@ -9,7 +9,8 @@ from pyramid.config import ConfigurationError
 from pyramid_oereb.lib.adapter import FileAdapter
 from pyramid_oereb.lib.records.office import OfficeRecord
 from pyramid_oereb.lib.records.image import ImageRecord
-from pyramid_oereb.lib.records.theme import ThemeRecord
+from pyramid_oereb.lib.readers.theme import ThemeReader
+from sqlalchemy.exc import ProgrammingError
 
 log = logging.getLogger(__name__)
 
@@ -22,6 +23,7 @@ class Config(object):
     """
 
     _config = None
+    themes = None
 
     @staticmethod
     def init(configfile, configsection, c2ctemplate_style=False):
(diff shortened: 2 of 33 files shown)


pyramid_oereb/lib/readers/extract.py
```diff
@@ -138,7 +138,7 @@ def read(self, params, real_estate, municipality):
 
         else:
             for plr_source in self._plr_sources_:
-                themes_without_data.append(Config.get_theme(plr_source.info.get('code')))
+                themes_without_data.append(Config.get_theme_by_code(plr_source.info.get('code')))
```
vvmruder, line 141: I noticed, you switched functions here. In the original `Config.get_theme` method there was a important detail=>It respects the order how the themes are configured in the YML. This means this order is still present in the extract. We should not lose this behavior. Sorry I forgot about that when we talked it through.

vvmruder: @kdeininger What do you think about this. In the new approach we have also stored the order of themes in the data. Can we rely only on this now? Or should we have a dualistic approach where data order AND the YML-Config order is respected?

vvmruder (commented):
I would like to put this PR on hold to see @kdeininger s opinion.

kdeininger (approved):

pyramid_oereb/lib/readers/extract.py
kdeininger: A dualistic approach might by confusing. I think we should use the new attribute `extract_index`.

vvmruder (approved):

vuilleumierc:
Thanks for checking - I just noticed I forgot to change the placeholder text in the new class description, so I fixed this and will merge now

vvmruder:
Ok. I crosschecked some of the behavior and its working like a charm 😍 ! Great work.

PR #1268, opened 2021-07-15
New class for document type texts

voisardf:
Implementing the new class for the multilingual document type values


voisardf:
ok, I got the extract working with the new approach, but I have two questions:
Do we really want this format for DocumentType (code & text)?
py_oereb_doc_type_format
or just the text?

And for now I put the lookup in the renderer:
https://github.com/openoereb/pyramid_oereb/pull/1268/files#diff-a5b38687817e31d60774e39a77d7be5e9a4c6249176c69fe7e130e8144569bc4R332
Should I rather put it in the `get_document_type_by_code` method in the config.py?
https://github.com/openoereb/pyramid_oereb/pull/1268/files#diff-f546f3cb0b29a3e1a5190c0051fd19122b6509307f5bdb58ce46e464bc0ea787R141

vvmruder:
> ok, I got the extract working with the new approach, but I have two questions:
> Do we really want this format for DocumentType (code & text)?
> py_oereb_doc_type_format
> or just the text?
> 
> And for now I put the lookup in the renderer:
> https://github.com/openoereb/pyramid_oereb/pull/1268/files#diff-a5b38687817e31d60774e39a77d7be5e9a4c6249176c69fe7e130e8144569bc4R332
> Should I rather put it in the `get_document_type_by_code` method in the config.py?
> https://github.com/openoereb/pyramid_oereb/pull/1268/files#diff-f546f3cb0b29a3e1a5190c0051fd19122b6509307f5bdb58ce46e464bc0ea787R141

First: Spec says on page 11/26 in OEREB-Data-Extract.pdf that there is a ComplexType `DocumentType` with Code and Text attribute. And then there is a SimpleType `DocumentTypeCode` which delivers only the simple type. I really don't know whats the matter here. Maybe @kdeininger has some idea?

vvmruder:
To answer your other question: I think we should try to keep things together in the Config. So I would move the lookup in the config.

voisardf:
I've got 4 tests failing. I guess that is fine?

vvmruder (approved):
Good job. I did some quick checks and think it's ready to merge.

vuilleumierc:
Looks good, thanks for the work!

> Good job. I did some quick checks and think it's ready to merge.

As mentioned in the chat, in your local instance (for the XML extract), isn't the `code` in English for other languages than German? It's the case for me. My understanding is that it should not be multilingual (see Rahmenmodell, plus that's the case for the JSON extract (always in German)). Could someone maybe check that on their local instance?
image

vvmruder:
> Looks good, thanks for the work!
> 
> > Good job. I did some quick checks and think it's ready to merge.
> 
> As mentioned in the chat, in your local instance (for the XML extract), isn't the `code` in English for other languages than German? It's the case for me. My understanding is that it should not be multilingual (see Rahmenmodell, plus that's the case for the JSON extract (always in German)). Could someone maybe check that on their local instance?
> image

Ok. I will check again.

vvmruder:
This is how it looks for me in `de`:
Bildschirmfoto 2021-07-23 um 07 52 02

And this is how it looks in `it`:
Bildschirmfoto 2021-07-23 um 07 52 57

So in this case it works like expected.

I will check about the model again.

vuilleumierc:
Thanks for checking! Seems to work fine indeed. I guess that's something I have locally, even though I cleaned everything (or so I think..)

vuilleumierc (approved):

vvmruder:
Ok.

In addition I checked [Ramenmodell.pdf](https://github.com/openoereb/pyramid_oereb/files/6442555/Rahmenmodell-de.pdf) which states that DocumentType has a multilingual text now:
Bildschirmfoto 2021-07-23 um 07 56 57

 and [OEREB-Data-Extract_de.pdf](https://github.com/openoereb/pyramid_oereb/files/6442562/OEREB-Data-Extract_de.pdf) which also has this definition:
Bildschirmfoto 2021-07-23 um 07 58 21

We can do a short call, maybe I'am missing something here.

vuilleumierc:
I think that's fine - the `text` attribute is multilingual, no doubt about that. My question was about the `code`, which should not be and I was surprised is what sometime displayed in German (`LANG=de`), sometimes in English (`LANG=fr`) on my local instance.
(On a side note I am surprised the code are listed in English OEREB-Data-Extract_de.pdf, I have seen them in German in the Rahmenmodell)

vvmruder:
> I think that's fine - the `text` attribute is multilingual, no doubt about that. My question was about the `code`, which should not be and I was surprised it was sometimes displayed in German (`LANG=de`), sometimes in English (`LANG=fr`) on my local instance.
> (On a side note I am surprised the code are listed in English OEREB-Data-Extract_de.pdf, I have seen them in German in the Rahmenmodell)
> Thanks again for checking and sorry for the confusion

No problems. Good that you are curious 👍 

I think the difference in the documents we can put on the stash of federal mixups.

voisardf:
As I see it, I think the code should always be in english according to the model, or am I wrong?

@vvmruder @vuilleumierc  If that is ok with you I'll merge this PR now so we can progess with the project. And if there is a problem with our way to set the code I'll open a new issue and treat the fix seperately.

vuilleumierc:
OK to merge :+1:
For me it's not clear if it should be English or German but your implementation is fine

voisardf:
I just checked with Rolf Zürcher and it's like the theme codes or any code/id: The value in https://models.geo.admin.ch/V_D/OeREB/OeREBKRM_V2_0_Texte_20210714.xml is the one it should be. So it's German in that case :)

Issue #1223, opened 2021-06-03
Geometries become LV95 only


vvmruder:
---

*OLD*
* Geometries could be in LV03 or LV95

*NEW*
* Geometries are in LV95

##### Consequences:
---
* This has an effect on the import script.

jwkaltz:
Closed via #1272

PR #1272, opened 2021-07-20
Drop LV03 support and upgrade to pyproj 2

### LINKED ISSUE #869: Update pyproj
There are now some new versions of pyproj available. The version downgrade done in PR #810 is probably not needed anymore.

Issue depending on https://gitlab.com/gf-bl/python-reprojector/issues/2, because this project also restricts the version used.

vuilleumierc:
Closes #869 

When fixing the tests (there were some floating point adaptations required) I removed one which was with LV03 and which was also failing due to some floating point inaccuracy (bbaae2220baba008886dc416117953c411238306) (As I understood, pyramid_oereb won't support LV03)

In case anyone else is wondering about these coordinates:
https://github.com/openoereb/pyramid_oereb/blob/bbaae2220baba008886dc416117953c411238306/tests/webservice/test_getegrid.py#L71
It comes from the fact that all test geometries are in the range (0,0) - (10,10) but with SRID=2056 (LV95). Because of the way the SRID is evaluated in the webservice.py: https://github.com/openoereb/pyramid_oereb/blob/bbaae2220baba008886dc416117953c411238306/pyramid_oereb/views/webservice.py#L496-L498
for the coordinates (2,0), the SRID is assumed to be LV03 and they are reprojected and the real estate geometry is not intersected. Probably something to keep in mind when/if dropping support for LV03.

### DIFF
--- a/pyramid_oereb/contrib/sources/address.py
+++ b/pyramid_oereb/contrib/sources/address.py
@@ -72,7 +72,7 @@ class AddressGeoAdminSource(AddressBaseSource):
                 for item in data.get('results'):
                     attrs = item.get('attrs')
                     if isinstance(attrs, dict) and attrs.get('origin') == 'address':
-                        x, y = rp.transform((attrs.get('lon'), attrs.get('lat')), to_srs=srid)
+                        x, y = rp.transform((attrs.get('lat'), attrs.get('lon')), to_srs=srid)
                         self.records.append(AddressRecord(
                             street_name=street_name,
                             zip_code=zip_code,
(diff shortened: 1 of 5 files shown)


voisardf (approved):
Looks fine to me as far as I can tell. thanks

vuilleumierc:
As suggested by @vvmruder this PR now also addresses #1223 (in webservice.py and part of the tests, not in xtf_import which will be done in a further step)

vvmruder (approved):
Thank you for this. I think you can merge it

PR #1277, opened 2021-07-22
General information handling

### LINKED ISSUE #1264: Information handling
The information is configured in the yml today:
https://github.com/openoereb/pyramid_oereb/blob/080ae00316f0411598e6fa45cddb26300c8daf42/pyramid_oereb/standard/pyramid_oereb.yml.mako#L436-L440

The new [XML](http://models.geo.admin.ch/V_D/OeREB/OeREBKRM_V2_0_Texte_20210714.xml) containing text definitions provided by federation offers better possibilities then storing this mapping hardcoded in the yml.

The new approach should be like for municipalities and glossaries.

New Model in the main.py with model name:

- `Information `

and attributes:

- `code` (PrimaryKey, String)
- `text`(multilingual/JSON)

In addition the places where usage and reference to this exists need to be cleaned and updated to support this approach:
https://github.com/openoereb/pyramid_oereb/blob/787ce93db13fa9af2b775b9a80a7b4078fee56a3/pyramid_oereb/lib/readers/extract.py#L149

Test data needs to be adapted. The content can be consumed out of the XML file.

The new approach should use a reader and a source as it is used for e.g Glossary.

A config part should be added to the yml like it exists for e.g. Glossary:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/pyramid_oereb.yml.mako#L380-L391

Dedicated methods should offer access to the information stored in database directly at the config object. This could be `get_information` which returns a dict containing the data as defined in the XML.

If necessary other dedicated methods at the config object should be added.

vuilleumierc:
Handle general information in DB + Config
Closes #1264


pyramid_oereb/lib/readers/general_information.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+        source_class = DottedNameResolver().maybe_resolve(dotted_source_class_path)
+        self._source_ = source_class(**params)
+
+    def read(self):
+        """
+        The central read accessor method to get all desired records from configured source.
+        .. note:: If you subclass this class your implementation needs to offer this method in the same
+            signature. Means the parameters must be the same and the return must be a list of
+            :ref:`api-pyramid_oereb-lib-records-document_types-generalinformationrecord`.
```
voisardf, line 32: typo, should probably be : `api-pyramid_oereb-lib-records-general_information-generalinformationrecord` or something like that

pyramid_oereb/lib/readers/general_information.py
```diff
@@ -0,0 +1,40 @@
...
+            Otherwise the API like way the server works would be broken.
+        Returns:
+            list of pyramid_oereb.lib.records.document_types.GeneralInformationRecord:
```
voisardf, line 35: same here, should not be `.document_types.`

pyramid_oereb/lib/renderer/extract/templates/xml/extract.xml
```diff
@@ -73,11 +73,9 @@
     %if extract.qr_code:
         <data:QRCode>${extract.qr_code}</data:QRCode>
     %endif
-    %if extract.general_information:
-        <data:GeneralInformation>
-            <%include file="multilingual_text.xml" args="text=extract.general_information"/>
-        </data:GeneralInformation>
-    %endif
+    %for general_information in extract.general_information:
```
voisardf, line 76: why did you remove the condition?
As far as I remember, general information is never empty. But I might be wrong

pyramid_oereb/lib/sources/general_information.py
```diff
@@ -0,0 +1,25 @@
...
+from pyramid_oereb.lib.sources import Base
+from pyramid_oereb.lib.records.general_information import GeneralInformationRecord
+
+
+class GeneralInformationBaseSource(Base):
+    """
+    Base class for GeneralInformation values source.
+
+    Attributes:
+        records (list of pyramid_oereb.lib.records.document_types.GeneralInformationRecord): List of general
```
voisardf, line 11: document_types pops up again :)

voisardf (changes requested):
Some typos (copy/paste) to fix

pyramid_oereb/lib/sources/general_information.py
vuilleumierc: Thanks ! I'm caught, I've been copy-pasting your PR :)

pyramid_oereb/lib/renderer/extract/templates/xml/extract.xml
vuilleumierc: The difference is that it is now a list (because it comes from the DB and we take the whole table) even though there should be exactly one element there. So I changed the `if` condition for a `for` loop. I think it's the case for many other things in the extract (lists with only one element, always).
But maybe we should implement it so that is can only have one element (-> would not be a list anymore)? I am not sure at which "layer" (source, reader, etc) it would best to do it but we can do it for sure. What do you think?

pyramid_oereb/lib/sources/general_information.py
voisardf: it's fair game - I did it before on your 'theme' PR 😄

pyramid_oereb/lib/renderer/extract/templates/xml/extract.xml
voisardf: I think it is at least one element but it could be a list. So that is fine. But as you said, we might want to check elsewhere that there is at least one element or throw an error.
But that is not a high priority.

@vvmruder do we want to put this on a 'beautify and check exceptions' issue list once the main work is done?

vvmruder: @voisardf I think we should fix this in place now. And your suggestion is good. Lets implement a throw to assure minimal one element is there.

vvmruder (approved):
I think @voisardf requested changes are good. I already approve here :)

vuilleumierc:
Thanks for your suggestions @voisardf @vvmruder 
I added a `ConfigurationError` if the general information table is empty (https://github.com/openoereb/pyramid_oereb/pull/1277/commits/6b99b6ef10f6be2cce9a22fbcaaa5ff15165898c). Is it OK like this, or is this too strict?
I tested the extract request and `make test` and everything seems OK but feel free to test again

voisardf (approved):
Thanks, I think it's ready to be merged now :) (well, once the conflicts resolved 🙈 )

vuilleumierc:
I guess it was my turn :)

PR #1281, opened 2021-07-27
basic model, sources, record and reader implementation for logos

voisardf:
PR to implement multilingual logos read from the data base


voisardf:
ok, as I see it (from https://github.com/openoereb/pyramid_oereb/files/6442562/OEREB-Data-Extract_de.pdf p. 15), in the new data extract, the logo objects are either given as base64 binary or as URI. But neither of it is defined multilingual in the data-extract definition.
Do you agree? @vvmruder @jwkaltz @kdeininger

vvmruder:
You are right. Then I would go for the solution to solve multilinguality in rendering step and deliver it like it is requested (with images, or not).

vvmruder:
One last thing fails. The Image records mime type is detected wrongly. So Extract does not work-Iam on it. Otherwise you can start to have a look.

vvmruder:
Found the problem... Test data and probably the Images from FED are wrong. At least the RM lang type of plr symbol is JPG which is not allowed.

pyramid_oereb/lib/renderer/extract/json_.py
```diff
@@ -8,6 +8,7 @@
 from pyramid.testing import DummyRequest
 
 from pyramid_oereb import Config, route_prefix
+from pyramid_oereb.lib import get_multiligual_element
```
vuilleumierc, line 11: I just notice now that there is a typo in this function name (multiligual -> multilingual)

tests/test_config.py
```diff
@@ -64,63 +63,34 @@ def test_get_plr_cadastre_authority():
...
-    if language in ['fr', 'it']:
-        logos = Config.get_logo_config(language=language)
-        assert isinstance(logos, dict)
-        logo_oereb = logos.get('oereb')
-        assert logo_oereb.content == FileAdapter().read(Config.get('logo').get('oereb').get(language))
-
-    else:
-        with pytest.raises(HTTPInternalServerError):
-            assert Config.get_logo_config(language)
+    Config.init('./pyramid_oereb/standard/pyramid_oereb.yml', 'pyramid_oereb')
```
vuilleumierc, line 89: Why not use the test config anymore?

pyramid_oereb/standard/models/main.py
```diff
@@ -58,6 +58,20 @@ class Theme(Base):
     extract_index = sa.Column(sa.Integer, nullable=False)
 
 
+class Logo(Base):
+    """
+    The container for all logos and municipality coat of arms
+
+    Attributes:
+        code (str): The identifier given by a code
+        logo (str): The image encoded in base64
```
vuilleumierc, line 67:
```suggestion
        logo (dict): The image encoded in base64 (multilingual)
```

vuilleumierc (approved):
Tested here and it works well, thanks! I have only a few minor comments

tests/test_processor.py
```diff
@@ -30,6 +32,9 @@
     'EGRID': 'TEST'
 }
 
+Config._config = None
+Config.init('./pyramid_oereb/standard/pyramid_oereb.yml', 'pyramid_oereb')
+
```
vuilleumierc, line 37: Same question as above

pyramid_oereb/lib/renderer/extract/json_.py
vvmruder: I fixed this for all appearances.

tests/test_config.py
vvmruder: True-I changed it for that :)

tests/test_processor.py
vvmruder: Fixed that

vvmruder:
I fixed all, I think. Its ready for a new round :)

vuilleumierc (approved):
Perfect, thanks, LGTM

PR #1286, opened 2021-08-05
Improve DevContainer

vvmruder:
add zsh as in container standard shell

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -11,4 +11,9 @@ RUN apt-get update && \
     python3-venv \
     virtualenv \
     ${DEV_PACKAGES} \
+    zsh \
     postgresql-client
+
+ENV SHELL /bin/zsh
+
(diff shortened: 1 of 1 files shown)


marionb (approved):
It works. The auto completion is working well on the first word of the command but id there is a second part it is a bit odd (like `make` is recognized but then it stops). It is still an improvement :-D

kdeininger:
Maybe the zsh-completions need to be installed:
https://github.com/zsh-users/zsh-completions

Alternatively, we could use bash.

vvmruder:
If I use `make c` and then press tab I get `clean` and `clean-all` as candidates. What would you suggest would be the right behavior?

marionb:
> If I use make c and then press tab I get clean and clean-all as candidates. What would you suggest would be the right behaviour?

that would be the behaviour I'd expect after a restart of my pc that seems to also be working now :+1: (was not expecting to have to restart).

vvmruder:
@kdeininger I thought zsh would be nice because it is more integrated with git for instance and also offers many plugins for a feature rich behavior

PR #1287, opened 2021-08-05
Add new extract format "URL"

### LINKED ISSUE #1184: New format kind in URL
__GetExtractById Service:__
---

New format kind in URL

**OLD:**

- xml
- json
- pdf

`https://example.com/oereb/extract/{FORMAT}/CH887722167773`

`https://example.com/oereb/extract/xml/CH887722167773`
`https://example.com/oereb/extract/json/CH887722167773`
`https://example.com/oereb/extract/pdf/CH887722167773`

**NEW:**

- xml
- json
- pdf
- url

`https://example.com/oereb/extract/{FORMAT}/CH887722167773`

`https://example.com/oereb/extract/xml/CH887722167773`
`https://example.com/oereb/extract/json/CH887722167773`
`https://example.com/oereb/extract/pdf/CH887722167773`
`https://example.com/oereb/extract/url/CH887722167773`

If format is URL then a 303 redirect is expected to the actual cantonal dynamic
client instance. This behaviour is completely new and needs a new configurable 
URL for redirection.

kdeininger:
Closes #1184

### DIFF
--- a/pyramid_oereb/standard/pyramid_oereb.yml.mako
+++ b/pyramid_oereb/standard/pyramid_oereb.yml.mako
@@ -485,6 +485,9 @@ pyramid_oereb:
     sort_within_themes_method: pyramid_oereb.standard.hook_methods.plr_sort_within_themes
     # Example of a specific sorting method:
     # sort_within_themes_method: pyramid_oereb.contrib.plr_sort_within_themes_by_type_code
+    # Redirect configuration for type URL. You can use any attribute of the real estate RealEstateRecord
+    # (e.g. "{egrid}") to parametrize the URL.
+    redirect: https://geoview.bl.ch/oereb/?egrid={egrid}
 
   # All PLRs which are provided by this application. This is related to all application behaviour, especially
   # the extract creation process which loops over this list.
--- a/pyramid_oereb/standard/pyramid_oereb.yml.mako
+++ b/pyramid_oereb/standard/pyramid_oereb.yml.mako
@@ -486,7 +486,7 @@ pyramid_oereb:
     # Example of a specific sorting method:
     # sort_within_themes_method: pyramid_oereb.contrib.plr_sort_within_themes_by_type_code
     # Redirect configuration for type URL. You can use any attribute of the real estate RealEstateRecord
-    # (e.g. "{egrid}") to parametrize the URL.
+    # (e.g. "{egrid}") to parameterize the URL.
     redirect: https://geoview.bl.ch/oereb/?egrid={egrid}
 
   # All PLRs which are provided by this application. This is related to all application behaviour, especially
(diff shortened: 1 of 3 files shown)


pyramid_oereb/views/webservice.py
```diff
@@ -549,6 +556,22 @@ def __has_params__(self, needed):
...
+        Args:
+            real_estate (pyramid_oereb.lib.records.real_estate.RealEstateRecord):
+                The found real estate.
+
+        Returns:
+            pyramid.httpexceptions.HTTPFound: The redirect response.
+        """
+        url = Config.get_extract_config().get('redirect')
+        if url is None:
+            raise HTTPInternalServerError('Missing configuratioin for redirect to dynamic client.')
```
voisardf, line 572: typo: configuration

pyramid_oereb/standard/pyramid_oereb.yml.mako
```diff
@@ -485,6 +485,9 @@ pyramid_oereb:
     sort_within_themes_method: pyramid_oereb.standard.hook_methods.plr_sort_within_themes
     # Example of a specific sorting method:
     # sort_within_themes_method: pyramid_oereb.contrib.plr_sort_within_themes_by_type_code
+    # Redirect configuration for type URL. You can use any attribute of the real estate RealEstateRecord
```
voisardf, line 488: I would think only identifiers like EGRID or IDENTDN&NUMBER should be possible or am I wrong?

voisardf (commented):
thanks. A typo and a question. Looks good to me otherwise.

pyramid_oereb/standard/pyramid_oereb.yml.mako
kdeininger: I am not sure. The whole real estate object is passed to the method, so all attributes are available to parameterize the URL. Maybe some canton needs also the type of the real estate. But I think in most cases, the `egrid` will be enough.

What do the others think? Should we restrict the available attributes to `egrid`, `identdn` and `number`?

vvmruder: I think EGRID is enough because we have a service to obtain it with other identifiers.

vvmruder (approved):
Looks good. To repeat my comment above: I think EGRID is fine for this because we have already a Service getEGRID to offer a way to get it by other parameters.

kdeininger:
> Looks good. To repeat my comment above: I think EGRID is fine for this because we have already a Service getEGRID to offer a way to get it by other parameters.

I think, the question is how it is implemented in the dynamic client. For example, our client accepts only `egrid` as parameter. But there are a lot of different dynamic clients and they might use different parameters. Maybe we even need add a hook method to generate the URL, e.g. if the client needs the coordinates of the real estate.

vvmruder:
Ok. Then maybe we could add all possible parameters a configurable. If I understood you correctly we have all possible RealEstate attributes available. So it would be possible to use them vor configuration.

kdeininger:
Yes, currently we can use all attributes of the real estate record to generate the URL. I think, this should be sufficient in most cases and could be extended by adding a hook method if really needed.

vvmruder:
My approval is there already
:)

PR #1290, opened 2021-08-09
Chatch non Existing Default Lang. Values

### LINKED ISSUE #1275: Catch server crash on not existing default language
If the server is configured with a default language which is actually not in the data, the server crashes. This should be handled with a catch. It should gracefully issue a message to the client that default language was not found in the data. A log message error should be added to the server logs.

marionb:
Fixes #1275

TO TEST:
In order to test you can:
- Set the default language to 'it' 
     1. some values do not exist and it will cause an error with an entry in the logs if the default language is requested
     2. requesting 'de' will still return an extract because there is nothing missing

TODO:
- [x] handle `logo` if not available in the default language
- [x] handle `reference_wms` if the reference is not available in the default language
- [x] handle `rendering` of multilingual values in xml and json if the default language is not available
- [x] write tests

TO DISCUSS:
For the flowing point, what is currently in the code is a suggestion:
- [x] What error should be thrown and the exact error message
- [x] placement of the function `get_multilingual_element` -> perhaps there is a better place for this function. (I did not want to place it in the renderer becaus it is also used in the config https://github.com/openoereb/pyramid_oereb/blob/2883840b1894c823806c661a65da1a3041479c04/pyramid_oereb/lib/config.py#L534 and the view_sevice https://github.com/openoereb/pyramid_oereb/blob/48f468eea4c7d452262232cd63472e9d394d1b4d/pyramid_oereb/lib/records/view_service.py#L237


tests/renderer/test_base.py
```diff
@@ -83,6 +83,32 @@ def test_get_localized_text_from_dict(language, result):
...
+        'it': u'Questo è un test'
+    }
+
+    if language in ['en', 'it'] or not disallowNull:
+        localized_text = renderer.get_localized_text(multilingual_text, disallowNull)
+        assert isinstance(localized_text, dict)
+        assert localized_text.get('Language') in ['de', 'it', 'en']
+        assert localized_text.get('Text') == result
+    else:
+        with pytest.raises(HTTPInternalServerError):
```
vuilleumierc, line 108: :+1: I did not know this function

vuilleumierc (approved):
Nice work, thank you, and thanks for the thorough description :+1:
It works as expected here.

pyramid_oereb/lib/__init__.py
```diff
@@ -1 +1,18 @@
...
+
+
+def get_multiligual_element(value, language, not_null=True):
+    multilingual_value = value.get(language, None)
+
+    if multilingual_value is None and not_null:
+        msg = 'Default language "{language}" is not available in: \n {element}'\
+            .format(language=language, element=value)
+        log.error(msg)
+        raise HTTPInternalServerError()
```
vvmruder, line 16: I like this error and I think we can keep it. 👍

pyramid_oereb/lib/__init__.py
```diff
@@ -1 +1,18 @@
 # -*- coding: utf-8 -*-
+import logging
+
+from pyramid.httpexceptions import HTTPInternalServerError
```
vvmruder, line 4: Placing of the method is fine.

vvmruder (approved):
👍 Great work and it will make the server more stable. Thank you.

Issue #1222, opened 2021-06-03
Theme display by legal state and effect


vvmruder:
---
In the new version, topics have now three possible legal states. In the extract all three states were delivered.
In the PDF extract, each legal state should appear on a new page (like a new topic).
The legal state is mentionned as a subtitle (multilingual)

*OLD:*
* in force
* ongoing modification

*NEW:*
* in force
* ongoing modification with immediate effects
* ongoing modification without immediate effects

We need to discuss this point, as has a major implication on the refactoring:
* should we create a seperate topic schema for each legal state
or
* keep one schema for each topic and do the legal state filtering while serving the data

##### Consequences:
---
* Code needs to be refactored - work load depends on the approach

Issue #1194, opened 2021-05-07
New behavior based on law status


vvmruder:
The federal specification defines 3 law statuses instead of 2.

They are:

- in_force
- running_modifications
- running_modifications_(mit Vorwirkungen) => how to translate???

Until now pyrmid_oereb only have this three places where it takes care of the difference (in_force/running_modifications):

- [Docker configuration](https://github.com/openoereb/pyramid_oereb/blob/ab16f87ae728769d4cdf5f5d57a0fa42218dd80c/docker/config.yaml.tmpl#L28)
- [Example configuration](https://github.com/openoereb/pyramid_oereb/blob/787ce93db13fa9af2b775b9a80a7b4078fee56a3/pyramid_oereb/standard/pyramid_oereb.yml.mako#L28) Zeile 28
- [Example configuration](https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/pyramid_oereb.yml.mako#L459) Zeile 459
- [Config Object](https://github.com/openoereb/pyramid_oereb/blob/787ce93db13fa9af2b775b9a80a7b4078fee56a3/pyramid_oereb/lib/config.py#L454)
- [docs](https://github.com/openoereb/pyramid_oereb/blob/7458324e2241a27c495f9f46c5b52908f352ab8f/doc/source/changes.rst#L998)

pyramid_oereb only delivers this content in the requested language until now.

### What needs to be done

Translation needs to be updated in the config:

- `in_force = in Kraft` should be changed to `in_force = Rechtskräftig`
- `running_modifications = laufende Änderungen` should be changed to `running_modifications_with_??? = Änderung ohne Vorwirkung`
- `running_modifications_without??? = Änderung mit Vorwirkung`

In the future approach pyramid_oereb must produce topics on runtime based on data content. This means:

1. before some calculation takes place we need to iterate over the data which might be affected
2. group data into 3 groups  in_force/running_modifications/running_modifications_???
3. create new topics based on the groups we have (if a group has 0 (zero) data elements we don't need a new topic
4. all calculation can be done like before

[...]

PR #1297, opened 2021-08-20
Group PLRs by law status

vuilleumierc:
Issue #1194 and #1222


voisardf:
Code looks good to me, but I can not get the pdf call working despite following the instructions outside the dev container.
It seems that there is a problem with the database creation/connection

voisardf:
I can confirm now, that we are getting a pdf based on the new data. thanks @vuilleumierc !
ping me once discussed improvments are ready for review

vuilleumierc:
Thanks for testing @voisardf 
As the ConcernedTheme list in the data extract does not have any law_status attribute, I implemented something in the print proxy where we delete this list and recreate it based on the public law restriction list. But maybe someone has a better suggestion. Thanks in advance for reviewing!

pyramid_oereb/lib/readers/extract.py
```diff
@@ -129,11 +129,7 @@ def read(self, params, real_estate, municipality):
                 # Filter topics due to topics parameter
                 if not params.skip_topic(plr.theme.code):
                     if isinstance(plr, PlrRecord):
-                        contained = False
-                        for theme in concerned_themes:
-                            if theme.code == plr.theme.code:
-                                contained = True
-                        if not contained:
+                        if plr.theme.code not in [theme.code for theme in concerned_themes]:
```
vuilleumierc, line 132: This should not change the logic at all and is not directly linked to this PR, but I struggled to understand this piece of code so I re-wrote this in a way which hopefully is more understandable for others as well

sample_data/groundwater_protection_zones/legend_entry.json
```diff
@@ -1,9 +1,9 @@
 [
   {
     "legend_text": {
-      "fr": "Pollué, investigation nécessaire",
-      "de": "Belastet, untersuchungsbedürftig",
-      "it": "Inquinato, è necessario procedere a un?indagine"
+      "fr": "Zone de captage (zone S1)",
```
vuilleumierc, line 4: Same here: not directly linked to the PR, but I thought our sample data could be a little more realistic :)

PR #1298, opened 2021-08-23
Simplified approach for Make/Docker usage

vvmruder:
- remove devcontainer settings
- directly use docker compose for better compatibiity
- use makefile only as scope inside running container
- fix some tests
- improve readme


Makefile
```diff
@@ -58,7 +58,7 @@ PACKAGE = pyramid_oereb
 	touch $@
 
 .venv/requirements-timestamp: .venv/timestamp setup.py requirements.txt requirements-tests.txt dev-requirements.txt
-	$(VENV_BIN)/$(PIP_COMMAND) install --upgrade pip
+	$(VENV_BIN)/$(PIP_COMMAND) install --upgrade pip wheel
```
vvmruder, line 61: This seems to imrove speed of venv build time a little

vuilleumierc:
Here is what I tried (after cleaning Docker containers and images):

```
docker-compose build
docker-compose up -d
docker-compose exec oereb-server bash
make serve-dev
```

And I get a Python error:

```
sql -h oereb-db -U postgres -d pyramid_oereb_test -f .db/12-create.sql
touch .db/.create-db-dev-tables
.venv/bin/python pyramid_oereb/standard/load_sample_data.py --configuration pyramid_oereb/standard/pyramid_oereb.yml --sql-file .db/13-fill.sql --dir sample_data
  File "pyramid_oereb/standard/load_sample_data.py", line 334, in <module>
    _run()
  File "pyramid_oereb/standard/load_sample_data.py", line 329, in _run
    sql_file=sql_file
  File "pyramid_oereb/standard/load_sample_data.py", line 37, in __init__
    Config.init(self._configuration, self._section, c2ctemplate_style)
  File "/workspace/pyramid_oereb/lib/config.py", line 50, in init
    Config.init_logos()
  File "/workspace/pyramid_oereb/lib/config.py", line 183, in init_logos
    Config.logos = Config._read_logos()
  File "/workspace/pyramid_oereb/lib/config.py", line 192, in _read_logos
    raise ConfigurationError("Missing configuration for logos")
pyramid.exceptions.ConfigurationError: Missing configuration for logos
make: *** [Makefile:107: .db/13-fill.sql] Error 1
```

On my side I would keep the docker-* targets as I find it much more convient as doing all the above commands. Otherwise I think it's a good thing to remove devcontainer

vvmruder:
> _read_logos

As always before you should do a `make clean-all serve-dev` to remove old configuration. This wasn't changed (https://github.com/openoereb/pyramid_oereb/blob/master/Makefile#L133).

vuilleumierc:
> > _read_logos
> 
> As always before you should do a `make clean-all` to remove old configuration when you freshly roll out a server or run tests after a you switched branches. This wasn't changed (https://github.com/openoereb/pyramid_oereb/blob/master/Makefile#L133).

Then this should be in the readme, and solidifies my point that this way of starting the server is cumbersome.

vvmruder:
but isnt it the same in the docker-Target solution?

vuilleumierc:
No, we could already set up the server in the Dockerfile, and just start the pserve command when doing `docker-compose up`

vvmruder:
but the changed pyramid_oereb.yml stays until cleaned.

vuilleumierc:
That's right, if you look in version 1, the rendering of the config.yml was done when starting the container.

vvmruder:
To make a point here: Your call of docker-compose calls out of the Makefile are nothing else but linking make targets out of the container. So the behavior of Make doesn't change at all. To work completely this way means you always need to recreate containers all the time you change anything at the venv (because its baked in: https://github.com/openoereb/pyramid_oereb/blob/master/Dockerfile#L20-L22). As you can see in addition this unnecessarily copies lines defined by the Makefile (https://github.com/openoereb/pyramid_oereb/blob/master/Makefile#L56-L63) which is hard to maintain and leads to unpredictable behavior once something is changed. Especially because the target [docker-serve](https://github.com/openoereb/pyramid_oereb/blob/master/Makefile#L183-L185) hides that the venv creation was done already while image was built. This is hard to understand because you offer the target [docker-clean-all](https://github.com/openoereb/pyramid_oereb/blob/master/Makefile#L204-L208) which does restart the db-container but does not recreate the .venv in the oereb-server container because the corresponding Make-dependencies are not checked. To reproduce:

```
make docker-serve
make docker-clean-all
```

So using the docker-Target is misleading and yet not documented at all.

To conclude: I can understand that you like your approach. But since its not documented-barely someone else then you understands how to use it.

vuilleumierc:
I never said my solution was the best one, I also agree that to have only one scope for the Makefile would be best. But this scope should be out of the container in my opinion, and the image should be a product of the build process, as we did in version 1 and it worked well. So I got to this hybrid solution because I wanted to preserve the way the project was built for other developers. As for the documentation, I wrote it in the PR and in Gitter for other developers to see. I can update the readme if you want but I don't really see the point if you are to remove those targets.
But anyway, I think I already spoke more than enough on this and will let others voice their opinion.
By the way the connection with the print seems to work on this branch, but there is a error in the print. It is not clear where it comes from but I think @marionb is investigating

PR #1302, opened 2021-08-24
Setup development environment for both local and docker workflows

danduk82:
This PR enhances the docker composition and Makefile to allow both docker-based and local development on the project.

In particular:
- use default `postgresql-client` environment variables both in Makefile and docker-compose.yml, such as `PGHOST` insead of `PG_HOST` for example
- set default values both in Makefile and docker-compose.yaml file, but if these variables are present on the host system it will use the ones that are already set. This is useful is we work with an already existing database, or a shared database.
- it adds the possiblity to change the port on which pyramid_oereb is listening (default: 5432) by setting the environment variable `PYRAMID_OEREB_PORT`. For this reason the `development.ini` file is now a mako template (the requrirements have been updated)
- add the `development.ini` target in the Makefile
- it fixes the requirement of `DROP DATABASE` at the startup

@vvmruder fyi it is based on your `refine_make_docker_setup` branch


vvmruder:
@danduk82 Thank you for this PR-I will have a look soon.

Issue #1299, opened 2021-08-23
Logos are not PDF/A conform


marionb:
Logos no longer seem to be no PDF/A conform. This is breaking the possibility to print a PDF for the moment.

**Note:**
- In the Logo.xml from the confederation, the ÖREB-Logo for de is PNG, the other languages in JPEG! That is probably the issue, as jpeg has not transparency
- Simplest workaround for now in my opinion is to update the sample data replacing the jpgs with the german png data
- To fix this we need an update from the base data from Swisstopo (a first request has been sent by @voisardf )

PR #1305, opened 2021-08-27
replaced fr, it, rm and en plr logo with png base64 encoded

voisardf:
This should fix issue https://github.com/openoereb/pyramid_oereb/issues/1299 at least until we get a conforming logo.xml from the confederation.
I'm just checking it now locally so you might wan't to wait until this afternoon before testing/validating


voisardf:
@marionb : I will have a look into it - maybe there is still some jpg format out there...

voisardf:
if the `<property name="net.sf.jasperreports.export.pdfa.conformance" value="pdfa1a"/>` line is activated in the mfp template, I get an error. So there is still a pdf compliance issue for now

vuilleumierc (approved):
When turning off PDF-A, I can get a static extract in French, which is already an improvement. Thanks!

Issue #1210, opened 2021-06-03
PDF: general points


voisardf:
With the latest specifications the PDF format version has been updated to use the version PDF/A-2a by default. Also, the error messages to return if a PDF can not be produced have been defined.

### What needs to be done:
- [x] Set version to PDF/A-2a (or PDF/A-1 (ISO 19005-1:2005) if not possible) -_> stay at PDF/A-1, as JasperReports support this standard only_
- [ ] If a data set is not available due to a server or a service failure, a predefined message (multilingual) has to be returned:
DE: «Ein oder mehrere ÖREB-Themen stehen momentan nicht zur Verfügung. Daher kann kein Auszug
erstellt werden. Versuchen Sie es zu einem späteren Zeitpunkt erneut. Wir entschuldigen uns für die
Unannehmlichkeiten.»
FR: «Un ou plusieurs thèmes RDPPF sont momentanément indisponibles. L’extrait ne peut donc pas
être établi. Veuillez réessayer plus tard. Nous vous prions de nous excuser pour ce désagrément.»
The error handling exists in the print porxis for MFP. An adaption of the current text `a problem occurred while generating the pdf file` is needed depending on the language. See also https://github.com/openoereb/pyramid_oereb/blob/f5d65b35e6ab85cdaad54cf10ed6d00aac9fd549/pyramid_oereb/contrib/print_proxy/mapfish_print.py#L134-L137

jwkaltz:
@voisardf regarding "Set version to PDF/A-2a (or PDF/A-1 (ISO 19005-1:2005) if not possible)":

Our current underlying reporting technology, https://github.com/TIBCOSoftware/jasperreports, does not support PDF/A-2. 
The Weisung says "Der statische Auszug wird im Dateiformat PDF/A-1 ... oder PDF/A-2a ... erzeugt".

Reading that, it seems that we can simply stay with PDF/A-1, which we implemented with pyramid_oereb 1.5, see notes here: https://openoereb.github.io/pyramid_oereb/changes.html#mapfish-print-pdf-a-conformance

What do you think?

voisardf:
@jwkaltz: Yes, the specifications allows to stay with PDF/A-1, and if the new standard is not supported yet by Jasper Reports we do not have a lot of options.
But we should keep an eye on that and 'upgrade' if Jasper Reports starts to support the new version.

jwkaltz:
> we do not have a lot of options.

[...]

PR #1312, opened 2021-09-02
Use specific error message in print proxy

vuilleumierc:
Closes https://github.com/openoereb/pyramid_oereb/issues/1210 (part 2)
I left the original error messages in the server logs, as I think it would be too confusing otherwise.
As already mentioned I think this error message should be returned for any error ocurring during a static extract request (not just in the print proxy) in order to conform to the specification.
Thanks in advance for reviewing

To test:
Turn on PDF-A in `oereb-print` and request http://localhost:6543/oereb/extract/pdf?EGRID=CH113928077734

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print.py
@@ -71,6 +71,9 @@ class Renderer(JsonRenderer):
         if 'lang' in self._lowercase_GET_dict:
             self._language = self._lowercase_GET_dict.get('lang')
 
+        self._static_error_message = Config.get('static_error_message').get(self._language) or \
+            Config.get('static_error_message').get(self._fallback_language)
+
         # Based on extract record and webservice parameter, render the extract data as JSON
         extract_record = value[0]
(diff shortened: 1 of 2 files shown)


pyramid_oereb/contrib/print_proxy/mapfish_print.py
```diff
@@ -71,6 +71,8 @@ def __call__(self, value, system):
         if 'lang' in self._lowercase_GET_dict:
             self._language = self._lowercase_GET_dict.get('lang')
 
+        self._static_error_message = Config.get('static_error_message').get(self._language)
```
marionb, line 74: If the error message does not exist in the requested language I think this will crash or give an error.
You can avoid this when you use the fallback_lang for this case.

vuilleumierc: Good catch, I thought it was already handled above :+1:

voisardf (approved):
looks good to me, thanks!

marionb (approved):

vvmruder (approved):
Works as expected. This can be merged.

PR #1313, opened 2021-09-02
Fix geometry model

### LINKED ISSUE #1310: Fix Geometry model
The Geometry model still implements the relation to the Office. This is not in the new datamodel. We need to remove it and fix all problems.

This needs to be removed:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/models/theme.py#L310-L315

vvmruder:
closes #1310


voisardf (approved):
seems all good. thanks!

Issue #1311, opened 2021-09-02
Check png transparency and pdf/A-1 compliance


voisardf:
As follow-up to PR https://github.com/openoereb/pyramid_oereb/pull/1305 we still have an PDF format compliance issue.
We need to make sure the used logos and all other binary elements in the static extract are PDF/A-1 compatible.

voisardf:
Maybe we should move this issue to the pyramid_oereb_mfp project?
Unless the WMS requests returns map images with transparency?

vuilleumierc:
New logos to test (2021.09.03): 
[OeREBKRM_V2_0_Logos_20210903.txt](https://github.com/openoereb/pyramid_oereb/files/7152344/OeREBKRM_V2_0_Logos_20210903.txt)

vuilleumierc:
I decoded the images from OeREBKRM_V2_0_Logos_20210903.xml and there is still transparency in the ch.plr logos (all languages) so it is still not possible to comply with PDFA-1. Here are the images:

![ch de](https://user-images.githubusercontent.com/18527151/133260968-5514b6a1-325b-4ae7-9f98-9152d8d45093.png)
![ch fr](https://user-images.githubusercontent.com/18527151/133260971-bde96039-5bfb-4470-bc46-9bf46f9bf374.png)
![ch it](https://user-images.githubusercontent.com/18527151/133260973-01dfeaeb-18ae-4b38-b6c1-6ef5c9c068b6.png)
![ch plr de](https://user-images.githubusercontent.com/18527151/133260978-e8f1d367-4d28-4a59-82b4-8b615c3f8f28.png)
![ch plr en](https://user-images.githubusercontent.com/18527151/133260981-f20633e2-1bab-4fbd-b343-fbff3b72c283.png)
![ch plr fr](https://user-images.githubusercontent.com/18527151/133260982-85c62c95-47f9-475a-89f0-e0fee18e376a.png)
![ch plr it](https://user-images.githubusercontent.com/18527151/133260986-27f40632-95e4-4483-bc24-919410330de9.png)
![ch plr rm](https://user-images.githubusercontent.com/18527151/133260990-0742023a-84c5-4c5f-b781-3b3c2b46cee1.png)
![ch rm](https://user-images.githubusercontent.com/18527151/133260993-22e4e59e-8fad-4f06-9633-87d0b276551b.png)

Issue #1296, opened 2021-08-20
Some logos are not PNG 


vvmruder:
The logos have to be PNG or SVG regarding to the specification. Some of them seems not to be. Especially the plr-logo in language RM seems to be JPG.

What needs to be done:
Check all logos in https://github.com/openoereb/pyramid_oereb/blob/master/sample_data/logo.json and https://github.com/openoereb/pyramid_oereb/blob/master/tests/init_db.py to be PNG. Fix it to PNG where necessary. Gather which image was not here in the Issue comments to be able to inform federal offices to fix this.

PR #1329, opened 2021-09-15
Remove transparency from logos

vuilleumierc:
After careful inspection it appeared that there was an alpha channel in all Confederation logos (ch and ch.plr) and in our sample canton logo. This was the problem blocking the PDFA-1 compliance. 
Here are the images without transparency:
[logos-no-alpha.zip](https://github.com/openoereb/pyramid_oereb/files/7168319/logos-no-alpha.zip)
 ~~This PR adds them in our sample data but they should still be updated on the Confederation's side (OeREBKRM_V2_0_Logos_XXXXXXXX.xml)~~

Edit: @voisardf I replaced the ch and ch.plr logos with the images from OeREBKRM_V2_0_Logos_20210915.xml 
Here it works fine for de, fr, and it but not rm. The image ch.plr.rm does not seem to be valid (I cannot decode it)

Closes https://github.com/openoereb/pyramid_oereb/issues/1311
Closes https://github.com/openoereb/pyramid_oereb/issues/1299
Closes https://github.com/openoereb/pyramid_oereb/issues/1296


voisardf:
@vuilleumierc : ping me once the new logo file is integrated and I'll check

voisardf:
there you go 😉 @vuilleumierc

vuilleumierc:
@voisardf Works now in rm, thanks a lot! For en I get 400 bad request but this has nothing to do with the logos https://github.com/openoereb/pyramid_oereb/blob/93c3ccc03f9af8e9e6a5f9f6767977982e4bbc16/pyramid_oereb/standard/pyramid_oereb.yml.mako#L14-L21
Should we merge this?

voisardf:
@vuilleumierc: Good to go for me

PR #1330, opened 2021-09-15
Adapt mapfish print test to v2 extract

### LINKED ISSUE #1288: Adapt print proxy to new model
Some of the attributes passed in to the print templates now come as multilingual dictionaries. The proxy needs to be adapted to be able to extract the proper language and pass only the needed value to the templates/ MFP engine.
This should also fix the currently failing tests `test_mapfish_print_entire_extract`.

vuilleumierc:
Closes #1288


marionb (approved):
LGTM!
Thanks

PR #1338, opened 2021-10-14
Use modelfactory for OEREBlex themes

### LINKED ISSUE #1314: Use modelfactory for OEREBLex
To have less redundant code, also oereblex models should implement the approach of model factory.

kdeininger:
Replace OEREBlex models with modelfactory.

Closes #1314


pyramid_oereb/contrib/models/oereblex/theme.py
```diff
@@ -0,0 +1,273 @@
...
+        office_at_web = Column(JSONType, nullable=True)
+        uid = Column(String(12), nullable=True)
+        line1 = Column(String, nullable=True)
+        line2 = Column(String, nullable=True)
+        street = Column(String, nullable=True)
+        number = Column(String, nullable=True)
+        postal_code = Column(Integer, nullable=True)
+        city = Column(String, nullable=True)
+
+    class DataIntegration(Base):
```
michmuel, line 89: Perhaps some more classes in https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/standard/models/theme.py could be packed into functions and used here as much code is equal in both files.

sample_data_OeREBKRMtrsfr_V2_0/oereb_fed_test.sql
```diff
@@ -2971,8 +2971,8 @@ OeREBKRM_V2_0.Dokumente.Dokument.Typ	typ	dokument	\N
 --
 
 COPY contaminated_civil_aviation_sites.t_ili2db_basket (t_id, dataset, topic, t_ili_tid, attachmentkey, domains) FROM stdin;
-2	1	OeREBKRM_V2_0.Dokumente	\N	OeREBKRM_V2_0_Gesetze_20210414.xml-2	
-81	80	OeREBKRMtrsfr_V2_0.Transferstruktur	\N	interlis.xtf-81	
+2	1	OeREBKRM_V2_0.Dokumente	\N	OeREBKRM_V2_0_Gesetze_20210414.xml-2
```
michmuel, line 2974: What was the issue here?

pyramid_oereb/contrib/models/oereblex/theme.py
michmuel: Maybe analogue to the Interlis-Modell, the classes "Document" and "Office" in one function (as it is already; the two classes are used in the main.py) and the other classes in another function.

vvmruder: I know, it might be overkill. But instead of grouping them, maybe we offer them each in a function? So any possible future reuse will be covered already.

vvmruder (commented):
Looks good-see my comment

michmuel (approved):

PR #1339, opened 2021-10-14
Language sensitive geolink cache

### LINKED ISSUE #1025: Bug when using improvments of pull request #993 in multilingual setting
I had some troubels with the geolink request when using pyramid_oereb Version 1.7.3 in a multilingual environment. 
The toubels are caused by the improvement implemented by the pull request #993 "use per topic -store- to verify already queried documents" .

The problem shows when switching to the "second" language. Because the geolink still has the same id, the document is not requested again an therefore the language is not chaged. 

For the moment I solved the problem by using a costomised plr_oereblex.py source. I added the following code to the "document_records_from_oereblex function" to make the imporovements of pull request #993 language sensitiv:

        if params.language is None:
            lexlink_id=lexlink
        else:
            lexlink_id=str(lexlink) + params.language

This works fine for me for the moment but I think an imporevements should be included in pyramid_oereb/contrib/sources/plr_oereblex.py

kdeininger:
Closes #1025

### DIFF
--- a/pyramid_oereb/contrib/sources/document.py
+++ b/pyramid_oereb/contrib/sources/document.py
@@ -25,7 +25,7 @@ class OEREBlexSource(Base):
 
         Keyword Args:
             host (uri): Host URL of OEREBlex (without /api/...).
-            version (str): The used geoLink schema version. Default is 1.2.0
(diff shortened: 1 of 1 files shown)


michmuel (approved):
Looks fine for me.

PR #1384, opened 2021-11-19
fix-oereblex (adaptions to changes in config.py)

michmuel:
Functions "get_law_status_by_law_status_code" and "get_document_type_by_code" not available anymore in config.py. Therefore, they were replaced.


vvmruder (approved):

vuilleumierc:
Hi @michmuel

The JSON extract requests seem to work fine with this. For the XML extract (http://localhost:6543/oereb/extract/xml?EGRID=CH113928077734), I get a 204 "no content" status code. Same thing when I request a PDF. Is this new behavior intentional, or am I missing something?

I cannot run the tests either, after trying out several `make` commands I still get these errors on `make test`:
```
ERROR tests/init_db.py - sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) null value in column "id" of relation "real_estate_type" violates not-null constraint
ERROR tests/renderer/test_xml.py - pyramid.exceptions.ConfigurationError: Logo for code: ch.plr not found in the application configuration
```

What should I do to run the tests? It would be nice to update the readme as well, it is also not up-to-date on how to run `make serve`.

Regarding the code, I am probably missing some context, but as far as I can tell the function `get_law_status_by_law_status_code` is still used in the tests (files tests/records/test_document.py and tests/records/test_plr.py) - wouldn't it be a good time to replace these calls as well?

As for the function `get_document_type_by_code` it seems it's still defined in config.py: https://github.com/openoereb/pyramid_oereb/blob/250fcf9cf0a8fac87201338361f2486109461e25/pyramid_oereb/lib/config.py#L564 - so why replace it?

michmuel:
Hi @vuilleumierc 
The documentation is not up-to-date as the start of the application is a temporary construct.
Start-up works by the command "bash start_oereb-server_with_sample_data_OeREBKRMtrsfr_V2_0.sh" (after docker-compose up).

You are right the functions are still used in the tests but there is no error when running "make test".

I will replace the functions in a new commit.

vuilleumierc:
Thanks for your feedback. I did run `bash start_oereb-server_with_sample_data_OeREBKRMtrsfr_V2_0.sh` (otherwise I couldn't run `make serve`) - it did not help with running the tests though. It starts the server directly, so I kill it and try `make serve` but I get the aforementioned errors.
As for the temporary construct, the readme can be updated anytime.

michmuel:
The bash script loads some fed-data in a separate database. Maybe - after docker-compose up - you run only lines 3-5 of the script and the use the normal commands docker-compose make test.

You are right concerning the readme.

michmuel:
Letting the config read the pyramid_oereb.yml for the tests is not ideal, right?

vuilleumierc:
Thanks for editing init_db.py, I can run the tests now :+1: I get 131 failed, some due to the function calls we already mentioned.
For the test config I think we had some dedicated YAML files, for instance tests/resources/test_config.yml

michmuel:
Ok, I will add one plr to tests/resources/test_config.yml.

michmuel:
Do you agree when I merge this PR and we fix the tests after Clemens' outsourcing of the test-db to a separate project/container?

PR #1385, opened 2021-11-25
restructure project

vvmruder:
As a part of the clean up and as preparation for more logical testing this reorganization should provide a better and more logic file structure.


michmuel:
The scripts "create_tables.py", etc. in https://github.com/openoereb/pyramid_oereb/tree/reorganize_project/pyramid_oereb/contrib/data_sources/standard work in parts also for other data sources than standard, right? So, maybe they could be placed in data_sources/.

michmuel:
https://github.com/openoereb/pyramid_oereb/tree/reorganize_project/pyramid_oereb/contrib/data_sources/standard/xtf_import
I am not sure if this is still required for some users and if it is working after all our changes.

vvmruder:
> https://github.com/openoereb/pyramid_oereb/tree/reorganize_project/pyramid_oereb/contrib/data_sources/standard/xtf_import I am not sure if this is still required for some users and if it is working after all our changes.

As far as I understand, this is not necessary any longer since we can use Interlis directly now. So Usecase should be obsolete. Right?

Makefile
```diff
@@ -252,11 +246,10 @@ $(DEV_CREATE_TABLES_SCRIPT) $(DEV_CREATE_STANDARD_YML_SCRIPT): setup.py $(BUILD_
 	$(VENV_BIN)/python $< develop
 
 .PHONY: build
-build: install $(DEV_CREATE_TABLES_SCRIPT) $(DEV_CREATE_STANDARD_YML_SCRIPT)
+build: install $(DEV_CREATE_TABLES_SCRIPT) $(DEV_CONFIGURATION_YML) create_dev_db_scripts
```
vuilleumierc, line 249: Shouldn't the build also depends on the target `development.ini`? I get a file not found error if I don't call it explicitly

docker-compose.yml
```diff
@@ -1,13 +1,10 @@
...
+version: "3.7"
 
 services:
   oereb-server:
     build:
       context: .
       dockerfile: Dockerfile
-    depends_on:
-      oereb-db:
-        condition: service_healthy
```
vuilleumierc, line 10: Is there a particular reason for removing the health check?
This is an issue for me because when `oereb-server` first tries to connect with `oereb-db`, the DB is not up yet -> the server crashes with a Python error

docker-compose.yml
```diff
@@ -44,9 +34,7 @@ services:
     ports:
       - ${EXPOSED_PGPORT:-15432}:${PGPORT:-5432}
```
vuilleumierc, line 35: Using only default values, I initially got a mismatch for the exposed DB port (15432 / 5432). I suggest to change it here (in the docker-compose) to 5432 since the default value is 5432 everywhere else

vuilleumierc (changes requested):
I tested here and it almost worked, see my comments for details.
After these modifications I tested:

- JSON extract: is produced
- XML extract: is produced
- PDF extract: almost OK, fails because transparency in some images -> probably not something to fix in this PR (have the federal images been corrected on their side?)
- `make test-core`: runs correctly, many failed tests but again I guess this is not the topic of this PR

Makefile
vvmruder: Thats true. Thankyou.

docker-compose.yml
vvmruder: I tried to use a newer docker-compose version (3.7) to be able to support the `user` setting to sync in container user with host user. This way we wouldn't have file generated on host mount as root. But this is  [WIP]

docker-compose.yml
vvmruder: I'm not sure I got what you mean by that.

vuilleumierc: I meant replace all occurrences of `15432` to `5432` (I couldn't do it with code suggestion as you did not edited this part in your PR). The goal would be that the build works with the specified default values (without needing to specify a port)
I can do it in another PR if necessary

docker-compose.yml
danduk82: @vvmruder  you can change the command of the container to use `exec` with a wait condition that checks if the database is available with a timeout. Something like 
```bash
exec wait-for -t 120 <some-health-check-code-that-points-to-postgres> -- <your-usual-startup-command-here>
```
 as workaround. Otherwise it should be the main application (pyramid_oereb wsgi) that should wait until the DB is up at startup with a timeout as well.

docker-compose.yml
vvmruder: Ahh-got it. I will change that :)

docker-compose.yml
vvmruder: Thats true. I will see if I find a simple solution to empower pyramid_oereb to check when DB is available.

docker-compose.yml
vvmruder: I did update. Please have a look :)

docker-compose.yml
vvmruder: I did an update. Please have a look.

Makefile
vvmruder: Also did update this. Can you recheck?

vuilleumierc (approved):
Works for me :+1:

voisardf:
Regarding the PDF issue with an image, I think that one of the example municipality logos is not yet fixed to png without transparency

docker-compose.yml
danduk82: I checked your code, it seems good to me, thanks

vvmruder:
It was the NE logo :see_no_evil:

michmuel (approved):

voisardf:
ooops 🙈 sorry!

Issue #1381, opened 2021-11-11
provide logs when configuration is faulty


jwkaltz:
The V2 should always provide at least a log entry of level warning or error when the configuration is faulty.

Candidates for improvements: 
* config.py

Possible improvements:
* At the moment a `ConfigurationError` is thrown if a config table does not exists in the DB, but not if the table exists and is empty -> add a log entry or an error for such cases

svamaa:
https://github.com/openoereb/pyramid_oereb/pull/1869

PR #1389, opened 2021-11-27
Add more infos upon error

jwkaltz:
Partially addresses #1381 

Notably regarding the get_multilingual_element, adding a stack trace, to show where the call came from.


voisardf (approved):
looks good to me, thanks

Issue #1293, opened 2021-08-13
module contrib/stats is broken in v2


danduk82:
It requires `c2cwsgiutils` in version 3 or superior. I suppose that it is enough to add `c2cwsgiutils` in the requirements.

jwkaltz:
Solved in #1391

PR #1415, opened 2021-11-30
Temporary deactivation of stats / see #1293

jwkaltz:
Temporarily deactivate stats functionality to avoid side-effects in error logging, until full repair in #1293


jwkaltz:
Merging without review in order to do a new beta release for live test

PR #1418, opened 2021-12-01
Remove symlinks

vvmruder:
Fix to support Windows again


vvmruder:
Please check if i missed some.

voisardf (approved):
seems to work for me, symlinks error gets resolved

svamaa:
The symlink problem seems to be solved for me. Didn't get the issue again.

Issue #1414, opened 2021-11-30
Prevent duplicated documents in the extract


michmuel:
Documents for a particular PLR are combined from different data sources such as Gesetze.xml (main schema) and database/OerebLex depending on the theme. It may happen that documents are available in more than one data source and are duplicated in the extract. This might cause issues for the oereb clients.

If the information from Gesetze.xml shall has priority (right?), following adaption can solve the problem:
Here (https://github.com/openoereb/pyramid_oereb/blob/26596efd7924437c4d4c6e34060cd30e45b00765/pyramid_oereb/standard/sources/plr.py#L327), all the indexes of the documents from the main schema can be queried and the documents from the function get_document_records can be filtered accordingly.

vvmruder:
If I remember correctly this is the case already now. Filter them is a task for client (dynamic, PDF,...). @kdeininger Was it not an issue, that this links were shown some time ago by accident? And we decided this is wiped by client?

kdeininger:
I think we have two different cases:

1. Documents, especially laws, are linked to each restriction and therefore they may be contained multiple times in the XML/JSON extract (identical titles and URLs). This case should be handled by die Client, which is already the case.
2. Documents provided by the federation can be linked to a topic directly. The same document may also be delivered by OEREBlex. In this case, at least the URLs but usually also the titles are different from the corresponding documents of the federation. That is why the current matching by URL and title fails.

A possible solution could be to adapt the OEREBlex source to replace all federal documents using the official indexes.

michmuel:
For the oereb_server being queried from a third party application, I think it would be better to not send duplicated documents as this application would have to perform the sorting and filtering as well.

kdeininger:
There may be duplicate documents due to multiple restrictions of the same topic, but the list of documents for each restriction should be unique.

PR #1420, opened 2021-12-03
eliminate duplicated records (for a plr) in the extract

michmuel:
Closes https://github.com/openoereb/pyramid_oereb/issues/1414

Criteria for corresponding document records: same index + same document_type code + same number


kdeininger (approved):
Looks good :+1:

vvmruder (changes requested):
Maybe add some logging for debug when a document is marked as duplicate and is not used any more. I already imported the logging.

vvmruder (approved):

PR #1423, opened 2021-12-06
BugFix GetEGRID

vvmruder:
fixes #1422

PR #1424, opened 2021-12-06
Update README.rst

vvmruder:
Add more precise start commands

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,19 +4,11 @@
 
 Temporary Startup of the dev server:
 
-Required package: ``xsltproc``
+1. ``docker build -t pyramid_oereb:dev .``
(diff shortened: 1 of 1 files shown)


vuilleumierc:
Could you give a bit of context please? (why are the build instructions completely different than last week - did I miss a PR?)

vvmruder:
Based on that https://github.com/openoereb/pyramid_oereb/pull/1423 and the linked issue I thougt this is a more save way to describe it. Especially because xsltproc is not necessary and it might work on Windows too (I try to find someone to approve that). However this is still the temporary way to start it and not that different. I only use the built image to produce the SQL files to fill the db instead of doing it locally.

vuilleumierc (approved):

svamaa (approved):
tip top!

Issue #1421, opened 2021-12-03
Fix core tests


vuilleumierc:
Restructure core tests following new project structure and fix broken tests

vuilleumierc:
Duplicate from  #1390

Issue #1396, opened 2021-11-29
Fix and enhance test for contrib.print_proxy.mapfish_print


vvmruder:
Due to new project structure its possible to run test for specific parts of the application. This task is for the contrib.print_proxy.mapfish_print.

- move tests depending to core part into the new dir
- fix all broken tests
- add tests where necessary to raise coverage

PR #1425, opened 2021-12-07
Fix test initialization

vuilleumierc:
Issue #1421 (partial implementation)
The aim of this PR is to be able to run `pytest` (in the current state on master, `pytest` is interrupted at startup due to errors).
This is a first step in fixing the core tests for v2 and includes:

- new directory structure to reflect the new project structure
- introduction of fixtures to set up the test config and test data
- various fixes in imports which were broken after the project restructuring (WIP)
- update of the test configuration files (mainly `tests/resources/test_config.yml`) to v2 (WIP)


vvmruder:
Thank you. I looked at it and its like I expected. But I suggest to use a folder naming like `contrib.print_proxy.mapfish_print` for the contrib packages. So we could easily adress tests directly to a contrib package doing a `PYTEST_PATH=contrib.print_proxy.mapfish_print make tests` and have an overwiew of all contrib tests right away.

In addition I suggest to remove all __init__.py files in the test folders recursively. This really slows down the testing and we agreed to don't use this pattern.

What do you think?

vvmruder:
I added a suggestion how it could look like.

vuilleumierc:
> Thank you. I looked at it and its like I expected. But I suggest to use a folder naming like `contrib.print_proxy.mapfish_print` for the contrib packages. So we could easily adress tests directly to a contrib package doing a `PYTEST_PATH=contrib.print_proxy.mapfish_print make tests` and have an overwiew of all contrib tests right away.
> 
Yes, I agree, I didn't do it in this PR because I originally targeted only the core tests, but in the end I had to fix some thing also in contrib otherwise pytest would not start as mentioned. Sorry if this was confusing. I suggest addressing it when we start with the dedicated issue #1396

> In addition I suggest to remove all **init**.py files in the test folders recursively. This really slows down the testing and we agreed to don't use this pattern.
> 
> What do you think?

Yes I agree, that's precisely the goal of using fixtures and I already removed the main ones (For the remaining ones I again suggest doing it with dedicated PR as this can quickly be time consuming. The goal of this PR was only to have `pytest` running again so that we can starting working on the tests and splitting the work among us)

vuilleumierc:
@vvmruder I hadn't noticed you had directly pushed to my branch - I guess the suggestion about the directory structure is now solved :+1:

vuilleumierc:
@vvmruder is it OK to merge in this state (so that we can start splitting further work) or are there things that you want we fix now?

vvmruder (approved):
Yes. this is fine to me

PR #1427, opened 2021-12-08
Fix core tests

vuilleumierc:
Issue  #1390


tests/core/readers/conftest.py
```diff
@@ -0,0 +1,287 @@
...
+
+    view_services = {
+        'view_service99': models.ViewService(
+            id=99,
+            reference_wms=wms_url_contaminated_sites,
+            layer_index=1,
+            layer_opacity=.75,
+        )
+        }
+    dbsession.add_all(view_services.values())
```
mki-c2c, line 42: for all occurrences of add_all:
is there a benefit of creating a dict and then using .values() to get the value iterator ?

Wouldn't it be easier to use a list like:
```
   view_services = [
        models.ViewService(
            id=99,
            reference_wms=wms_url_contaminated_sites,
            layer_index=1,
            layer_opacity=.75,
        )
        }
    dbsession.add_all(view_services)
```

mki-c2c: or even organize the imports differently so that it is possible to copy paste the code and only make a search - replace operation:

my idea:
original: 
```
        connection.execute(contaminated_sites.Office.__table__.insert(), {
            'id': '1',
            'name': {'de': u'Test Office'}
        })
```
fixture:
```
        dbsession.add(Office(**{
            'id': '1',
            'name': {'de': u'Test Office'}
        })
```
so you only have to replace `.__table__.insert(), ` by `(**` and `connection.execute` by `dbsession.add(`
and somehow mange the model import...

voisardf: I think that joins the idea from Clemens as I understood it: having always the same kind of load process and to put the json part to fill the db tables in a file. But I might be wrong with my understanding @vvmruder

vuilleumierc: That would indeed save a lot of time for the migration and it seems it would still be possible to use the fixture for asserts in the test - new PR https://github.com/openoereb/pyramid_oereb/pull/1429/

vvmruder: Yes. The idea was to have data (also for tests) in the database already. This way we can improve it easily and even see it through e.g. QGIS. Should I support you by adding the data as JSONS into the sampte-data structure?

jwkaltz: This morning in the meeting the consensus was a different one... But we can rediscuss it if needed, a live meeting would be needed I think.

vvmruder: Ah-ok then nevermind

vvmruder: I only thought we want get rid of the in python dynamic db interactions.

mki-c2c: @vvmruder we started heading into the direction where CI tests are conducted on a lightweight disposable DB. See PR https://github.com/openoereb/pyramid_oereb/pull/1431 for a fixture of this temp DB.
This temporary DB can also be consulted from QGIS if:
- the transaction - rollback mechanism in the tests is deactivated so that everything is committed to the DB (could be a parameter when launching the test script)
- a breakpoint is set somewhere in the test sequence before the DB is dropped. (or we could introduce a parameter like DB_NO_DROP)
I consider this type of tests suitable for CI since setup and teardown are really fast and few data needs to be put into the repo.

However it does not exclude functional tests on a larger, static DB created from json files. But maybe these tests should not be run systematically in CI ?

It would be good to discuss and clarify this point quite quickly, hopefully we can organize a meeting with Wolfgang.

vvmruder: No its ok for me. I agree on this approach if ti fits the needs :)

vvmruder:
Congrats to the milestone of again running CI tests @vuilleumierc & @mki-c2c :rocket:

tests/core/test_processor.py
```diff
@@ -260,18 +275,20 @@ def test_processor_get_legend_entries():
...
     request.matchdict.update(request_matchdict)
     request.params.update(request_params)
     processor = create_processor()
     webservice = PlrWebservice(request)
     params = webservice.__validate_extract_params__()
-    real_estate = processor.real_estate_reader.read(params, egrid=u'TEST3')
+    real_estate = processor.real_estate_reader.read(params, egrid=u'TEST2')
     extract = processor.process(real_estate[0], params, 'http://test.ch')
     plrs = extract.real_estate.public_law_restrictions
-    assert len(plrs) == 4
```
vuilleumierc, line 273: @mki-c2c I think the goal of this test was to test the sorting mechanism according to the law statuses, but in the current test data we only have `InKraft` at the moment so this test is useless. I can prepare new test data

mki-c2c: Thanks, that's a good remark. So far, there is a lot of interdependent test data in the DB. Using local patches, we can customize the data for specific tests.

voisardf:
Hi, on Windows I can't get any tests running for now. I always get
`make: .venv/bin/py.test: Command not found`

I don't know it it's a windows specific issue due to the .venv setup? I try to clarify that with my colleagues

vvmruder:
Yes I think this is a windows specific thing. You could try the following:
```
docker-compose build
docker run --rm -v $(pwd):/workspace -u $(id -u):$(id -g) pyramid_oereb_oereb-server:latest make clean-all build
```

then you need to change the line 7 of the test configuration file tests/resources/test_config.yml from
`postgresql://postgres:postgres@localhost:5432/oereb_test_db`
to
`postgresql://postgres:postgres@oereb-db:5432/oereb_test_db`

save this file.

Then running `docker-compose exec oereb-server make test-core` should work and you should get the test results.

Keep in mind to not commit the changes on the test config YAML.

voisardf:
@vvmruder  Thanks. And yes, that works (299 test ok, none failed) - when testing is executed in the (unix) docker container👍

Which leads me to the question if we just leave it at that, Meaning we should adapt the documentation and specify (on L11-L12) that windows user need to run it in the container with `docker-compose exec ...`
Or if on the long run we enable the testing also in the windows venv, meaning that we need to add the switch in the Makefile for VENV_BIN too and allow the win-specific .venv/Scripts/ setting.

vvmruder:
I think we should add it to the readme because windows has its specifics and I remember it was not that straight forward to get it running.

vvmruder:
Other suggestion might be a ENV variable to override the behaviour. @mki-c2c @vuilleumierc what do you think about that?

vuilleumierc:
> Other suggestion might be a ENV variable to override the behaviour. @mki-c2c @vuilleumierc what do you think about that?

Yes that's fine with me, and if you want the default config could be to run the tests in docker.

vvmruder:
I think this should not block your merge. We can add this later to clarify.

voisardf:
I agree, no need to block this PR for this question! 👍

vvmruder (approved):
Thank you again for the work!

voisardf (approved):
Branch works for me locally and tests run through. 👍
As discussed, further improvements of test coverage and check if all tests still make sense should be treated as new issues once this PR is in the master

PR #1429, opened 2021-12-09
Fix reader tests

vuilleumierc:
Issue #1421


vvmruder (commented):
Thank you @vuilleumierc for the adaption. I have a point on that. Since the readers are not database related (the source is the link to the underlying data) I think we can git rid of database handling around the readers completely. To simulate the Source we could mock one and return a list of records for every reader. This is what all readers expect. going this way the tests for the readers would be fairly small and completely independent from a database. What do you think about it?

vuilleumierc:
> Thank you @vuilleumierc for the adaption. I have a point on that. Since the readers are not database related (the source is the link to the underlying data) I think we can git rid of database handling around the readers completely. To simulate the Source we could mock one and return a list of records for every reader. This is what all readers expect. going this way the tests for the readers would be fairly small and completely independent from a database. What do you think about it?

You mean like I did for `themes`, `law_status`, `document_types` and `logos` for the `test_extract` test?
Of course that should be possible, although it implies some extra work now that the tests have already been reimplemented. I'll look into it

jwkaltz:
> To simulate the Source we could mock one and return a list of records for every reader.

Thanks for the idea, @vvmruder , I think we could enter that as a potential follow-up issue?
As we have already reimplemented some of the tests and have limited budget for the test writing, I think we must finalize a "minimum viable product" for the tests first, and then re-prioritize again.

vvmruder:
Yes. We can leave it like it is. For now. I think it should be the way we look at the new core/contrib separation. Because one of the goals was the separation for more precise tests. Maybe we can take care of this in the test implementations we still have to do in the core part? Its ok to me to leave it for now in the reader part. I can do some work on this when its merged.

vvmruder (approved):

mki-c2c (approved):

PR #1434, opened 2021-12-12
Simplify config initialization

vuilleumierc:


tests/conftest.py
```diff
@@ -128,7 +122,8 @@ def test_db_engine(base_engine, test_db_name, test_db_url, config_path):
 
 @pytest.fixture(scope='session')
 @pytest.mark.usefixtures('config_path')
-def pyramid_oereb_test_config(config_path, test_db_engine):
+def pyramid_oereb_test_config(config_path, dbsession):
+    del dbsession
     # Reload the standard test configuration and now initialize it
     Config._config = None  # needed to force reload due to internal assertion
```
mki-c2c, line 128: I think that now as the url is read directly from the file and the scope for this fixture is "session", it might be possible to remove the line with the Config._config = None hack. Or did you experience any problems with multiple initialisation on this fixture ?

mki-c2c (approved):
That's a very nice improvement; getting rid of the "base_config".
One comment to try with a non-reset Config.init()

vvmruder (approved):
Good approach. :+1:

tests/conftest.py
vuilleumierc: Good point, although I tried and I get the `AssertionError` again (from when the Config is already initialized). I think I will leave it like this for now, and hopefully we'll find a better solution later on

PR #1443, opened 2021-12-14
Allow start without federal themes

### LINKED ISSUE #1442: entries in table theme_document are not mandatory
Entries in tables documents and theme_documents in the main schema are not mandatory although that will probably be the normal case. However, someone could start with cantonal themes only where all the documents are directly related to the PLR. This will cause an error.

I suggest to add an if statement to handle this case at https://github.com/openoereb/pyramid_oereb/blob/9d6d3138e4d70b1ffbed203cb1ca7120985bde43/pyramid_oereb/core/config.py#L133

vvmruder:
fixes #1442

### DIFF
--- a/pyramid_oereb/core/config.py
+++ b/pyramid_oereb/core/config.py
@@ -141,7 +141,7 @@ class Config(object):
                 if len(theme_documents) > 0:
                     theme.document_records = theme_documents
         else:
-            log.info('No global documents, related to themes were provided!')
+            log.info('No global documents related to themes were provided!')
 
     @staticmethod
     def _read_themes():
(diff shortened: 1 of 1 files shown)


pyramid_oereb/core/config.py
```diff
@@ -128,17 +128,20 @@ def init_offices():
...
+                    if theme_document.theme_id == theme.identifier:
+                        for document in Config.documents:
+                            if theme_document.document_id == document.identifier:
+                                document_copy = document.copy()
+                                document_copy.article_numbers = theme_document.article_numbers
+                                theme_documents.append(document_copy)
+                if len(theme_documents) > 0:
+                    theme.document_records = theme_documents
+        else:
+            log.info('No global documents, related to themes were provided!')
```
michmuel, line 144: The comma could possibly removed.

michmuel (approved):

pyramid_oereb/core/config.py
vvmruder: Solved it. Thx

PR #1450, opened 2021-12-16
Update README.rst

vvmruder:
Add some temp explanation about running the tests in docker.


voisardf (changes requested):
You might want to add at L12, that windows user should continue with the step 'locally but inside Docker'
thanks!

voisardf (approved):

PR #1479, opened 2022-01-24
1410 fix doc

mki-c2c:
Reactivate automatic doc generation on push @master.

Additionally, we could activate automatic doc generation on another well defined branch (eg. doc_test_branch) as had been done temporarily for test purposes here: https://github.com/openoereb/pyramid_oereb/commit/8adae1aba66f645682ec6e7aa8363bf4109aaabb#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddL84

The results of the doc build are visible on the github pages https://openoereb.github.io/pyramid_oereb/
(the current pages have been deployed by this workflow https://github.com/openoereb/pyramid_oereb/runs/4898582295?check_suite_focus=true)

### DIFF
--- a/.codacy.yaml
+++ b/.codacy.yaml
@@ -1,3 +1,8 @@
+engines:
+  bandit:
+    exclude_paths:
+      - "doc/source/conf.py"
+
 exclude_paths:
   - 'tests/**'
   - 'sample_data/**.json'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,6 +79,26 @@ jobs:
           path_to_write_report: ./coverage/codecov_report.txt
           verbose: true
 
+  doc:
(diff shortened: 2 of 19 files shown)


mki-c2c:
now Codacy in CI is unhappy with mako vulnerabilities (I'm not sure if we should take this into account)

jwkaltz:
> now Codacy in CI is unhappy with mako vulnerabilities (I'm not sure if we should take this partial CI failure into account)

These are not relevant, because they are only used for development, there is no mako templating in the resulting software library. So we should take them out of Codacy scope.

vvmruder:
I agree about removing Mako from codacy

jwkaltz (approved):

vvmruder (approved):

PR #1480, opened 2022-01-26
1391 test stats

mki-c2c:
reactivated stats output by adding c2cwsgiutils dependency
added test for a single entry point (logo) with stats test db


jwkaltz (approved):
Looks good, thanks!

jwkaltz (approved):

PR #1485, opened 2022-01-28
Fix Makefile

vvmruder:
Once more a sudden change on the XML's of the fed definitions gave some headaches...


jwkaltz (approved):

mki-c2c (commented):
would it be possible/useful to request an index page where the latest definition is referenced ?

PR #1493, opened 2022-02-08
Remove cryptography lib

jwkaltz:
Remove cryptography lib, not needed by pyramid_oereb


mki-c2c (approved):

vvmruder (approved):
Good catch-seems to be not used any more indeed.

PR #1494, opened 2022-02-08
use unique entry point for all tests both in CI and daily checks

mki-c2c:
daily checks broke because contrib tests were renamed

### DIFF
--- a/.coveragerc.contrib-stats
+++ b/.coveragerc.contrib-stats
@@ -1,0 +1,3 @@
+[run]
+source =
+    pyramid_oereb/contrib/stats/*.py
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -58,18 +58,10 @@ jobs:
       - run: sudo rm /etc/apt/sources.list.d/*.list
       - run: sudo apt update
       - run: sudo apt-get install virtualenv libpq-dev libgeos-dev
-      - name: Run core tests for Python ${{ matrix.python-version }}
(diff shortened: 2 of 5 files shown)


vvmruder (approved):

vvmruder:
In this context it might be useful to also separate tests of contrib stats (https://github.com/openoereb/pyramid_oereb/tree/master/pyramid_oereb/contrib/stats) into a separate part to support the new project/test structure?

mki-c2c:
> In this context it might be useful to also separate tests of contrib stats (https://github.com/openoereb/pyramid_oereb/tree/master/pyramid_oereb/contrib/stats) into a separate part to support the new project/test structure?

Currently the stats / logging tests are done in core: https://github.com/openoereb/pyramid_oereb/blob/fix_daily_check/tests/core/test_logger.py.  Should they be part of contrib ?

vvmruder:
I think we should provide it like we provide teh tests for other contrib stuff like here: https://github.com/openoereb/pyramid_oereb/blob/fix_daily_check/Makefile#L293-L294

In for the stats this would mean to have a target like: `test-contrib-stats` and the test files situated in the the already prepared folder: https://github.com/openoereb/pyramid_oereb/tree/fix_daily_check/tests/contrib.stats

vvmruder:
but we can solve this in an additional PR

vvmruder (approved):
:star_struck: I like it-now it really gets shape Thank you!!!

PR #1498, opened 2022-02-08
Reduce dependency

vvmruder:
add pillow as test dep only


jwkaltz (approved):
Thanks, in fact I was wondering about that.

PR #1499, opened 2022-02-08
further improve tests of plr source

### LINKED ISSUE #1497: should test some more functions of the plr module:
should test some more functions of the plr module:
- DatabaseSource.read
- DatabaseSource.from_db_to_xxx
- handle_collection
- bbox
but this can also be done later on in another PR

_Originally posted by @mki-c2c in https://github.com/openoereb/pyramid_oereb/pull/1456#discussion_r801505338_

vvmruder:
fixes #1497

### DIFF
--- a/pyramid_oereb/contrib/data_sources/standard/sources/plr.py
+++ b/pyramid_oereb/contrib/data_sources/standard/sources/plr.py
@@ -293,6 +293,18 @@ class DatabaseSource(BaseDatabaseSource, PlrBaseSource):
         return geometry_records
 
     def from_db_to_geometry_records(self, geometries_from_db):
+        """
+        Translates a geometry element read from database (SQLAlchemy) into the internal record format.
+
+        Args:
+            geometries_from_db
+                (list of pyramid_oereb.contrib.data_sources.standard.models.get_geometry.<locals>.Geometry):
+                    The element read out of the database.
+
+        Returns:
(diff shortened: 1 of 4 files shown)


marionb (changes requested):
@vvmruder thanks for the work. The tests look good to me from code point of view. When I run them every seccond time I have a lot of failed tests. So the first time all pass. Then on a seccond run I get a lot of failed tests: 

I tested with both the make target and directly with the docker-compose command. Same result.
I attached the summary of the case when it fails.
[pyramid_OEREB_test_summary.txt](https://github.com/openoereb/pyramid_oereb/files/9597697/pyramid_OEREB_test_summary.txt)

mki-c2c:
hello @marionb 
I suppose your problem concerning tests is related to this issue:
https://github.com/openoereb/pyramid_oereb/issues/1570

What I observed is the following:
when the tests fail, the make file does not reach the docker-compose down statement and the docker-compo is left running, so at the next test run the DB is ready.

However if the tests succeed, the docker-compo is stopped (https://github.com/openoereb/pyramid_oereb/blob/master/Makefile#L332)

And the next time the start up timing is tricky, the tests do not wait for the DB container to come up

Therefore you get the behaviour that the tests fail every second time but this is not related to the PR.

There is no problem in CI since the DB is an external component.

michmuel (approved):

marionb (approved):
@mki-c2c 
yes that is the behavior. 
OK good if the issue is already known.

PR #1500, opened 2022-02-09
use db host from ENV in test.ini

vvmruder:
@mki-c2c Please have a look. Since we have the use case running tests locally inside docker-compose I implemented env variable parsing to set db-host regarding to the env local docker or CI.

This works now for the `test.ini`

Still there is the `test_config.yml` which lacks of this possibility in the moment. You worked on this tests, so I ask you to look for a convenient way to have the db host also parsed from env in here: https://github.com/openoereb/pyramid_oereb/blob/master/tests/conftest.py


mki-c2c:
this modification seems to break local development tests. In which way do you launch the tests locally ?

vvmruder:
My and others usecase is to start the docker composition `docker-compose up` and afterwards run `docker-compose exec oereb-server make tests`/`docker-compose exec oereb-server make lint` etc.

This way its completely encapsulated in docker and no other deps are needed. For me this is convenient because using make locally conflicts with system lib versions like proj and I simply cant use it. As far as I know at least @voisardf and @michmuel are using this way too.

But if we use it like this we always need to change the two test config files because the DB-Host inside docker-compose is not `localhost` but `oereb-db`

mki-c2c:
I have found the library pyaml-env https://dev.to/mkaranasou/python-yaml-configuration-with-environment-variables-parsing-2ha6
It seems to be very suitable for what we want to do, parsing with environment values.
I can write a quick modification to config.py to use this lib.

mki-c2c:
@vvmruder I just talked to Wolfgang. He reminded me that the c2c tools already include a templating engine. c2ctemplates supports environment variables in the yaml configuration file. This is being used in the schwyz project: https://github.com/camptocamp/schwyz_oereb/blob/V2/config.yaml#L110
I'll check if it also works locally and so we can easily update the test files.

vvmruder:
> @vvmruder I just talked to Wolfgang. He reminded me that the c2c tools already include a templating engine. c2ctemplates supports environment variables in the yaml configuration file. This is being used in the schwyz project: https://github.com/camptocamp/schwyz_oereb/blob/V2/config.yaml#L110 I'll check if it also works locally and so we can easily update the test files.

Thats ok to me.

kdeininger:
I think, pyaml-env would have one advantage:

c2ctemplate, which is currently used to render the configuration files, is a template rendering tool using mako or jinja2, that can also access environment variables and provides a lot more features. As it is a separate tool, it needs to be run before the application is started. So we need this additional step, e.g. on container start-up, to respect the currently set environment variables. pyaml-env applies the values of the environment variables on parsing the configuration file. This way, no additional script would be necessary.

tests/core/test_logger.py
```diff
@@ -14,8 +16,8 @@
 @pytest.fixture(scope="session")
 def webtestapp():
     Config._config = None
-    setup_logging('tests/resources/test.ini#main')
-    app = get_app('tests/resources/test.ini#main')
+    setup_logging('tests/resources/test.ini#main', global_conf=os.environ)
+    app = get_app('tests/resources/test.ini#main', options=os.environ)
```
mki-c2c, line 20: hello,
I have huge trouble running tests in a local environment. If there are escaped environment variables (%s) this command fails.
I am testing some stuff, but I would recommend extracting the required values from os.environ like
```
app = get_app('tests/resources/test.ini#main', options={'PGHOST':os.environ.get('PGHOST', 'oereb-db')
```

.github/workflows/ci.yaml
```diff
@@ -38,6 +38,7 @@ jobs:
           - 5432:5432
         env:
           POSTGRES_PASSWORD: postgres
+          PGHOST: localhost
```
mki-c2c, line 41: I would invert the logics:
set PGHOST to localhost by default in the Makefile, and override this by setting oereb-db in docker-compose.yml

mki-c2c (changes requested):
We should continue to work on this.
The setup is not so robust for execution in different contexts.
Either the docker composition has to be very reliable or we must enable running the tests in a local environment.

vvmruder:
> We should continue to work on this. The setup is not so robust for execution in different contexts. Either the docker composition has to be very reliable or we must enable running the tests in a local environment.

What is unstable there? I don't experience any issues running it inside the composition?

tests/core/test_logger.py
mki-c2c: I have a local environment variable LESSOPEN='| /usr/bin/lesspipe %s'
this variable being present in the environment messes up the get_app command

mki-c2c:
> > We should continue to work on this. The setup is not so robust for execution in different contexts. Either the docker composition has to be very reliable or we must enable running the tests in a local environment.
> 
> What is unstable there? I don't experience any issues running it inside the composition?

for example the CI is not running in the docker composition, so I find it interesting that the config cannot be messed up with local environment variables.

I extracted a problematic code sample for easier analysis below.

to reproduce my problem, on can open an interactive python session on the docker instance with an environment variable with ```
docker-compose exec -e DUMMY_VAR='bla %s' oereb-server python
```
try to create the app
```
from pyramid.paster import get_app, setup_logging
import os
app = get_app('tests/resources/test.ini#main', options=os.environ)
```
and you get the following error:
`configparser.InterpolationSyntaxError: Error in file /workspace/tests/resources/test.ini: '%' must be followed by '%' or '(', found: '%s'`

**Comment by mki-c2c** [comment]
hello @vvmruder , do you mind if I do a rebase on this branch ?
Otherwise I'll just merge master, but I fear, it might not be very pretty

**Comment by vvmruder** [comment]
Sure, go on and rebase it

**Comment by mki-c2c** [comment]
improved version of makefile with env vars is ready, hope it still works for all of you

**Review by mki-c2c — APPROVED** [review]

**Comment by mki-c2c** [comment]
decrease in CI coverage is only due to deleted lines which were covered previously

File: Makefile [hunk 3]
```diff
@@ -282,35 +296,46 @@ test-postgis:
...
 .PHONY: test-contrib-stats
 test-contrib-stats: .venv/requirements-timestamp
 	$(VENV_BIN)/py.test -vv $(PYTEST_OPTS) --cov-config .coveragerc.contrib-stats --cov $(PACKAGE)/contrib/stats --cov-report=xml:coverage.contrib-stats.xml tests/contrib.stats
 
 .PHONY: tests
-tests: .venv/requirements-timestamp test-core test-contrib-data_sources-standard test-contrib-print_proxy-mapfish_print test-contrib-data_sources-standard test-contrib-stats
+tests: ${VENV_ROOT}/requirements-timestamp test-core test-contrib-data_sources-standard test-contrib-print_proxy-mapfish_print test-contrib-data_sources-standard test-contrib-stats
+
+.PHONY: docker-tests
+docker-tests:
```
vvmruder, line 319: Minor: About this integrated targets it's ok to have them but this means you need to have Make installed on the machine you issue them which is a bit cat bites own tail. These people have to run the docker-compose commands by hand then as they do now.

vvmruder (commented):
Since I can't approve my own PR I approve it with this comment :laughing:

vvmruder:
Great work @mki-c2c now it is like I would expect a dev setup :)

@michmuel @voisardf @svamaa => please play around with it. Hope you are satisfied too

Issue #1474, opened 2022-01-20
mapfish_print.py 


michmuel:
mapfish_print.py stops with error:

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/.venv/lib/python3.7/site-packages/waitress/channel.py", line 397, in service
    task.service()
  File "/app/.venv/lib/python3.7/site-packages/waitress/task.py", line 168, in service
    self.execute()
  File "/app/.venv/lib/python3.7/site-packages/waitress/task.py", line 434, in execute
    app_iter = self.channel.server.application(environ, start_response)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/router.py", line 270, in __call__
    response = self.execution_policy(environ, self)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/router.py", line 276, in default_execution_policy
    return router.invoke_request(request)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/router.py", line 245, in invoke_request
    response = handle_request(request)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/tweens.py", line 43, in excview_tween
    response = _error_handler(request, exc)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/tweens.py", line 17, in _error_handler
    reraise(*exc_info)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/util.py", line 733, in reraise
    raise value
  File "/app/.venv/lib/python3.7/site-packages/pyramid/tweens.py", line 41, in excview_tween
    response = handler(request)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/router.py", line 144, in handle_request
    registry, request, context, context_iface, view_name
  File "/app/.venv/lib/python3.7/site-packages/pyramid/view.py", line 674, in _call_view
    response = view_callable(context, request)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/config/views.py", line 170, in attr_view
    return view(context, request)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/config/views.py", line 196, in predicate_wrapper
    return view(context, request)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/viewderivers.py", line 392, in viewresult_to_response
    result = view(context, request)
  File "/app/.venv/lib/python3.7/site-packages/pyramid/viewderivers.py", line 113, in _class_requestonly_view
    response = getattr(inst, attr)()
[...]

PR #1501, opened 2022-02-09
change sort_legal_provision for officialNumber

michmuel:
closes https://github.com/openoereb/pyramid_oereb/issues/1474

It is possible that legal provision contains multiple entries with the same title but different official numbers (options for official number: string, empty string, or official number not in dict). Sorting of legal provisions should be able to handle this case.

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
@@ -685,8 +685,13 @@ class Renderer(JsonRenderer):
         """
 
         sort_title = elem['Title'] if 'Title' in elem else ''
-        sort_number = elem['OfficialNumber'] if 'OfficialNumber' in elem else None
-        return sort_title, sort_number
+        if 'OfficialNumber' in elem and elem['OfficialNumber']:
+            sort_number_id = 0
+            sort_number = elem['OfficialNumber']
+        else:
+            sort_number_id = 1
+            sort_number = ''
+        return sort_title, sort_number_id, sort_number
 
     @staticmethod
     def sort_hints_laws(elem):
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
@@ -676,7 +676,9 @@ class Renderer(JsonRenderer):
         """
         Provides the sort key for the supplied legal provision element as a tuple consisting of:
          * title
-         * Official number
+         * official number type ('0': if official number is defined and not an empty string;
+                                 '1': if official number is empty string or not defined)
+         * official number
 
         Args:
             elem (dict): one element of the legal provision list
@@ -686,12 +688,12 @@ class Renderer(JsonRenderer):
 
         sort_title = elem['Title'] if 'Title' in elem else ''
         if 'OfficialNumber' in elem and elem['OfficialNumber']:
-            sort_number_id = 0
+            sort_number_type = 0
             sort_number = elem['OfficialNumber']
         else:
-            sort_number_id = 1
+            sort_number_type = 1
             sort_number = ''
-        return sort_title, sort_number_id, sort_number
+        return sort_title, sort_number_type, sort_number
 
     @staticmethod
     def sort_hints_laws(elem):
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
@@ -686,14 +686,12 @@ class Renderer(JsonRenderer):
             sort key (tuple)
         """
 
-        sort_title = elem['Title'] if 'Title' in elem else ''
-        if 'OfficialNumber' in elem and elem['OfficialNumber']:
-            sort_number_type = 0
-            sort_number = elem['OfficialNumber']
-        else:
-            sort_number_type = 1
-            sort_number = ''
-        return sort_title, sort_number_type, sort_number
+        sort_title = elem.get('Title', '')
+        sort_number = elem.get('OfficialNumber', '')
+        # the additional sorting key below must be produced so that records without OfficialNumber
+        # are sorted after records with OfficialNumber
+        sort_empty_official_number_last = (len(sort_number) == 0)
+        return sort_title, sort_empty_official_number_last, sort_number
 
     @staticmethod
     def sort_hints_laws(elem):
--- a/tests/contrib.print_proxy.mapfish_print/test_mapfish_print.py
+++ b/tests/contrib.print_proxy.mapfish_print/test_mapfish_print.py
@@ -266,6 +266,37 @@ def test_get_sorted_legal_provisions(DummyRenderInfo):
             "DocumentType": "LegalProvision",
             "Lawstatus_Code": "inKraft",
             "Lawstatus_Text": "Rechtskräftig",
+            "OfficialNumber": "3891.100",
+            "ResponsibleOffice_Name": "Bundesamt für Verkehr BAV",
+            "ResponsibleOffice_OfficeAtWeb":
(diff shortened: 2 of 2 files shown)


pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -685,8 +687,13 @@ def sort_legal_provision(elem):
...
         sort_title = elem['Title'] if 'Title' in elem else ''
-        sort_number = elem['OfficialNumber'] if 'OfficialNumber' in elem else None
-        return sort_title, sort_number
+        if 'OfficialNumber' in elem and elem['OfficialNumber']:
+            sort_number_type = 0
+            sort_number = elem['OfficialNumber']
+        else:
+            sort_number_type = 1
+            sort_number = ''
+        return sort_title, sort_number_type, sort_number
```
mki-c2c, line 696: what do you think about:
```
        sort_title = elem.get('Title', '')
        sort_number = elem.get('OfficialNumber', '')
        # the additional sorting key below must be produced so that records without OfficialNumber
        # are sorted after records with OfficialNumber
        sort_empty_official_number_last = ( len(sort_number) == 0 )
        return sort_title, sort_empty_official_number_last, sort_number
```

mki-c2c (approved):
thanks for this good find and the tests.
I just had an idea to use default values in dict get methods for more compact and more readable code, but that's optional.

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
michmuel: Good suggestion. Changes already committed.

tests/contrib.print_proxy.mapfish_print/test_mapfish_print.py
```diff
@@ -261,6 +261,49 @@ def test_get_sorted_legal_provisions(DummyRenderInfo):
                 "http://www.bav.admin.ch/themen/verkehrspolitik/00709/index.html",
             "TextAtWeb": [{"URL": "https://oereb-gr-preview.000.ch/api/attachments/198"}],
             "Title": "Baugesetz"
+        }, {
+            "Canton": "BL",
+            "DocumentType": "LegalProvision",
+            "Lawstatus_Code": "inKraft",
+            "Lawstatus_Text": "Rechtskräftig",
+            "OfficialNumber": "3891.100",
```
mki-c2c, line 269: @michmuel 
can there be 2 records with same title but different OfficialNumber ?
In that case you could add one more record into this test and check that OfficialNumbers are sorted correctly (as a third rank sort criteria after title and number-not-empty)

michmuel: @mki-c2c In BL we have legal provisions with same title from which only one has a number. I think there are no restrictions concerning the tittle-number combinations in the official OEREB docs.

I split the tests in Baugesetz2 and Baugesetz3 as the sort function is not sensitive to official number not defined and official number=empty string.

peterschaer: Why is the sorting done on Title and OfficialNumber? Should we not use the index attribute here to sort the legal_provisions (and the laws and the hints)? In the federal spec for the pdf there is a certain order required (federal laws => federal ordinances => cantonal laws => cantonal ordinances). This order cannot be produced by sorting on title and number. In the JSON extract the order is as required.

vvmruder: I agree on that. Sorting by index would be the most clean and preferable solution. But @jwkaltz should have the best insight what is requested by the members so far.

@michmuel Did you checked if you can use sort by index in your case to get the output you want?

michmuel: I can create a branch with that capability and test it in our environment.

michmuel: @jwkaltz Where do you think the dcuments are sorted? I did not found it in https://github.com/openoereb/pyramid_oereb/blob/b8620f0372c6951b150c5bb999f17d7b3ce193ca/pyramid_oereb/core/renderer/extract/json_.py

I guess sorting in mapfish_print.py is fine as "Index" is mandatory and is the parameter for this purpose.

jwkaltz: If we sort them already when preparing the data extract, it would need to be at the place where we gather the documents, so that it's already sorted when serialized into JSON or XML.
Alternatively, each consumer needs to sort them, thus mapfish_print.py, and all oereb client implementations.

vvmruder: I completely agree-we should sort the documents in the core already. Like we do for themes. In the moment this is not done or implemented in pyramid_oereb

michmuel: I am not sure about that as there is no specification that elements in the extracts have to be sorted. Only the information is mandatory that enables sorting.

vvmruder: hmm this is true too

vvmruder: This is, why the index is delivered in the extract json/xml

michmuel: I created a PR to test sorting by index: https://github.com/openoereb/pyramid_oereb/pull/1504. Some comments are added there.
A general thing concerning the tests (@mki-c2c, @vvmruder): the json at https://github.com/openoereb/pyramid_oereb/blob/bugfix-1474/tests/contrib.print_proxy.mapfish_print/resources/test_extract.json still is in the old schema version 1.0.

voisardf: sorry for my late answer, I was really busy

Sorting of documents is defined in appendix E, page 50 and following:
`/** Ordnungszahl für die Sortierung im Auszug **/
AuszugIndex : MANDATORY -1000 .. 1000;`

A lot of the positive range is already blocked by the way the confederation defines it's indexes for the legal bases, e.g.:
> 000 Raumplanung
010 Bundesgesetz über die Raumplanung (SR 700)
100 Strassen
110 Bundesgesetz über die Nationalstrassen (SR 725.11)
120 Nationalstrassenverordnung (SR 725.111)

leaving following ranges for legal bases on state or municipality level:

> 000 Raumplanung
020 bis und mit 099 für kantonale und kommunale gesetzliche Grundlagen der Kategorie Raumplanung
100 Strassen
130 bis und mit 199 für kantonale und kommunale gesetzliche Grundlagen der Kategorie Strassen

In the Extract-model, the Index is also defined. But we can not use only the index to sort the documents, we need a second criteria - either the document type or alphabetical order on the title or something else.

Why? - It may happen that you have more legal bases on cantonal level than the available index range for a theme (but unlikely)
But we need to sort the legal dispositions too and there obviously exist more documents for some themes than the available ranges.
So, I'd suggest we use the document type & index as the sort criteria which allows us to use the complete range from -1000 to 1000 for other documents than legal bases. - I'm still unsure wheter 2000 positions are sufficient for large cantons in the Land Use Plans theme...

michmuel: In https://github.com/openoereb/pyramid_oereb/pull/1504 sorting is by index separately performed for 'LegalProvisions', 'Laws', 'Hints'.

It is possible that the index range is to small to be able to assign a number to each cantonal legal provision. However, the index does not have to be unique. Two documents that are never related to the same PLR may have the same index.

michmuel: Discussion in the pyramid_oereb weekly on 2022-02-14. Decision that sorting should be performed by index separately in the categories "legal_provisions" / "laws" / "hints". This PR is closed and the issue is continued in https://github.com/openoereb/pyramid_oereb/pull/1504.

PR #1503, opened 2022-02-10
clean up README

jwkaltz:


jwkaltz:
@vvmruder @voisardf @michmuel @svamaa 
the current state of the README is really unusable, so here a first proposal to fix it. With the help of @mki-c2c I was able to run the tests and updated that part. 
Regarding the part of the running the development server, I kept only the first section that describes this, assuming it is the most up-to-date. Is it correct? Or how to you start the development server when you use it?

vvmruder (approved):
:+1:

vvmruder (changes requested):
When I think about this seems to be related to [`docker-compose exec oereb-server make tests`](https://github.com/openoereb/pyramid_oereb/pull/1500) and I would vote for a in-docker way only. So we could skip the splitting docs about unix/windows and suggest the pure docker/docker-compose way. Since also on unix there has to be conditions met which might not exist on any system like on mine (wrong version of proj and other spatial libs).

kdeininger (approved):

jwkaltz:
> When I think about this again, it seems to be related to #1500 and I would vote for a in-docker way only. So we could skip the splitting docs about unix/windows and suggest the pure docker/docker-compose way. Since also on unix there has to be conditions met which might not exist on any system like on mine (wrong version of proj and other spatial libs).

@vvmruder for which section do you request changes? For the "Starting the development server", or for the "Running the tests"?

PR #1504, opened 2022-02-10
mapfish_print_sorting_by_index

michmuel:
Closes https://github.com/openoereb/pyramid_oereb/issues/1474

Sorting of documents by index separately in the categories "legal_provision" / "laws" / "hints".


pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -689,17 +689,21 @@ def sort_legal_provision(elem):
...
          * index
 
         Args:
-            elem (dict): one element of the hints list
+            elem (dict): one element of the hints / laws / legal provisions list
         Returns:
             sort key (tuple)
         """
-        return elem.get('index', 1000)
+        index = elem.get('Index', 1000)
```
jwkaltz, line 703: Why 1000, and not something really high, so that documents without index are sorted at the end for sure?

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -689,17 +689,21 @@ def sort_legal_provision(elem):
...
+        if index is None:
```
jwkaltz, line 704: I think this "if" is redundant to above (there is a default value for the get, so index can't be None after that)

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
michmuel: Specs (https://models.geo.admin.ch/V_D/OeREB/OeREBKRM_V2_0.ili) define the index as range from -1000...1000.

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
michmuel: Geolink formatter might return a None value: https://github.com/openoereb/geolink_formatter/blob/fdad0e3573b4492c2182c6d539ba5df88185c79b/geolink_formatter/entity.py#L91 (Index defined but none). But you are right, that is not according to the specs.

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
jwkaltz: Ah OK, thanks.
Could you add that as a comment to the code? That way it will be completely clear why we use 1000.

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
jwkaltz: Ah, good point.
Can you please add that as well as a comment to the code? (I didn't think about the case where index was explicitly set to "None" somewhere in the program).

jwkaltz (approved):
LGTM (just some suggestions for code comments, to make the motivation of the implementation a bit clearer)

vvmruder (approved):
Nice and clean now :) Thank you @michmuel

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
voisardf: @jwkaltz : I guess that the confederation was not aware that 1000 would not suffice as they are lacking a pratical implementation...

voisardf (commented):
Code looks good. I have yet to try it locally. But if anybody else is faster, take it. :)

PR #1505, opened 2022-02-14
Fix legend entry symbol selection

### LINKED ISSUE #1476: fix hook method for symbol
We tracked it down with @voisardf to the hook method which gets the legend symbol. I will take care about this.

Conclusion-we have more than one problem:

1. fix hook method for symbol
2. sanitizing of lower/upper case in print proxy
3. possibly need to fix legend entry collection in print proxy. Legend entries might have the same type code but a different ID. So selection by type code is not enough any more.

_Originally posted by @vvmruder in https://github.com/openoereb/pyramid_oereb/issues/1463#issuecomment-1018354643_

vvmruder:
- fixes #1476
- fix regarding test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,4 +84,5 @@ DataExtract.json
 /coverage.contrib-data_sources-standard.xml
 /coverage.contrib-print_proxy-mapfish_print.xml
 /coverage.core.xml
+/coverage.contrib-stats.xml
 /coverage.xml
--- a/dev/config/pyramid_oereb.yml.mako
+++ b/dev/config/pyramid_oereb.yml.mako
@@ -1116,7 +1116,7 @@ pyramid_oereb:
           model_factory: pyramid_oereb.contrib.data_sources.interlis_2_3.models.theme.model_factory_integer_pk
           schema_name: contaminated_civil_aviation_sites
       hooks:
-        get_symbol: pyramid_oereb.contrib.data_sources.standard.hook_methods.get_symbol
+        get_symbol: pyramid_oereb.contrib.data_sources.interlis_2_3.hook_methods.get_symbol
         get_symbol_ref: pyramid_oereb.core.hook_methods.get_symbol_ref
       law_status_lookup:
         - data_code: inKraft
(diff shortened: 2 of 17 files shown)


michmuel:
I got an error at https://github.com/openoereb/pyramid_oereb/blob/d7d9e445a89db6835ba236983a8c47fbeae1ac30/pyramid_oereb/contrib/data_sources/standard/hook_methods.py#L51:

```
AttributeError: type object 'LegendEntry' has no attribute 'id'
```
For interlis `id` is `t_id`.

voisardf:
I tried this branch locally and I still get following error:
```
  File "/workspace/pyramid_oereb/core/views/webservice.py", line 809, in get_image
    body, mimetype = method(dict(self._request_.params), theme_config)
  File "/workspace/pyramid_oereb/contrib/data_sources/standard/hook_methods.py", line 51, in get_symbol
    model.id == identifier
AttributeError: type object 'LegendEntry' has no attribute 'id'
```

vvmruder:
Yes I will fix that soon ;)

voisardf:
ok, let me know and I'll test again :)

vvmruder:
@michmuel @voisardf I fixed it, I think. Please have a look.

You need to adapt your configuration yml where ever you use interlis 2.3 sources (probably all federal themes). The new config needs to be `get_symbol: pyramid_oereb.contrib.data_sources.interlis_2_3.hook_methods.get_symbol`

I added a hook method fitting the interlis 2.3 models

voisardf:
@vvmruder I confirm, changing the hook methods for federal data layers, the error is gone. And I also get the symbols in the PDF.

One issue though is still left: For the two different restrictions of the same type but different geometry (a point and a polygon) I only get one legend entry for now (see attached pdf page 6). I guess that we have not fixed the sorting and counting somewhere (print proxy maybe?) to handle this properly. 

[20220216123905_extract.pdf](https://github.com/openoereb/pyramid_oereb/files/8080023/20220216123905_extract.pdf)

voisardf (approved):
Tested locally and works for me without errors.

vvmruder:
@voisardf could you provide corresponding JSON and XML as well?

vvmruder:
Ok. I agree here but want to wait for feedback from @michmuel

michmuel (approved):
Run in test environment. Mechanism to retrieve the legend symbols seems to work properly.

PR #1507, opened 2022-02-16
Remove Python 3.6 support

jwkaltz:
We must remove Python 3.6 support, as the end-of-life date for Python 3.6 has passed, and there are no longer any security updates for Python 3.6.

This also explains why some libraries that we use no longer support Python 3.6 in their current version.


voisardf (approved):
ok for me

PR #1511, opened 2022-02-17
Add multiple disclaimers

marionb:
To test the PDF use with the PR https://github.com/openoereb/pyramid_oereb_mfp/pull/88

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
@@ -497,16 +497,14 @@ class Renderer(JsonRenderer):
         extract_dict['RealEstate_RestrictionOnLandownership'] = restrictions
         # End one restriction entry per theme
 
-        for i, item in enumerate(extract_dict.get('Disclaimer', [])):
+        # Flatten the disclaimer dictionary
+        for item in extract_dict.get('Disclaimer', []):
             self._multilingual_text(item, 'Title')
             self._multilingual_text(item, 'Content')
 
-            if i == 0:
-                extract_dict['DisclaimerLandRegister'] = item
-                self._flatten_object(extract_dict, 'DisclaimerLandRegister')
-            if i == 1:
-                extract_dict['DisclaimerPollutedSites'] = item
-                self._flatten_object(extract_dict, 'DisclaimerPollutedSites')
+        # Separate the first item, so it can be placed in a different column in the template
+        extract_dict['DisclaimerLandRegister'] = extract_dict.get('Disclaimer', []).pop(0)
+        self._flatten_object(extract_dict, 'DisclaimerLandRegister')
 
         extract_dict['features'] = {
             'features': {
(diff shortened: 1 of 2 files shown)


pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -497,16 +497,17 @@ def convert_to_printable_extract(self, extract_dict, feature_geometry):
...
 
-            if i == 0:
-                extract_dict['DisclaimerLandRegister'] = item
-                self._flatten_object(extract_dict, 'DisclaimerLandRegister')
-            if i == 1:
-                extract_dict['DisclaimerPollutedSites'] = item
-                self._flatten_object(extract_dict, 'DisclaimerPollutedSites')
+        # Separate the first item, so it can be placed in a different column in the template
+        disclaimer = extract_dict.get('Disclaimer', [])
+        if len(disclaimer) > 0:
```
mki-c2c, line 507: does it also work with 0 disclaimers ?
Wouldn't it be good to add a test with 0 disclaimers, if there is time for it ?

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -497,16 +497,17 @@ def convert_to_printable_extract(self, extract_dict, feature_geometry):
...
+            extract_dict['DisclaimerLandRegister'] = disclaimer.pop(0)
```
mki-c2c, line 508: what about the other disclaimers, shall they be dropped ?

>> edit: I saw in mfp repo that disclaimers become a datasource, so this is fine, thanks
https://github.com/openoereb/pyramid_oereb_mfp/pull/88/files#diff-3c2b1d00d32412910451c36351e2094e2430d12d3cf966a8059a8123e88011a0R163

mki-c2c (approved):
Additional tests with 0 disclaimers and multiple (>2) disclaimers might be useful 
(as you mentioned in the JIRA ticket)
But this is no blocking point for me, it's good to make some progress.

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
marionb: The print specs will be generated with 0 disclaimers but the MFP part would crash because it expects something in the templates.
From what I understand there should always be a min of 2 disclaimers. The reason why I added the check for `> 0` is because there seems to be an issue with this test: [test_mfp_servicea](https://github.com/openoereb/pyramid_oereb/blob/503d396835aca8c3a0e6f308a7fe9557050d11db/tests/contrib.print_proxy.mapfish_print/test_mapfish_print.py#L830-L834). I think this does not return any data.
(The reason it passed before is because it was done with a `for`.)

I agree that more tests will be useful. I'll add some :+1:

voisardf:
@marionb: Thanks, multiple disclaimer works. Please check padding before merging. The spacing seems not exactly be as it should.
[20220222104249_extract.pdf](https://github.com/openoereb/pyramid_oereb/files/8116025/20220222104249_extract.pdf)

I had to comment transparency in pdfextract.jrxml, I guess there is still a transparent image remaining somewhere. Maybe only in my data.

voisardf (approved):
Please check my comment. Otherwise fine for me

peterschaer (approved):

voisardf:
@marionb Thanks, fix looks good to me too:
[20220224095700_extract.pdf](https://github.com/openoereb/pyramid_oereb/files/8131500/20220224095700_extract.pdf)

PR #1514, opened 2022-02-18
update shapely to 1.8

### LINKED ISSUE #1111: Clarify requirements for invalid geometries
The requirements for processing by pyramid_oereb in case of invalid geometries should be clarified. This use case can happen when data was produced according to Interlis in a form which is valid in Interlis, but which is considered invalid by the OGC ("Ring Self-intersection").

In #1109 some tests were added to see the current behavior of pyramid_oereb when an invalid geometry is processed. When processing in Python 3.6 and Shapely 1.6, the processing still works when such a geometry is processed. However, when using Shapely 1.7 the same code generates an exception (see #1013). 

Therefore, the community needs to specify the requirements for processing by pyramid_oereb when an invalid geometry (invalid according to OGC):
* should the processing always fail explicitly in this case?
* should pyramid_oereb explicitly accept invalid geometries, and add specific code for that (needed for updating shapely)?

### LINKED ISSUE #1513: CI does not take into account python version
CI is launched for a matrix of python versions, but each run uses the same python version 3.8 since the variable PYTHON_TEST_VERSION has been removed in the V2 makefile

mki-c2c:
shapely version 1.8 fixes the problems experienced with 1.7
So there is no reason not to use version 1.8
fixes #1111 

the CI is being improved so that different versions of python are tested correctly
(python 3.11 is removed until it is released and compatible with the used libraries)
fixes #1513

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ filetype==1.0.7
 geoalchemy2==0.8.5 # rq.filter: <0.9
 pyramid==1.10.8 # rq.filter: <2
 pyramid-debugtoolbar==4.9
-shapely==1.7.1  # rq.filter: == 1.7.1
+shapely==1.8.1  # rq.filter: == 1.8.1
 SQLAlchemy==1.4.23 # rq.filter: >= 1.4
 urllib3[secure]==1.26.8
 waitress==2.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ filetype==1.0.7
 geoalchemy2==0.8.5 # rq.filter: <0.9
 pyramid==1.10.8 # rq.filter: <2
 pyramid-debugtoolbar==4.9
-shapely==1.6.4.post1  # rq.filter: ==1.6.4.post1
+shapely==1.7.1  # rq.filter: == 1.7.1
 SQLAlchemy==1.4.23 # rq.filter: >= 1.4
 urllib3[secure]==1.26.8
 waitress==2.0.0
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,7 +51,6 @@ jobs:
           - '3.8'
           - '3.9'
           - '3.10'
-          - '3.11.0-alpha.5'
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python v${{ matrix.python-version }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,11 +47,11 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - 3.7
-          - 3.8
-          - 3.9
-          - 3.10
-          - 3.11
+          - '3.7'
+          - '3.8'
+          - '3.9'
+          - '3.10'
+          - '3.11'
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python v${{ matrix.python-version }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
 
   test-py:
     name: Test Python
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-18.04
     services:
       # Label used to access the service container
       postgres:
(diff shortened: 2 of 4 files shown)


vvmruder:
Thankyou @mki-c2c for this scientific excursion of geometry validity. I still do not know what problem is solved with it. We all in the user group agree that handling invalid geometries is not the concern of pyramid_oereb. All thoughts in #1111 are suggesting ways to handle invalid geometries outside pyramid_oereb. And I do not see what is introduced here. If it is some on-on-the-fly make-valid thing I clearly reject this because this works only for example geometries and not for real life real estate boundaries or others which can contain arbitrary number of points and rings. pyramid_oereb HAS TO FAIL on invalid geometries. It HAS TO REJECT CALCULATION on such. We need to discuss this before we merge this.

mki-c2c:
there are 2 problems:
- there are geometries which are valid for Interlis, but invalid for shapely. These geometries must be handled and cannot be rejected. shapely 1.8 is able to handle those. There is no magic about repairing geometries. These invalid geometries have minor problems with an easy fix (geometries which touch in one point can be translated into a geometry with a hole or a MultiPolygon)
- we have an outdated library (shapely 1.6) which will be more and more difficult to maintain.

The PR closes these 2 long-running issues, but some validation testing is probably needed on real infrastructure.

vvmruder:
About the update of shapely: :+1:

requirements.txt
```diff
@@ -3,7 +3,7 @@ filetype==1.0.7
 geoalchemy2==0.8.5 # rq.filter: <0.9
 pyramid==1.10.8 # rq.filter: <2
 pyramid-debugtoolbar==4.9
-shapely==1.6.4.post1  # rq.filter: ==1.6.4.post1
+shapely==1.8.1  # rq.filter: == 1.8.1
```
jwkaltz, line 6: If we are no longer stuck with a specific version of shapely, you can simply remove the "rq.filter" part.

jwkaltz (approved):

jwkaltz:
> But we need to be sure about the new behavior of shapely.

@vvmruder My understanding is in fact that we do not change the behavior in this regard compared to V1, when V1 is running with shapely 1.6 and Python <= 3.7. Because of technological side-effects, behavior changes in V1 when using shapely 1.7 or when using shapely 1.6 with Python >= 3.8. This is not good.

Keeping the behavior that V1 has (when it is used with Python <= 3.7) is, from the user feedback I have received so far, what is expected by the user group.

vvmruder:
Good morning together. If really nothing changes between shapely 1.6.4.post1 (which was in use until now) and 1.8.1 (which we want use now) we should add 1.6.4.post1 to the lookups in #1111 to show that. In the moment we have only compared 1.6.4.post2 which already changed a lot to 1.6.4.post1. If then the output of the viewed functionality is the same in 1.6.4.post1 and 1.8.1 we have a conclusion. Until now we don't.

mki-c2c:
@vvmruder Actually, shapely 1.6 post1 has been tested successfully with the latest logics in this commit:
https://github.com/openoereb/pyramid_oereb/pull/1514/commits/3450f7304e1ea9e74eb9e7ef7f4f586371d56a72

requirements have not been modified from old version, only additional tests logics have been introduced.
https://github.com/openoereb/pyramid_oereb/blob/3450f7304e1ea9e74eb9e7ef7f4f586371d56a72/requirements.txt#L6

vvmruder:
So we do add it in #1111 tables lookup to sumarize the behavior?

vvmruder:
@mki-c2c do you add you conclusion also in the table of #1111 as mentioned above?

jwkaltz:
@vvmruder can we merge now?

vvmruder (approved):
@mki-c2c Thx again for your work. We have it clarified now and everybody involved is aware of the changes comming through that.
Sorry for the endurance run.

It can be merged.

michmuel:
At the moment, it seems that only two geometries in the federal data that are not valid (after stroking):

schema_name, table, t_id 
contaminated_military_sites, geometrie, 12225
contaminated_military_sites, geometrie, 12542

One is located in canton ZH, the other in TI.

PR #1515, opened 2022-02-21
Update c2ctemplate library

jwkaltz:
Update the library c2ctemplate. The explicit dependency is deprecated (it was for Python 2 compatibility).


vvmruder (approved):
Works well as far as I can tell :+1:

PR #1517, opened 2022-02-21
Finalize reactivation of stats functionality

### LINKED ISSUE #1516: Stats views are empty
Hi everybody.

We are using in Ticino waitress also in production.
The new codebase produces JSON w/ "extras": null in logs table.msg, maybe because in webservice.py, the statements like
`# response.extras = OerebStats(service='GetEGRID', output_format=output_format) `
are commented, etc.

Due to this, the stats views are always empty
immagine

because they are parsing the response.extras "service" field as you can see in DDL:

...
```
   FROM logs
  WHERE logs.logger::text = 'JSON'::text AND (((logs.msg::json -> 'response'::text) -> 'extras'::text) ->> 'service'::text) = 'GetExtractById'::text;
```
...

Is it possible to fix this @jwkaltz ?
Thank you in advance.

jwkaltz:
Fix #1516

### DIFF
--- a/pyramid_oereb/core/views/webservice.py
+++ b/pyramid_oereb/core/views/webservice.py
@@ -17,8 +17,7 @@ from pyramid_oereb.core.readers.address import AddressReader
 from pyramid_oereb.core.renderer import Base as Renderer
 from timeit import default_timer as timer
 
-# TODO reactivate after fix of issue #1293: all usages of OerebStats in this file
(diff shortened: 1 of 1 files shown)


mki-c2c (approved):
Tested locally with a minor update of configs. I created a new PR on this here https://github.com/openoereb/pyramid_oereb/pull/1522

PR #1527, opened 2022-03-02
fix docker-compose file and makefile for linux and windows platforms

mki-c2c:


svamaa (approved):
works fine for me now with the sample data.
Thank you for the changes!

PR #1529, opened 2022-03-07
fix legend entry collection

### LINKED ISSUE #1477: possibly need to fix legend entry collection in print proxy
possibly need to fix legend entry collection in print proxy. Legend entries might have the same type code but a different ID. So selection by type code is not enough any more.

_Originally posted by @vvmruder in https://github.com/openoereb/pyramid_oereb/issues/1463#issuecomment-1018354643_

marionb:
fixes #1477
- fix logic in code
- add new test case

There is no real test data for this just the tests. Can someone who has data with this case test this and check if it is working as expected with the print templates?

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
@@ -479,7 +479,7 @@ class Renderer(JsonRenderer):
         for restriction in restrictions:
             legends = {}
             for legend in restriction['Legend']:
-                type_ = legend['TypeCode']
+                type_ = f"{legend['TypeCode']}_{legend['Geom_Type']}"
                 if type_ in legends:
                     for item in ['AreaShare', 'LengthShare', 'PartInPercent', 'NrOfPoints']:
                         if item in legend:
--- a/tests/contrib.print_proxy.mapfish_print/resources/expected_getspec_extract.json
+++ b/tests/contrib.print_proxy.mapfish_print/resources/expected_getspec_extract.json
@@ -38,6 +38,10 @@
     {
       "Code": "ch.LaermempfindlichkeitsstufeninForce",
       "Text": "Lärmempfindlichkeitsstufen (in Nutzungszonen)"
+    },
+    {
+      "Code": "ch.LaermempfindlichkeitsstufeninForce",
+      "Text": "Lärmempfindlichkeitsstufen (in Nutzungszonen)"
     }
   ],
   "NotConcernedTheme": [
@@ -995,7 +999,6 @@
         }
       ],
       "Hints": [],
-      "Geom_Type": "AreaShare",
       "Legend": [
         {
           "TypeCode": "ES_III",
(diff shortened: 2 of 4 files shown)


tests/contrib.print_proxy.mapfish_print/resources/expected_getspec_extract.json
```diff
@@ -35,6 +35,10 @@
       "Code": "ch.LaermempfindlichkeitsstufeninForce",
       "Text": "Lärmempfindlichkeitsstufen (in Nutzungszonen)"
     },
+    {
+      "Code": "ch.LaermempfindlichkeitsstufeninForce",
+      "Text": "Lärmempfindlichkeitsstufen (in Nutzungszonen)"
```
mki-c2c, line 40: maybe there should be the geometry type also appended here, so that there are no 2 identical entries, since this can be rather confusing.

mki-c2c (approved):
I think the implemented behaviour is up to expectations, I just added a comment about the themes list. It is usually confusing when there are two identical entries, so maybe this should be discussed with the "power" users ?

vvmruder (approved):
Code looks ok. I would like to see some real world test with the PDF :)

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -479,7 +479,7 @@ def convert_to_printable_extract(self, extract_dict, feature_geometry):
         for restriction in restrictions:
             legends = {}
             for legend in restriction['Legend']:
-                type_ = legend['TypeCode']
+                type_ = f"{legend['TypeCode']}_{legend['Geom_Type']}"
```
michmuel, line 482: I think, there are no constraints defined on TypeCode from the "Rahmenmodell". Therefore, additional similar issues might occur in the future. Would it not make sense to switch here to the legend entry id?

vvmruder: This is correct from the data point of view. Better use the id. It should be available as `identifier` on the LegendEntryRecord

voisardf (approved):
works for me, thanks! I get two distinct entries with our data 👍 
I do not have the icons, but that is probably due to the transparency issue with federal legend icons. I'll check that next week with updated symbols

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
marionb: @vvmruder should we fix this?
In the JSON I do not see much that I could use as a legend id. Do I need to add it from the DB? OR did I just miss this?

michmuel: Isn't the "SymbolRef" unique for a legend entry?

marionb: @michmuel OK I just changed it to use `SymbolRef` (its a long key :-) ) 
@voisardf can you check if this is also working for you?

michmuel:
@voisardf Do you use https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/contrib/data_sources/interlis_2_3/hook_methods.py for the interlis themes?

voisardf:
@michmuel : yes, I use this models for federal data, but standard for cantonal themes

michmuel:
I supposed that you might not use the explicit hook method for interlis_2_3. Just a quick thought.

voisardf:
@michmuel @marionb yes, the config was the issue - i fixed the symbol hook method and now it's fine.
[20220328094033_extract.pdf](https://github.com/openoereb/pyramid_oereb/files/8361507/20220328094033_extract.pdf)

so I can confirm that with our data the point/area symbol display works with symbol_ref key 👍

PR #1534, opened 2022-03-21
Fix JSON response of GetEgrid

### LINKED ISSUE #1530: JSON response of GetEgrid is invalid
The attribute `type` needs to be encoded as `RealEstateType` containing a valid `Code` and a `MultilingualText`.

```json
...
"type": {
  "Code": "RealEstate",
  "Text": [
    {
      "Language": "de",
      "Text": "Liegenschaft"
    }
  ]
}
...
```

kdeininger:
Closes #1530

### DIFF
--- a/pyramid_oereb/core/views/webservice.py
+++ b/pyramid_oereb/core/views/webservice.py
@@ -156,6 +156,20 @@ class PlrWebservice(object):
                     'Invalid parameters. You need one of the following combinations: '
                     'EN or GNSS or IDENTDN and NUMBER or POSTALCODE, LOCALISATION and NUMBER.'
                 )
+            supported_languages = Config.get_language()
+            for record in records:
+                real_estate_type = Config.get_real_estate_type_by_data_code(record.type)
+                text = []
+                for lang in real_estate_type.title:
+                    if lang in supported_languages:
+                        text.append({
+                            'Language': lang,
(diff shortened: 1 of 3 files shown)


kdeininger:
As the service provides no parameter to specify the language, the real estate type is returned containing all available and configured languages (similar to capabilities).

vvmruder:
Did you checked this also for the XML?

jwkaltz:
@vvmruder @kdeininger what was the conclusion regarding the question about whether a fix is also needed for the XML?
Jura is reporting that the getegrid/xml is not working (but the getegrid/json is working)

kdeininger:
The getegrid/xml was working and valid before. I think the following line

https://github.com/openoereb/pyramid_oereb/blob/e72a9e96b8516166fe463e054937b726f1e314a1/pyramid_oereb/core/renderer/getegrid/templates/xml/getegrid.xml#L19

needs to be changed as `Config.get_real_estate_type_by_data_code()` has already been called here:

https://github.com/openoereb/pyramid_oereb/blob/e72a9e96b8516166fe463e054937b726f1e314a1/pyramid_oereb/core/views/webservice.py#L473

kdeininger:
@jwkaltz XML should be fixed with https://github.com/openoereb/pyramid_oereb/pull/1545.

PR #1538, opened 2022-03-25
Object has no Attribute

### LINKED ISSUE #1532: 'DatabaseSource' object has no attribute '_theme_record'
immagine

in statement 
        # Add empty record if topic is not available
        else:
            self.records = [EmptyPlrRecord(self._theme_record, has_data=False)]

https://github.com/openoereb/pyramid_oereb/blob/ce8b5cf3901b9324bb56a13ed49831e134c39278/pyramid_oereb/contrib/data_sources/interlis_2_3/sources/plr.py#L566

marionb:
fix #1532


michmuel (approved):
Looks good to me (without having tested it).

PR #1539, opened 2022-03-29
Fix create_standart_templates

marionb:
TODO

- [x] create model for all themes except if the config is set to 'standard: False'
- [ ] adapt the config.yaml to include the pram.
- [ ] test the functionality


vvmruder:
@marionb Since I was a little blocked by this I did a quick fix to create the standard tables based on a "decision-parameter" in my PR https://github.com/openoereb/pyramid_oereb/blob/bugfix/1508-improve-performance/pyramid_oereb/contrib/data_sources/create_tables.py#L24

Is this maybe helping you? This way all tables can be created by the configured source. Of course this would mean to add some more functions for oereb-lex, etc. I guess INTERLIS is created by ili2db or another tool, so creation via script is not necessary for this source type.

marionb:
This will be fixed by the PR #1542 and is no longer necessary!

PR #1540, opened 2022-03-30
Improve Performance by changing availability behaviour

### LINKED ISSUE #1508: Possibly poor performance compared to pyramid_oereb 1.8.0
In our tests we are observing a noticeable performance difference between our current productive environment (based on pyramid_oereb 1.8.0) and the new version based on pyramid_oereb 2.0.0b14. One may observe the difference here:

pyramid_oereb 1.8: https://www.oereb-test.apps.be.ch/extract/reduced/json/CH173510804618
pyramid_oereb 2.0: https://www.oereb2-test.apps.be.ch/extract/json?egrid=CH173510804618 (the DNS entry is not public yet, you can use https://159.144.217.124/extract/json?egrid=CH173510804618 instead but you have to ignore the security warnings about the certificate)

Both instances are running in the very same kubernetes cluster in our data center. They are connecting to the same PostgreSQL 12 database (but different schemas because of the changes made to the transferstruktur data model). The data is the same (the old transferstruktur was transformed into the new one). With pyramid_oereb 1.8 the JSON extract is created in under one second (700-800ms), with pyramid_oereb 2.0 the JSON extract is created in 3.5-4 seconds!

I did some researching and profiling (with debug-toolbar und print-Statements). I have the impression that the difference has to do with the database queries. E.g. just creating the processor 
https://github.com/openoereb/pyramid_oereb/blob/e291d8a5adccf6e7aa61b58d60f66f4d1e655d5c/pyramid_oereb/core/processor.py#L321 takes about 2.5 seconds. The slow part ist querying the availabilities https://github.com/openoereb/pyramid_oereb/blob/e291d8a5adccf6e7aa61b58d60f66f4d1e655d5c/pyramid_oereb/contrib/data_sources/standard/sources/plr.py#L121 But this is a very simple query returning 338 rows in our case. I don't see anything to optimize here. This query takes about 0.07s. But it is done 31 times = 2.1s. But this part of the code hasn't changed from pyramid_oereb 1.8 and 2.0. The old version is also querying the availabilities of all 31 topics. But it does it much quicker. I suppose there's some overhead on each query and with more than 200 queries for an extract this adds up. But I can't narrow this down. May be there is some change is one of the used libraries or it has to do with the python version (pyramid_oereb 1.8 is running on Python 3.6, pyramid_oereb 2.0 is running on Python 3.7.

I think I can exclude the network as the source of error. Because when running both the application and the database as docker containers on my development environment I get the same results. In this setup there is no network between application and database.

I'm using the standard model for all 31 topics we have.

I cannot explain this difference therefore I'm posting this here. Has anybody during their tests observed something similar?

vvmruder:
This PR changes the setup of availabilities by removing them from the theme schema and put them to the main schema as a setting. This way the Availabilities can be read easily one time on server start.

The table of availabilities in the main schema can be empty and this means all plrs are published by default.

fixes #1508

What needs to be changed when this PR is tested on cantonal infrastructure:

1. configuration has to be updated: https://github.com/openoereb/pyramid_oereb/blob/bugfix/1508-improve-performance/dev/config/pyramid_oereb.yml.mako#L574-L589
2. application schema has to be updated => new table availability
3. content of availability has to be moved out of the theme schema availability table into the new table => new structure: https://github.com/openoereb/pyramid_oereb/blob/bugfix/1508-improve-performance/pyramid_oereb/contrib/data_sources/standard/models/main.py#L329-L354 (if no availability constraints are taking place new table can stay empty = means everything is published)
4. server needs a restart


vvmruder:
@peterschaer could you please check if this change is solving you issue?

pyramid_oereb/contrib/data_sources/standard/models/main.py
```diff
@@ -324,3 +324,31 @@ class MapLayering(Base):
...
+        id (str): The identifier of this list. It is used in DB only.
+        municipality_fosnr (int): The municipality identifier which the switch element belongs to.
+        theme_code (str): the theme code which this switch element belongs to.
+        available (bool): Switch wether the element is published or not.
+    """
+    __table_args__ = {'schema': app_schema_name}
+    __tablename__ = 'availability'
+    id = Column(String, primary_key=True)
+    municipality_fosnr = Column(ForeignKey(Municipality.fosnr), nullable=False)
+    theme_code = Column(String, nullable=False)
```
michmuel, line 353: Is theme_code sufficient, or is also sub_theme required?

michmuel, line 353: While reading I had a minute to recognize what is meant. Theme, sub-theme, code, topic, sub-topic, ... Maybe plr_code would be clearer here (or not ...).

michmuel:
I will update our setting in order to be able to test it (~begin of next week).

pyramid_oereb/contrib/data_sources/standard/models/main.py
vvmruder: I This naming follows the pattern of column names: The table it is soft related to is the [theme](https://github.com/openoereb/pyramid_oereb/blob/bugfix/1508-improve-performance/pyramid_oereb/contrib/data_sources/standard/models/main.py#L144-L163) and in that table it is the [code colum](https://github.com/openoereb/pyramid_oereb/blob/bugfix/1508-improve-performance/pyramid_oereb/contrib/data_sources/standard/models/main.py#L159). It is a soft relation because the code is not unique but enough for our publish switch because we do not want to decide on sub theme base. So how it it decided in other cases too the name of the column is `theme_code` like `public_law_restriction_id`

peterschaer:
Thanks! I will test this tomorrow or early next week.

pyramid_oereb/core/config.py
```diff
@@ -172,6 +176,13 @@ def init_offices():
         except ProgrammingError:
             Config.offices = None
 
+    @staticmethod
+    def init_availabilities():
+        """
+        Initializes availabilities from the configured source.
+        """
+        Config.availabilities = Config._read_availabilities()
```
michmuel, line 184: Do we need here an except as in line 176?

```
except ProgrammingError:
            Config.availabilities = None
``

vvmruder:
Ok. We will merge this. And dig a bit deeper for analyzing the problems. #1508 will stay open.

PR #1543, opened 2022-04-04
Remove unused method

vvmruder:
This method was not used in interlis context because interlis only delivers non collection type geometries.


vvmruder:
@michmuel can you confirm this?

mki-c2c (approved):
as far as I can see, this unused function can be removed

PR #1544, opened 2022-04-06
Further improvement of Performance

vvmruder:
@peterschaer 
I did a quick profile of the extract generation (based on our small test data set). It reveals several points to poke around. But one easy catch is right in the beginning of the process:
Screenshot_20220406_114602

Is related to #1508 

As you can see we lose almost 4% of process time between start and ExtractReder.read, I found that we had an old left over in the code. Instead of reading glossaries, disclaimers and municipalities for every extract they are read once on server start now.

Could you please try and tell what you think about this?

To use this patch it should be enough to install it. No config changes needed.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,4 +6,5 @@ McCabe==0.6.1
 Sphinx<1.6
 sphinx_rtd_theme
 c2c.template==2.3.0
+yappi
 -e .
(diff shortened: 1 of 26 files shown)


vvmruder:
@mki-c2c strange that tests fail in CI. Locally all tests run through.

michmuel (commented):
I deployed the branch in our test setting. Seems to work smoothly. Thanks.

michmuel (approved):

PR #1545, opened 2022-04-07
Fix real estate type in XML for GetEgrid

kdeininger:
Render real estate type with code and multilingual text.


michmuel (approved):
Looks good (not tested).

kdeininger:
> What was wrong with using the real estate type from `real_estate_type.xml`?

`real_estate_type.xml` expects an object with the attributes `code` and `title` as input but the value for the GetEgrid XML is a dictionary containing the properties `Code` and `Text`. Furthermore, the `multilingual_text.xml` (included in `real_estate_type.xml`) only returns the localized text for the requested language but for GetEgrid all available languages should be returned as there is no language parameter for this service.

vvmruder (approved):
thx for the explanation :)

PR #1542, opened 2022-04-04
Fix creation of standart tables

marionb:
TODO:

- [x] Take apart table creation for OEREBlex and standard tables
- [x] Create the Main schema separateness from the theme tables

Currently I have noticed the following issues (might not be cause by the changes from this PR):

- When building the image new and removing all the containers there is an issue that the test db is not there. The PostgreSQL container continuously logs: `` The issue disappear after having run the tests once. (the application seems to be running)
- Running the tests twice:
    1. Run: the test pass
    2. Run: there is an issue because the containers have been stopped and can no longer be accessed by the tests
- Multiple `oereb/main_container` exist simultaneously

PR #1547, opened 2022-04-07
Separate table creation for interlis and standart themes

### LINKED ISSUE #1541: Generalize create_standart_tables
- [ ] take apart creation of standard tables for oereblex and standard
- [ ] do not create standard tables for interlis model

marionb:
Fixes #1541
This replaces the old PR #1542 

- make create_oereblex_tables a command line tool

- separate creation of main schema & tables

- fix and add tests

- update docu

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -100,7 +100,7 @@ script:
  create_standard_tables -c pyramid_oereb_standard.yml --sql-file=sqlFile.sql
  create_oereblex_tables -c pyramid_oereb_standard.yml --sql-file=sqlFile.sql
 
-Then load the generate sql file into your DB
+Then load the generated sql file into your DB
 
 .. warning:: We assume you did not specify a custom name for the configuration file. If you have used a custom
    name, you will have to replace `pyramid_oereb_standard.yml` with your specified name in this and all
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -96,7 +96,11 @@ script:
 
 .. code-block:: shell
 
- create_standard_tables -c pyramid_oereb_standard.yml
+ create_main_schema_tables -c pyramid_oereb_standard.yml --sql-file=sqlFile.sql
+ create_standard_tables -c pyramid_oereb_standard.yml --sql-file=sqlFile.sql
+ create_oereblex_tables -c pyramid_oereb_standard.yml --sql-file=sqlFile.sql
+
+Then load the generate sql file into your DB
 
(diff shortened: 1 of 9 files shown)


doc/source/installation.rst
```diff
@@ -96,7 +96,11 @@ script:
 
 .. code-block:: shell
 
- create_standard_tables -c pyramid_oereb_standard.yml
+ create_main_schema_tables -c pyramid_oereb_standard.yml --sql-file=sqlFile.sql
+ create_standard_tables -c pyramid_oereb_standard.yml --sql-file=sqlFile.sql
+ create_oereblex_tables -c pyramid_oereb_standard.yml --sql-file=sqlFile.sql
+
+Then load the generate sql file into your DB
```
jwkaltz, line 103: generate -> generated

vvmruder (approved):
I quickly fly over. Its the state we already discussed extensively on the old PR. So I approve here. :)

marionb:
@vvmruder I currently have an issue that the DB does not get filled any more when I am on this PR. Master seems to work though. I think I am breaking something here do I will try to investigate before I merge this.
I am not sure though I will be able to provide a fix before I leave for vacations next week until the 25.4

vvmruder:
Well. You can leave me an info here what is the state at the end of the day. I will have a look later then and see how I can make it work. Good holidays then. :) :palm_tree:

marionb:
@vvmruder 
I saw two points
1. the script that initializes the DB in the DB container under `docker-entrypoint-initdb.d/02_create_dev_tables.sql` contains the whole creation of all the schemas 3 times. This script is generated from the 3 `creation_*` scripst. Since I am appending to the file and not overwriting it it it seems the the `creation_*` function are executed 3 time. I am not sure why.
2. somehow the stats db is not creatd. On start up there is an error. Here I am not on the current where the DB is used and why it is not created.
I don't know perhaps there is an an other issue but I did not find it yet.

voisardf:
@marionb : what is the state of this PR? are you still working on that, or do you need us to test something?

marionb:
@voisardf I planend some time beginnen this week to take up the work again on this again.
the branch will probably need a rebase and then check what needs to be fixed with after.
After that I would be glad fore someone to test by the beginning of next week.

marionb:
@vvmruder 
I've just tested this locally and it looks to me as though it is now working here as well. Do you want to also test again or can I merge this?

vvmruder:
@marionb Thanks for rechecking this. I think we can merge

Issue #1528, opened 2022-03-02
CI shall check XML urls


mki-c2c:
we experienced changed in XML urls without notice.
CI shall detect these changes and fail.
http://models.geo.admin.ch/V_D/OeREB/

quick fixes were required like in https://github.com/openoereb/pyramid_oereb/pull/1526 and https://github.com/openoereb/pyramid_oereb/commit/306ad921fe66deddb8aadc4dc031f0fa2c0d3fc1

mki-c2c:
include XML files into CI

CI checks if URL has changed => investigate if automatic pull requests can be created ??

PR #1548, opened 2022-04-07
1528 freeze xml

mki-c2c:
According to issue https://github.com/openoereb/pyramid_oereb/issues/1528 the json data files are now included in the repo.
In this way, the local build (make serve) will still work even if the upstream links are broken.

I did some refactoring of the Makefile to have a better organisation of the fed data generation.

There are some additional make recipes for checks and update.

- `make check_fed_data`: will compare the fixed local data files with freshly generated ones and fail if a difference is found
- `make update_fed_data_urls`: will try to find the newest versions of the XML files for Gesetze, Themen, Logos, Texte on the server index. The file fed.urls should be double checked, then committed
- `make update_fed_data`: generate fresh data files and copy them into the project. The modified files can then be committed
- `make auto_update_fed_data`: generate new URLs and new files (if you trust the tool ;-) ) => rather double check the generated files

CI and daily build do check the validity of the files.


voisardf (approved):
I checked this branch out locally rebuild my instance and run it (connecting also our dev database). It worked fine for me after having dealt with some configuration issues 😉 and the new availability table in the main schema.
I did NOT go through all the code reading and checking it.

vvmruder (approved):
@mki-c2c I did a check locally and it works as expected. However I found the files in .fed_check to be not git-ignored. I added them to the gitignore. You can roll back if I miss something if this was done on purpose :)

PR #1549, opened 2022-04-19
Move DataIntegration to application schema

vvmruder:
DataIntegration is a Table which holds information about state of data provided by pyramid_oereb.

**What changed:**

- move table data_integration to main schema
- gather all dev data from theme schema to the new table
- fix related models
- provide reader, source, records etc

This PR does not offer the possibility to load data integration on start up. Since this data is not offered through any API defined by FED. In the moment everything would be ready to do so. But config is simply not read. Users of pyramid_oereb can use this table as a bucket to drive data updating (store hashes).

**Changes for config/data import:**
Since table was deleted in each theme schema, the user might import data directly into the application schema table. Be aware that there is a relation between office and data integration. So corresponding offices have to be imported before.


michmuel:
For me it is not so clear how and if data_integration should be implemented in the pyramid_oereb code as its application/specification is a bit vague. Filling in date and office is probably not clearly defined if data is accepted in the transfer format on a daily basis even if nothing has changed.

Anyway, I support removing the tables from the topics as they are not accessed.

vvmruder:
@michmuel So what is the conclusion? Do we merge?

michmuel:
We do not use/access tables "data_integration" in our oereb environment at the moment. So in my view it is a vague feature. However, the PR looks fine and if there is a canton that would use it, we should merge.

michmuel (approved):

Issue #1560, opened 2022-05-12
Document that is relevant only for one municipality


michmuel:
There is no filter implemented in pyramid_oereb for parameter "only_in_municipality" (field in table "Dokument" in "Oereb-Transferstruktur").

PR #1561, opened 2022-05-13
bugfix-1560-doc-only-for-one-municipality

michmuel:
closes https://github.com/openoereb/pyramid_oereb/issues/1560

- Additional filter added in processor.py
- Tests for processor slightly adapted

To my knowledge "only_in_municipality" is not an attribute in OerebLex (true?). So "only_in_municipality" is always None for documents provided by OerebLex.
https://github.com/openoereb/pyramid_oereb/blob/8ca834de3f4abd9cd7d8fdfd61a443c3dfd71190/pyramid_oereb/contrib/data_sources/oereblex/sources/document.py#L205

### DIFF
--- a/pyramid_oereb/core/processor.py
+++ b/pyramid_oereb/core/processor.py
@@ -76,9 +76,11 @@ class Processor(object):
 
     def plr_tolerance_check(self, extract):
         """
-        The function checking if the found plr results exceed the minimal surface or length
-        value defined in the configuration and should therefor be represented in the extract
-        or considered 'false trues' and be removed from the results.
+        The function checks if the found plr results exceed the minimal surface or length
+        value defined in the configuration and therefore should be represented in the extract
+        or if they should be considered as 'false trues' and be removed from the results.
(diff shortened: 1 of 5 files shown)


pyramid_oereb/core/processor.py
```diff
@@ -78,7 +98,10 @@ def plr_tolerance_check(self, extract):
                 if public_law_restriction.calculate(real_estate, Config.get('geometry_types')):
                     log.debug("plr_tolerance_check: keeping as potentially concerned plr {}".
                               format(public_law_restriction))
-                    inside_plrs.append(self.filter_published_documents(public_law_restriction))
+                    public_law_restriction = self.filter_documents_by_fosnr(public_law_restriction,
```
vvmruder, line 101: I would add a little part to method doc that this is done. Otherwise it looks good.

vvmruder (approved):

vvmruder (approved):
:+1:  for the docs :)

PR #1564, opened 2022-05-19
fed definitions have changed upstream which has been caught by CI

mki-c2c:
update has been done simply by running make auto_update_fed_data


jwkaltz (approved):
Thanks!

PR #1566, opened 2022-05-23
Update library versions used in tests

jwkaltz:


jwkaltz:
Merging with review, as this update concerns only the automated tests, and these are green.

PR #1567, opened 2022-05-23
Update runtime libraries

jwkaltz:
Update all runtime libraries,
except PyPDF2, see https://github.com/openoereb/pyramid_oereb/issues/1568


michmuel (approved):
Seems to work fine.

PR #1571, opened 2022-06-02
add optional tolerance on geometric operations

mki-c2c:
tolerance can be configured on plr base in yaml file as shown below:

if there is a tolerance, geometries are compared with buffer / distance
if there is no defined tolerance, the simple operations are used (Intersects / Intersection)
```
pyramid_oereb:
  plrs:
    - code: ch.Nutzungsplanung
      geometry_type: GEOMETRYCOLLECTION
      thresholds:
        length:
          limit: 1.0
          unit: 'm'
          precision: 2
        area:
          limit: 1.0
          unit: 'm²'
          precision: 2
        percentage:
          precision: 1
      tolerance: 0.00001
```

### DIFF
--- a/.coveragerc.contrib-data_sources-interlis
+++ b/.coveragerc.contrib-data_sources-interlis
@@ -1,0 +1,3 @@
+[run]
+source =
+    pyramid_oereb/contrib/data_sources/interlis_2_3/*.py
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -69,7 +69,7 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           env_vars: OS,PYTHON
           fail_ci_if_error: true
-          files: ./coverage.core.xml,./coverage.contrib-data_sources-standard.xml,./coverage.contrib-print_proxy-mapfish_print.xml,./coverage.contrib-stats.xml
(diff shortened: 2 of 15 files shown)


vvmruder:
@mki-c2c 
Would it make sense to offer this not only for the standard source but also for the rest? Like Interlis? Maybe it is not relevant because it is solved through inheritance already: https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/contrib/data_sources/interlis_2_3/sources/plr.py

But maybe you could have a look.

mki-c2c:
yes I'll have a look at the tolerances in Interlis.
Thanks for the hint

voisardf:
@mki-c2c @jwkaltz Could you tell me who (which canton) asked for this enhancement?

jwkaltz:
> @mki-c2c @jwkaltz Could you tell me who (which canton) asked for this enhancement?

It comes from Geoinfo (I don't remember if it was Appenzell or St.Gallen). So far they have implemented successfully the solution that @svareg proposed, which needs a database adaptation, but Geoinfo would like to obtain the same effect without needing database changes.
So in the end, this might also be interesting for Basel-Stadt.

mki-c2c:
@vvmruder 
I have added the tolerance also for Interlis mode. Thanks for the suggestion.

vvmruder (approved):
for me this looks good now.

PR #1574, opened 2022-06-08
Improve performances by using eager loading in some cases

remyguillaume:
There are performance problems because of the lazy loading of legend entries and documents.
Too many SQL queries are sent to the database, and this make the whole generation slow, especially for "big" themes.

In order to improve this, I've enabled an eager loading for those objects, so that a unique initial big query will be made.
In my case, I could reduce the generation time of a export for a big parcel from 75 sec to 21 sec.

If the changes are ok for you, it would be cool if you could merge it.

### DIFF
--- a/pyramid_oereb/contrib/data_sources/standard/sources/plr.py
+++ b/pyramid_oereb/contrib/data_sources/standard/sources/plr.py
@@ -6,6 +6,7 @@ from geoalchemy2.shape import to_shape, from_shape
 from shapely.geometry import Point, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, \
     GeometryCollection
 from sqlalchemy import text, or_
+from sqlalchemy.orm import subqueryload
 
 from pyramid_oereb import Config
 from pyramid_oereb.core import b64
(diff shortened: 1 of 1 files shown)


peterschaer:
I was able to test your patch in my dev environment and I noticed an improvement in creating complex extracts. With this one [here](https://www.oereb2.apps.be.ch/extract/json?egrid=CH368746359439) it went down from 9 sec to 6 sec and the number of SQL statements issued went down from 1891 to 137! The improvement is smaller with simpler extracts but with complex extracts the improvement is considerable. Thank you very much!

michmuel:
Seems to be an interesting point. I will have a look at it tomorrow.

michmuel:
If we change standard/sources/plr.py it is also required to change oereblex/sources/plr.py due to inheritage. There are no documents in OerebLex models. In addition, also the interlis sources might profit from some improvements.

I have here (https://github.com/openoereb/pyramid_oereb/tree/subqueryload) a branch that works with our data (oereblex- and interlis-models). It contains adaptions for all three sources and, despite of the name, uses selectinload() instead of subqueryload(). @peterschaer could you test it on your system to see what is the performance?

peterschaer:
Thanks @michmuel I will give it a try next week (Tuesday or Wednesday).

peterschaer:
I tested the changes in https://github.com/openoereb/pyramid_oereb/tree/subqueryload and observed the same performance improvements as in the changes provided by @remyguillaume in his fork. So this seems ok to me too. Thanks!

vvmruder:
@remyguillaume @michmuel @peterschaer Is this ready for merge? Or do we have still investigation onging here?

@michmuel => does this apply to INTERLIS as well?

vvmruder:
@michmuel As far as I unsderstand this PR can be closed in the favour of this one #1580. Right?

michmuel:
@vvmruder Correct.

vvmruder:
As revealed in comments this PR is closed in favor of #1580

PR #1578, opened 2022-06-20
raise error instead of silently go over

### LINKED ISSUE #1537: Should an error be raised here
https://github.com/openoereb/pyramid_oereb/blob/af0e124f1fa20d6cca5df8e505da2346f58c8df0/pyramid_oereb/contrib/data_sources/standard/sources/plr.py#L248

@kdeininger @michmuel @jwkaltz @mki-c2c

vvmruder:
fixes #1537

### DIFF
--- a/pyramid_oereb/contrib/data_sources/standard/sources/plr.py
+++ b/pyramid_oereb/contrib/data_sources/standard/sources/plr.py
@@ -229,9 +229,10 @@ class DatabaseSource(BaseDatabaseSource, PlrBaseSource):
             ))
 
         else:
-            log.warning('Received geometry with unsupported type: {0}. Skipped geometry.'.format(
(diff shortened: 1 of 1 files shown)


jwkaltz:
I was wondering about that else for some time now, meaning, when it actually occurs.
Did you run into a case where this is occurs?

vvmruder:
actually this prevents pyramid_oereb from silently deliver wrong data :) We discussed it, that we want to see an error here instead producing an extract without the wrong geometry (where ever it came from).

michmuel:
It just checks the type, not the validity of the geometry, right?

vvmruder:
@michmuel Yes you are right. But in my opinion it makes no sense to go over a geometry type which is not implemented silently. This was what we discussed.

michmuel:
That's true.

michmuel (approved):

PR #1579, opened 2022-06-24
Qr code functions

voisardf:
Adding QR code to extract and offer webservice to get QRCodeRef

### DIFF
--- a/dev/config/pyramid_oereb.yml.mako
+++ b/dev/config/pyramid_oereb.yml.mako
@@ -90,6 +90,9 @@ pyramid_oereb:
     # Whether to display the Certification section in the pdf extract or not.
     # Default to true
     display_certification: false
+    # Whether to display the QR code in the pdf extract or not.
+    # Default to true
+    display_qrcode: true
     # Group elements of "LegalProvision" and "Hints" with the same "Title.Text" together yes/no
     # Disabled by default.
     group_legal_provisions: false
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
@@ -95,6 +95,10 @@ class Renderer(JsonRenderer):
             'display_certification', False
         )
 
+        extract_as_dict['Display_QRCode'] = print_config.get(
+            'display_qrcode', False
+        )
+
         spec = {
             'layout': Config.get('print', {})['template_name'],
             'outputFormat': 'pdf',
--- a/pyramid_oereb/core/readers/extract.py
+++ b/pyramid_oereb/core/readers/extract.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 import logging
+import qrcode
+import io
 from pyramid.path import DottedNameResolver
 
 from shapely.geometry import box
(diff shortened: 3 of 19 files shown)


michmuel (changes requested):
- I can only produce a pdf with this qr-branch when I deactivate pdfa conformance. Is that true? If yes, why?
- The qr-code is an optional element in the static extract. How can I configure if it is printed or not?

voisardf:
@michmuel regarding your first question: we obviously still have transparency in some of our legend symbols (contaminated transport & military sites). Calling a PDF on a real_estate where only restrictions with fixed symbols exist, the PDF is created with QR code
Regarding your second question, I have yet to implement this option

michmuel:
@voisardf 
I deployed the latest qr-branches of pyramid_oereb and pyramid_oereb_mfp in our test env using our cantonal data. When generating a pdf I get following error message from mapfish_print:
10:34:13.290 [http-nio-8080-exec-3] WARN o.m.print.servlet.BaseMapServlet - Error while processing request: com.lowagie.text.pdf.PdfXConformanceException: The /SMask key is not allowed in images.
As no federal data was concerned in the extract and I believe we have no transparent images in our cantonal data, I was wondering if you have tested if the generated qr image is compatible with PDFA1.

voisardf:
@michmuel so if I get you right, you have a pdf-A1 compliant PDF extract when deactivating the QR code option? and an error if not?
Because I can not reproduce it with our data. Only if I call the PDF for the test data (CH113928077734) I get the S/mask error.

here's what I get with the latest state of the QR PR's
[20220704071730_extract.pdf](https://github.com/openoereb/pyramid_oereb/files/9037262/20220704071730_extract.pdf)

https://tools.pdfforge.org/validate-pdfa says it's PDF/A-1A compliant

vvmruder:
What I can give as debug info about the QR-Code image:
```
[XXXXXX@XXXXX pyramid_oereb]$ exiftool qrcode.png 
ExifTool Version Number         : 12.42
File Name                       : qrcode.png
Directory                       : .
File Size                       : 572 bytes
File Modification Date/Time     : 2022:07:04 09:38:15+02:00
File Access Date/Time           : 2022:07:04 09:38:21+02:00
File Inode Change Date/Time     : 2022:07:04 09:38:15+02:00
File Permissions                : -rw-r--r--
File Type                       : PNG
File Type Extension             : png
MIME Type                       : image/png
Image Width                     : 330
Image Height                    : 330
Bit Depth                       : 1
Color Type                      : Grayscale
Compression                     : Deflate/Inflate
Filter                          : Adaptive
Interlace                       : Noninterlaced
Image Size                      : 330x330
Megapixels                      : 0.109
```

It is PNG but not transparent as far as I recognize. It would have RGBA as color. Or an other transparent channel.

michmuel:
I will redo the test this afternoon to see if I can resolve the issue or if I can report in more detail.

vvmruder:
@michmuel @voisardf I keep this open. But from my perspective it is done and tested. We wait what @michmuel will find with BL data

voisardf:
@michmuel cool! 
the _dc is related to the caching (disable caching) as far as I rember. you can ignore it
@vvmruder I guess that PR could be merged now, if the tests are satisfying - the PDF seems to be :)

michmuel (approved):

vvmruder:
Ok I merge

PR #1580, opened 2022-06-27
add "selectinload" in queries

michmuel:
Relates to issue https://github.com/openoereb/pyramid_oereb/issues/1508.
Basically extends (adding modifications to oereblex and interlis sources) and modifies (selectinload instead of subqueryload) PR https://github.com/openoereb/pyramid_oereb/pull/1574.

### DIFF
--- a/pyramid_oereb/contrib/data_sources/interlis_2_3/sources/plr.py
+++ b/pyramid_oereb/contrib/data_sources/interlis_2_3/sources/plr.py
@@ -7,6 +7,7 @@ from geoalchemy2.shape import to_shape, from_shape
 from shapely.geometry import Point, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, \
     GeometryCollection
 from sqlalchemy import or_
+from sqlalchemy.orm import selectinload
 
 from pyramid_oereb import Config
 from pyramid_oereb.core import b64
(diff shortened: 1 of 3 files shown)


vvmruder:
Just because I'am interested. How much performance do we gain with this?

vvmruder (approved):

svareg (approved):
Just tested, and it works well, thanks .

svareg:
@vvmruder The performance is a lot better with "big" real estate, where a lot of legends and a lot of documents must be loaded from DB. At BS, we can win up to 60% of the whole generation time for a Json. For small real estates, the gain is smaller (10-20%).

But I'll have to say that the performance are still not as good as with the previous 1.8 version. 
If I can find some time, I'll try to investigate this later.

vvmruder:
@svareg I know we lack of performance still. This is mostly related to new SQLAlchemy version: https://docs.sqlalchemy.org/en/14/faq/performance.html

We did apply some of the measures but it stays a lot slower than before.

In addition we have now some real relations between DB-entities which did not exist before in the FED model.

I do have a joker for that. I will provide it in separate PR.

For now I vote to merge this PR

vvmruder:
So we merge this too?

voisardf:
@michmuel It works for me, but I can't tell you, if there is really a significant difference - I did not test a lot of big real estates before. But even for our biggest real estates it's around 10 secs only
I think we should merge this PR as it brings performance gains

PR #1582, opened 2022-07-04
use env variable for virtualenv location consistently

mki-c2c:
some tests were skipped due due incorrect path to timestamp file


vvmruder (approved):
seems to be ok

marionb (approved):
LGTM!

PR #1586, opened 2022-07-07
bugfix other legend in extract

svamaa:


voisardf (approved):
@svamaa thanks, works for me on local instance

pyramid_oereb/contrib/data_sources/standard/sources/plr.py
```diff
@@ -531,14 +532,30 @@ def read(self, params, real_estate, bbox):  # pylint: disable=W:0221
...
+
                         self.records = []
-                        legend_entries_from_db = self.collect_legend_entries_by_bbox(session, bbox)
                         for geometry_result in geometry_results:
                             self.records.append(
                                 self.from_db_to_plr_record(
                                     params,
                                     geometry_result.public_law_restriction,
-                                    legend_entries_from_db
+                                    next(elem for elem in legend_entries_from_db
```
vvmruder, line 557: It looks good in general. I only want to mark that this approach leads to potential decrease of performance because we introduce a query to DB per geometry. Did you check this for a parcel witch big amount of PLR's?

vvmruder (commented):
Looks good so far. Please have a look at my in code comment.

pyramid_oereb/contrib/data_sources/standard/sources/plr.py
svamaa: The legend_entries used to be loaded once per PLR before as well. 
Now, they are loaded once per law_status, per PLR. So, only if there are more than one law status, the query will be executed more often. But I did not test it explicitly.

The alternative would be to adjust the LegendEntry Record to also store the law_status, I guess.

vvmruder: Ok thats true. So from my point of view this is ok.

PR #1589, opened 2022-07-13
Add image library dependency, needed by qrcode

jwkaltz:


voisardf (approved):
thanks. Works for me (both, with pillow or image)

PR #1591, opened 2022-07-22
Add default index for oereblex documents

jwkaltz:


voisardf (approved):
As it works in BS and SG, I think it's safe :wink:

PR #1592, opened 2022-08-03
Sort theme lists

jwkaltz:
The theme lists should be sorted in all cases, according to extract_index.
So far, we have been sorting them only in the special case when a theme gets shifted to the not concerned themes due to the tolerance check: https://github.com/openoereb/pyramid_oereb/blob/bfa5ddc491d611ec7f8a93e784d696e2621d9b94/pyramid_oereb/core/processor.py#L133-L134

### DIFF
--- a/pyramid_oereb/core/readers/extract.py
+++ b/pyramid_oereb/core/readers/extract.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import logging
+from operator import attrgetter
 from pyramid.path import DottedNameResolver
 
 from shapely.geometry import box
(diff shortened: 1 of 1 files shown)


vvmruder (approved):
It looks ok. I was a bit confused because we had the strategy to exactly deliver the order as defined in the configuration. But this seems to have no effect any more since extract index for themes is defined now in application database.

jwkaltz:
> It looks ok. I was a bit confused because we had the strategy to exactly deliver the order as defined in the configuration. But this seems to have no effect any more since extract index for themes is defined now in application database.

Indeed, and since the plrs are sorted by the theme extract_indexes, the theme list (which ends up in the table of contents) should also be sorted the same way.

michmuel:
I will test it in our environment this week.

michmuel (approved):
Works for us. Thanks!

PR #1597, opened 2022-08-10
1568 update dependencies

mki-c2c:
update dependencies
- sphinx cannot be updated easily: there is a conflict between the latest versions of sphinx and flake8
- geoalchemy2 already has the latest version, but is displayed incorrectly in requires.io


vvmruder (approved):
Tested locally and it passed. :+1:

jwkaltz (approved):

PR #1603, opened 2022-08-29
Geo 5769 tolerance per geom type

mki-c2c:
New possibilities for configuration:
either leagsxy conf: tolerance + scalar
or new conf per geometry:
{"Point": 0.1, "Polygon": 0.05}
or dict for all:
{"ALL": 0.2}

### DIFF
--- a/dev/config/pyramid_oereb.yml.mako
+++ b/dev/config/pyramid_oereb.yml.mako
@@ -48,7 +48,7 @@ pyramid_oereb:
     # Define whether all geometry data must be included when sending the data to the print service
     with_geometry: True
     # Base URL with application of the print server
-    base_url: https://oereb-dev.gis-daten.ch/oereb/report/create
+    base_url: https://oereb-pdf.gis-daten.ch/oereb/report/create
     token: 24ba4328-a306-4832-905d-b979388d4cab
     use_wms: "true"
     validate: "false"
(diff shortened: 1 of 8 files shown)


jwkaltz (commented):
Thanks @mki-c2c , the code looks good to me;
could we add an example configuration to https://github.com/openoereb/pyramid_oereb/blob/master/dev/config/pyramid_oereb.yml.mako ?

mki-c2c:
@jwkaltz : you mean a commented stub in the config ? Because I am a bit afraid to put a default tolerance which changes quite significantly the search behaviour. So we might put some comments with a default configuration which can be uncommented by someone who wants to use this config ?

jwkaltz:
> @jwkaltz : you mean a commented stub in the config ? Because I am a bit afraid to put a default tolerance which changes quite significantly the search behaviour. So we might put some comments with a default configuration which can be uncommented by someone who wants to use this config ?

Yes you're right, the "standard configuration" shouldn't have that enabled.
So adding an example configuration for one of the themes (the first one which is of type geometrycollection is probably best), but commenting it out, seems to be the best approach.

dev/config/pyramid_oereb.yml.mako
```diff
@@ -48,7 +48,7 @@ pyramid_oereb:
     # Define whether all geometry data must be included when sending the data to the print service
     with_geometry: True
     # Base URL with application of the print server
-    base_url: https://oereb-dev.gis-daten.ch/oereb/report/create
+    base_url: https://oereb-pdf.gis-daten.ch/oereb/report/create
```
jwkaltz, line 51: This shouldn't be in this PR

jwkaltz:
@svamaa perhaps this new feature is relevant for Basel-Stadt as well? As I understand, it addresses the need that @svareg had a while ago.

vvmruder (approved):
Looks ok code wise. About the tolerance check itself I fell still not convinced. But you get an approve.

PR #1606, opened 2022-09-22
change order of ExtractIdentifier & MunicipalityCode

svamaa:
After the technical acceptance from the Bund, we got a change request that the XML-Extract wasn't valid against its XSD-Schema.

The order of ExtractIdentifier/QRCodeRef and MunicipalityCode/SubunitOfLandRegister was false.

Files are corrected accordingly.


voisardf (commented):
Have you already validated the generated xml after this fix?

vvmruder (approved):
Looks good :) Hope fed is satisfied

svamaa:
In order to 

> > Have you already validated the generated xml after this fix?
> 
> Hello I've just validate the generated XML according to the XSD, and I've noticed something else:
> 
> I get an error on this line :
> 
> https://github.com/openoereb/pyramid_oereb/blob/65a6939cb69171603a0fd5a500fddd493b087715/pyramid_oereb/core/renderer/extract/templates/xml/extract.xml#L87
> 
> ```
> <data:UpdateDateCS>
>             ${extract.creation_date.strftime(date_format)}
> </data:UpdateDateCS>
> ```
> 
> When validating, I get the following error:
> 
> ```
> >>> schema.validate(xml_doc)          
> False
> >>> schema.error_log
> extract_test.xml:1071:0:ERROR:SCHEMASV:SCHEMAV_ELEMENT_CONTENT: Element '{http://schemas.geo.admin.ch/V_D/OeREB/2.0/ExtractData}SubunitOfLandRegister': This element is not expected. Expected is ( {http://schemas.geo.admin.ch/V_D/OeREB/2.0/ExtractData}MunicipalityCode ).
> extract_test.xml:5586:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element '{http://schemas.geo.admin.ch/V_D/OeREB/2.0/ExtractData}UpdateDateCS': '
>             2022-09-26T10:34:41
>         ' is not a valid value of the local atomic type.
> ```
> 
> This was not noticed from Bund as an error, but the validation fails with standard python tools. Just removing the `carriage returns` solves this issue, and then the validation passes 100%:
> 
> ```
> <data:UpdateDateCS>${extract.creation_date.strftime(date_format)}</data:UpdateDateCS>
> ```
> 
> ```
> >>> schema.validate(xml_doc)
> True
> ```
> 
> I don't have the right to push on this branch, so if one of you would do this for me, it would be great ! Otherwise we can just wait that @svamaa returns from her holiday :) Thanks 🙂

Done ;)

voisardf:
@svamaa I think you can merge this PR. thanks!

PR #1607, opened 2022-10-03
sort plr within themes

svamaa:


svamaa:
How does Codecov fail for the new method(https://app.codecov.io/gh/openoereb/pyramid_oereb/pull/1607) when a test was used to cover them?
When I run the tests locally, it says codecov is increased..

voisardf (approved):
Tested locally with our data. Works as expected. Thanks! 😃

pyramid_oereb/core/hook_methods.py
```diff
@@ -124,16 +125,52 @@ def produce_sld_content(params, real_estate_config):
     return template.render(**template_params)
 
 
+def compare(a, b):
+    """
+    This is the method to sort a plr list (while respecting the theme order).
+    This standard sorts the plrs within themes, you can set your configuration
+    to a different method if you need a specific sorting.
+
+    Args:
```
michmuel, line 134: Maybe, for Args the exact types could be listed. I would also reformulate the sentence for Returns.

michmuel (changes requested):
The code looks good to me, thanks. We do not have any subthemes, so I cannot test the sorting in real. But using the branch in our environment works well (using the sorting has no effect as expected).
I have only one question and a small issue concerning the commenting.

pyramid_oereb/core/hook_methods.py
```diff
@@ -124,16 +125,52 @@ def produce_sld_content(params, real_estate_config):
...
+        a: Public law restriction to be compared to b
+        b: Public law restriction to be compared to a
+
+    Returns:
+        Integer: is 1 if b comes before a -> reorder
+    """
+    if a.theme.code == b.theme.code and a.sub_theme == b.sub_theme:
+        # Only impact the order of elements which are in the same sub_code
+        value_a = a.law_status.code
```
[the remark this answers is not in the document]
svamaa: A theme in oereb can always have max. 2 different law status: 'inKraft' and ('AenderungOhneVorwirkung' OR 'AenderungMitVorwirkung').

pyramid_oereb/core/hook_methods.py
svamaa: Done.

michmuel (approved):

PR #1609, opened 2022-10-18
Minor library updates

jwkaltz:
Minor library updates, before release


michmuel (approved):
Works for us. Thanks.

voisardf (approved):
Worked locally for us too. thanks

PR #1611, opened 2022-10-24
Update example of using swisstopo address search

jwkaltz:
Update example configuration explaining how to use swisstopo address search
(new package path, and make it clearer that a model is still needed)


voisardf (approved):
looks good to me. Thanks

jwkaltz:
> In my case, the alternative config was not running until the following line at the end of the yml-File was uncommented. It would be helpful to mention this too
> 
> `# - pyramid_oereb.address.source.params.db_connection`

The configuration of the pyramid_oereb dev project uses Mako templating for the database connection, so it does not have this section.
In the aargau_oereb setup, we use environment variables instead, which is why the extra section listing the configuration items which use environment variables is required.

PR #1612, opened 2022-11-02
Static extract (PDF) without use_wms must embed the images

jwkaltz:
For static extract (PDF) generation, when the PDF generation process shall not call the pyramid_oereb service to retrieve all the images (legend icons etc.), we need to embed the images within the extract.

Currently, the xml2pdf proxy contains a "use_wms" flag, but setting use_wms to false is only consistent when the extract contains images.

### DIFF
--- a/pyramid_oereb/core/views/webservice.py
+++ b/pyramid_oereb/core/views/webservice.py
@@ -375,7 +375,7 @@ class PlrWebservice(object):
         # 2. use images if use_wms is not true
         if extract_format == 'pdf':
             with_geometry = Config.get('print', {}).get('with_geometry', True) or with_geometry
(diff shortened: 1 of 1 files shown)


voisardf (approved):
Code looks good to me, but I did not test it locally as we do not use the case without WMS.
do you have a possibility to test it for a canton?

jwkaltz:
> Code looks good to me, but I did not test it locally as we do not use the case without WMS. do you have a possibility to test it for a canton?

Partially, yes: I have tried this code change manually on the Glarus infrastructure, pyramid_oereb then does try to generate the extract with embedded images when use_wms = false. However, creating an extract WITHIMAGES currently generally fails at Glarus, but that's not directly related to this subject.

PR #1613, opened 2022-11-08
Xml2pdf proxy: less logs in error case

jwkaltz:
Xml2pdf proxy: less logs in error case (depending on log level), because always outputting extract data makes log analysis difficult.


voisardf (approved):
looks good to me, thanks

PR #1614, opened 2022-11-15
Fix xml2pdf usage when using embedded images

jwkaltz:
xml2pdf is already configured via "use_wms" regarding the issue of using embedded images or not, therefore this should never be disallowed.

This is a follow up to the pull request https://github.com/openoereb/pyramid_oereb/pull/1612 , which was thus incomplete.

Note: this code change was tested on Glarus server.


voisardf (approved):
looks good, thanks

PR #1615, opened 2022-11-15
Maintenance library updates

jwkaltz:
Maintenance library updates (presumed to be low risk) in preparation of new minor release


michmuel (approved):
Did not encounter any problems. Thanks.

PR #1619, opened 2022-12-13
Allow to always add real estate geometry in XML extract

jwkaltz:
Add configuration parameter to always add real estate geometry in XML extract, 
independently of general parameter GEOMETRY.

Use case: producing the extract for the xml2pdf service, which requires the real estate geometry (limit), but not all geometries. Having this configuration possibility reduces the size of the payload sent to the print service by up to 50%.

Tested on Glarus server.


michmuel (approved):
I did not test it, but the code looks good to me.

voisardf (approved):
looks good to me

PR #1625, opened 2022-12-27
Upgrade to pyramid 2 framework

### LINKED ISSUE #1137: Upgrade to pyramid 2
pyramid_oereb is currently not compatible with pyramid 2, there is an error
`oereb-server_1  | ModuleNotFoundError: No module named 'pyramid.compat'
`

Note that the CI tests did not detect that pyramid_oereb will not yet work with pyramid 2. They should be improved to detect this problem.

jwkaltz:
Fixes #1137 

In current state of the code, no incompatibility with the pyramid 2 framework could be found. The framework components which are problematic in this upgrade (see https://docs.pylonsproject.org/projects/pyramid/en/latest/whatsnew-2.0.html) are not used in pyramid_oereb. 

Additional manual tests for JSON/XML/PDF were successful.


voisardf (approved):
Tested locally and working for me

PR #1645, opened 2022-12-28
Simplify print proxy code

jwkaltz:


voisardf (commented):
I'm not sure if this structure has originally been implemented this way on purpose regarding to the interlis 2.3 multilingual definition? @michmuel @vvmruder

jwkaltz:
> I'm not sure if this structure has originally been implemented this way on purpose regarding to the interlis 2.3 multilingual definition? @michmuel @vvmruder

Perhaps, but in the current implementation, the code is redundant - the only purpose of such redundancy could be to have easier inheritance when adding a custom mapfish-print proxy, but for these functions, I don't see a possible use case.

voisardf (approved):
code looks good to me otherwise, thanks.

PR #1647, opened 2022-12-28
Fix shapely deprecation warnings

jwkaltz:
Fix shapely deprecation warnings (in preparation for upgrade to Shapely 2)


voisardf (approved):
Tested locally and working for me, including the PDF generation with identical results (areas) for a test real-estate

Issue #1646, opened 2022-12-28
Fix warnings in tests


jwkaltz:
Fix warnings in tests,
for example
`tests/core/records/test_document.py::test_copy
  /workspace/pyramid_oereb/core/records/documents.py:86: UserWarning: Type of "only_in_municipality" should be "int"
    warnings.warn('Type of "only_in_municipality" should be "int"')
`

Fixes within pyramid_oereb:
- [x] for Shapely, see https://github.com/openoereb/pyramid_oereb/pull/1647 and https://github.com/openoereb/pyramid_oereb/pull/1662
- [x] for other pyramid_oereb internally generated warnings, see https://github.com/openoereb/pyramid_oereb/pull/1649
- [x] for Interlis test, see https://github.com/openoereb/pyramid_oereb/issues/1651
- [x] for SQLAlchemy usage, see https://github.com/openoereb/pyramid_oereb/issues/1665
- [x] for PyPDF2 deprecation, see https://github.com/openoereb/pyramid_oereb/pull/1667
- [x] for Geoalchemy usage, see https://github.com/openoereb/pyramid_oereb/pull/1669
- [x] for capturing expected warnings, see https://github.com/openoereb/pyramid_oereb/pull/1648

Fixes depending on external libraries: 
- [x] for geolink_formatter, see https://github.com/openoereb/geolink_formatter/issues/198
- [x] for pyreproj library, see https://gitlab.com/geo-bl-ch/python-reprojector/-/issues/3
- [x] for c2cwsgiutils sqlalchemy warnings, done in https://github.com/camptocamp/c2cwsgiutils/pull/1716

PR #1648, opened 2022-12-28
Pytest: capture expected warnings

jwkaltz:
Pytest: explicitly check that warnings are actually emitted where expected in type check tests
(but don't have them outputted as test warnings, to avoid confusing warnings in test executions)

Partially addresses #1646


michmuel (approved):
Code looks good to me. However, reviewing this now I think some warnings should actually be errors.

PR #1243, opened 2021-06-30
Adapt document classes

kdeininger:
Closes #1176 and #1249

PR #1649, opened 2022-12-28
Remove wrong warnings when running tests

jwkaltz:
Reduce number of warnings in test suite:
* Correcting data in test data where it was not correct (except for test cases intentionally testing behavior when data is wrong)
* provide more infos in type warnings

Partially addresses https://github.com/openoereb/pyramid_oereb/issues/1646


pyramid_oereb/core/records/documents.py
```diff
@@ -60,34 +60,42 @@ def __init__(self, document_type, index, law_status, title, responsible_office,
                                  'pyramid_oereb.core.records.document_types.DocumentTypeRecord '
                                  'but was {0}'.format(type(document_type)))
         if not isinstance(index, int):
-            warnings.warn('Type of "index" should be "int"')
+            warnings.warn('Type of "index" should be "int"'
```
michmuel, line 63: Why do we agreed here to have warnings and not errors? Same question for the warnings in the other files.

jwkaltz: I don't know if that was discussed explicitly.
It was introduced here https://github.com/openoereb/pyramid_oereb/pull/1243

michmuel: I think if the right type is necessary an error would be better. But if it caused no problems so far, we can address it later.

jwkaltz: Yes, it would be a different issue.
In fact in this PR, I touched this part just to get more information on the cause of the warnings,
the PR itself removes the cause for the warnings, which was incomplete test data.

michmuel (approved):

PR #1650, opened 2022-12-29
Remove version in code

jwkaltz:
Remove version in code: it seems to be a left-over from initial development


voisardf (approved):
nice catch, thanks

PR #1662, opened 2023-01-05
Fix shapely deprecation warning for type attribute

jwkaltz:
Fix shapely deprecation warning for type attribute usage 
- this was missed in https://github.com/openoereb/pyramid_oereb/pull/1647


michmuel (approved):
Which shapely version used "type"?

jwkaltz:
> Which shapely version used "type"?

I suppose it has been available for a long time in Shapely, it was deprecated here: https://github.com/shapely/shapely/pull/1492

In our code, we had a mixture, sometimes using geom_type and sometimes type.
After this PR, we will be using geom_type everywhere (unless I missed something).

PR #1664, opened 2023-01-05
Test coverage: add a linestring theme

jwkaltz:
Partially addresses #1646, regarding geoalchemy warnings.

In the test coverage,
* define geometry_type to avoid invalid usage of geoalchemy - but this by itself creates errors due to geometry type mismatch in linestring tests,
* this PR thus also introduces a linestring theme, which shall be used to run tests with linestring tables

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -650,6 +650,74 @@ def contaminated_sites(pyramid_oereb_test_config, dbsession, transact, wms_url_c
 
 
 @pytest.fixture
+def forest_perimeters(pyramid_oereb_test_config, dbsession, transact, wms_url_contaminated_sites,
+                      file_adapter):
+    del transact
+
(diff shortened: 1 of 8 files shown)


michmuel:
@jwkaltz I do not completely understand the issue here. You changed a test to a Linestring theme. But would it be a problem to have Linestrings or Multilinestrings in a GeometryCollection?

michmuel:
We should be sure that Linestrings can be correctly be handled in GeometryCollections. We have e.g. line elements in the land use topic.

jwkaltz:
> We should be sure that Linestrings can be correctly be handled in GeometryCollections. We have e.g. line elements in the land use topic.

That's OK for me, so I propose #1669, as a replacement for this PR.

jwkaltz:
Replaced by #1669

PR #1668, opened 2023-01-09
Update standard dev and test python version to 3.8

jwkaltz:
Use Python 3.8 as the "main" Python version for development tests, because Python 3.7 is EOL.

This version is used when running local tests (whereas CI coverage includes several Python versions).


mki-c2c (approved):
All tests work correctly, there seems no risk for problems with python3.8

Issue #1679, opened 2023-01-13
Relatively poor performance of GetEGRID


peterschaer:
As mentioned in the [blog article by Stefan Ziegler](http://blog.sogeo.services/blog/2023/01/11/oereb-kataster-richtig-gemacht-13.html), the performance of the GetEGRID call is rather modest for the pyramid_oereb cantons. Measurements in our infrastructure have confirmed this. An[ example call](https://www.oereb2.apps.be.ch/getegrid/xml/?GEOMETRY=true&POSTALCODE=3013&LOCALISATION=Reiterstrasse&NUMBER=11) takes almost a second (800-900ms) and that for what is actually a very simple spatial query. I analysed this request and noticed that a complete processor is always initiated for the GetEGRID command.
https://github.com/openoereb/pyramid_oereb/blob/98843fa078c2623edec54d112aca509c89a105d5/pyramid_oereb/core/views/webservice.py#L198
This initialisation takes a long time but only a small part of the initialised processor is then used by GetEGRID (only the real_estate_reader). This leads to this rather long request duration. If the GetEGRID command only initialises the part of the processor that is actually used, performance should improve. I will try to create a PR.

PR #1680, opened 2023-01-13
Improving GetEGRID performance

peterschaer:
With this change the request time goes down to 200ms on my development environment. On production it may even be quicker. Maybe someone has a more elegant solution.

### DIFF
--- a/pyramid_oereb/core/processor.py
+++ b/pyramid_oereb/core/processor.py
@@ -300,7 +300,7 @@ class Processor(object):
         return extract
 
 
-def create_processor():
+def create_processor(real_estate_only=False):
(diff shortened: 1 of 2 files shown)


peterschaer:
This PR tries to fix #1679

michmuel (approved):
Looks good and works in our environment. I did not test the improvement in performance.

vvmruder (approved):
This is a good catch. It goes in the right direction-since we load config on server boot, the next improvement can be based on that. THX @peterschaer

jwkaltz:
Hi @peterschaer , please feel free to merge your request.

PR #1682, opened 2023-01-16
update-version-geolink-formatter

michmuel:
Update of geolink_formatter to version 2.0.3. This version does not use the package "defusedxml". PR should remove warning described in https://github.com/openoereb/pyramid_oereb/issues/1646.


jwkaltz (approved):
Thanks

PR #1744, opened 2023-05-04
logo ref hook method

svamaa:


michmuel (approved):
The PR looks good to me. I tested the branch in our setting and I did not encounter any problems. However, we do not use QR-codes in our setting.

voisardf (approved):
Same here. Tested it locally with our data and it worked fine. We use the QR-Code and it got displayed as it should in the pdf too.

Issue #1387, opened 2021-11-26
Missing index-attribute in disclaimer and general_information


peterschaer:
The tables `disclaimer` (Class `Disclaimer` respectively) and `general_information` (Class `GeneralInformation` respectively) don't have an index-attribute whereas in the correspoding data model (http://models.geo.admin.ch/V_D/OeREB/OeREBKRMkvs_V2_0.ili) those classes do have such an attribute (called `AuszugIndex`). Have these attributes been omitted intentionally or is this a bug?

vvmruder:
Hi @peterschaer We skipped this because the federal data does not deliver it http://models.geo.admin.ch/V_D/OeREB/OeREBKRM_V2_0_Texte_20211021.xml
In addition http://models.geo.admin.ch/V_D/OeREB/OeREBKRMkvs_V2_0.ili is a fed suggestion which we customized to our needs. We can think about introducing the index on the elements in question.

As addition on that: In the moment this means pyramid_oereb must sort on theses index values on the server side because the DataExtract does not know about the index. And this means the clients (dynamic/PDF) need to respect server sided sorting.

Are you ok with that?

peterschaer:
That's ok for me. Thanks.

vvmruder:
@peterschaer Maybe I described it not correctly. In the moment pyramid_oereb does not support sorting via index on the elements in question. Should we put it in the pipe to implement this?

peterschaer:
Sorry for the misunderstanding. I should not have closed the issue.

As the federation does provide this information in http://models.geo.admin.ch/V_D/OeREB/OeREBKRM_V2_0_Texte_20211021.xml (the attributes AuszugIndex are mandatory in https://models.geo.admin.ch/V_D/OeREB/OeREBKRMkvs_V2_0.ili) pyramid_oereb should make use of it and therefore implement the sorting.

vvmruder:
Update to clarify and spec the problem:

- [ ] compare current version of federal defined [organization schema model](http://models.geo.admin.ch/V_D/OeREB/OeREBKRMkvs_V2_0.ili) and the pyramid_oereb [main schema](https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/contrib/data_sources/standard/models/main.py) regarding the attribute `AuszugIndex` and top up application schema where necessary with attribute `extract_index`
- [ ] Update the records with the new attribute `extract_index`
[...]

PR #1753, opened 2023-05-16
GSPOE-132 add index column for disclaimer and general_information

### LINKED ISSUE #1706: Provide new index attributes
Regarding to #1387 (please read also there for more context) we want to implement the attributes in question with the following spec:

- [ ] compare current version of federal defined [organization schema model](http://models.geo.admin.ch/V_D/OeREB/OeREBKRMkvs_V2_0.ili) and the pyramid_oereb [main schema](https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/contrib/data_sources/standard/models/main.py) regarding the attribute `AuszugIndex` and top up application schema where necessary with attribute `extract_index` `NOT NULL`
- [ ] Update the records with the new attribute `extract_index`
- [ ] Make sure all sources and readers handle it correctly (keep in mind core sources/readers and all the contrib ones)
- [ ] make sure to update import scripts where necessary (especially take care of correct handling of [federal data](http://models.geo.admin.ch/V_D/OeREB/OeREBKRM_V2_0_Texte.xml))
- [ ] For rendering (JSON,XML,PDF) this attribute wont be exposed since the extract index is not available in the extract model. We need to handle it as a soft sort index. The elements are sorted after it by the server but are not exposed to any clients. The clients need to respect and NOT CHANGE the sort delivered by server. => This is not the best solution but will work until the extract model is improved by the fed
- [ ] refactor code and tests regardingly

mki-c2c:
Fixes #1706 

harmonize database definitions according to https://github.com/openoereb/pyramid_oereb/issues/1387

### DIFF
--- a/dev/database/fed/disclaimer.json.xsl
+++ b/dev/database/fed/disclaimer.json.xsl
@@ -15,7 +15,7 @@
             "<xsl:value-of select="ili:Language"/>": "<xsl:value-of select="ili:Text"/>"<xsl:if test="not(position() = last())">,</xsl:if>
         </xsl:for-each>
         },
-        "extract_index": <xsl:choose><xsl:when test="element-available(ili:AuszugIndex) = 0"><xsl:value-of select="position()"/></xsl:when><xsl:otherwise><xsl:value-of select="ili:AuszugIndex"/></xsl:otherwise></xsl:choose>
+        "extract_index": <xsl:choose><xsl:when test="ili:AuszugIndex"><xsl:value-of select="ili:AuszugIndex"/></xsl:when><xsl:otherwise><xsl:value-of select="position()"/></xsl:otherwise></xsl:choose>
     }<xsl:if test="not(position() = last())">,</xsl:if>
 </xsl:for-each>
 ]
--- a/dev/database/fed/disclaimer.json.xsl
+++ b/dev/database/fed/disclaimer.json.xsl
@@ -14,7 +14,8 @@
         "content": {<xsl:for-each select="ili:Inhalt/ili:LocalisationCH_V1.MultilingualMText/ili:LocalisedText/ili:LocalisationCH_V1.LocalisedMText">
             "<xsl:value-of select="ili:Language"/>": "<xsl:value-of select="ili:Text"/>"<xsl:if test="not(position() = last())">,</xsl:if>
         </xsl:for-each>
-        }
+        },
+        "extract_index": <xsl:choose><xsl:when test="element-available(ili:AuszugIndex) = 0"><xsl:value-of select="position()"/></xsl:when><xsl:otherwise><xsl:value-of select="ili:AuszugIndex"/></xsl:otherwise></xsl:choose>
     }<xsl:if test="not(position() = last())">,</xsl:if>
 </xsl:for-each>
 ]
--- a/dev/database/fed/general_information.json.xsl
+++ b/dev/database/fed/general_information.json.xsl
@@ -15,7 +15,7 @@
             "<xsl:value-of select="ili:Language"/>": "<xsl:value-of select="ili:Text"/>"<xsl:if test="not(position() = last())">,</xsl:if>
         </xsl:for-each>
         },
-        "extract_index": <xsl:choose><xsl:when test="element-available(ili:AuszugIndex) = 0"><xsl:value-of select="position()"/></xsl:when><xsl:otherwise><xsl:value-of select="ili:AuszugIndex"/></xsl:otherwise></xsl:choose>
+        "extract_index": <xsl:choose><xsl:when test="ili:AuszugIndex"><xsl:value-of select="ili:AuszugIndex"/></xsl:when><xsl:otherwise><xsl:value-of select="position()"/></xsl:otherwise></xsl:choose>
     }<xsl:if test="not(position() = last())">,</xsl:if>
 </xsl:for-each>
 ]
--- a/dev/database/fed/general_information.json.xsl
+++ b/dev/database/fed/general_information.json.xsl
@@ -14,7 +14,8 @@
         "content": {<xsl:for-each select="ili:Inhalt/ili:LocalisationCH_V1.MultilingualMText/ili:LocalisedText/ili:LocalisationCH_V1.LocalisedMText">
             "<xsl:value-of select="ili:Language"/>": "<xsl:value-of select="ili:Text"/>"<xsl:if test="not(position() = last())">,</xsl:if>
         </xsl:for-each>
-        }
+        },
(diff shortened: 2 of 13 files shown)


mki-c2c:
I have implemented the additional "extract_index" columns. I have not enforced the not null constraint on the column, so it should be backward compatible. Sorting is done on this data when available.hello 

@michmuel : as discussed (quite a while ago) could you please check with your data if the import will still work without errors ?

pyramid_oereb/contrib/data_sources/standard/models/main.py
```diff
@@ -263,12 +272,14 @@ class GeneralInformation(Base):
...
         title (dict): The title of the general information (multilingual)
         content (dict): The actual information (multilingual)
+        extract_index (int): index to sort the items in the extract
     """
     __table_args__ = {'schema': app_schema_name}
     __tablename__ = 'general_information'
     id = Column(String, primary_key=True)
     title = Column(JSONType, nullable=False)
     content = Column(JSONType, nullable=False)
+    extract_index = Column(Integer)
```
michmuel, line 282: Why is "extract_index" implemented differently than in class Disclaimer?

michmuel (changes requested):
Looks good to me. I have only one question concerning the different implementations of "extract_index" in the two db models.

Merging this PR would require that users have to have a column "extract_index" in the db tables "disclaimer" and "general_information". However, values could be Null. Right?

In general, I do not understand the reason for the extract_index for disclaimers and general information as it is - as far as I see - not part of the extract schema and the placing of these texts in the pdf is clearly defined.

pyramid_oereb/contrib/data_sources/standard/models/main.py
mki-c2c: At first I had used an auto-incremented extract_index if not provided (to use  a non-nullable field) but it seemed better to have an optional field.
I harmonized both definitions.

mki-c2c:
Hello @michmuel I am sorry, it took me a long while to finalize this topic.
Yes you are right, the extract_index for disclaimer and general information is optional and it is ignored if not provided.

The field is just added for consistency with the definitions in http://models.geo.admin.ch/V_D/OeREB/OeREBKRMkvs_V2_0.ili
If the index is provided, the data can be imported without loss of information.

If other data sources do not use the index, there is no difference.

Furthermore, I added some tests, to make sure the sorting is executed correctly.

dev/database/fed/disclaimer.json.xsl
```diff
@@ -14,7 +14,8 @@
         "content": {<xsl:for-each select="ili:Inhalt/ili:LocalisationCH_V1.MultilingualMText/ili:LocalisedText/ili:LocalisationCH_V1.LocalisedMText">
             "<xsl:value-of select="ili:Language"/>": "<xsl:value-of select="ili:Text"/>"<xsl:if test="not(position() = last())">,</xsl:if>
         </xsl:for-each>
-        }
+        },
+        "extract_index": <xsl:choose><xsl:when test="element-available(ili:AuszugIndex) = 0"><xsl:value-of select="position()"/></xsl:when><xsl:otherwise><xsl:value-of select="ili:AuszugIndex"/></xsl:otherwise></xsl:choose>
```
michmuel, line 18: "extract_index": <xsl:choose><xsl:when test="ili:AuszugIndex"><xsl:value-of select="ili:AuszugIndex"/></xsl:when><xsl:otherwise><xsl:value-of select="position()"/></xsl:otherwise>

dev/database/fed/general_information.json.xsl
```diff
@@ -14,7 +14,8 @@
         "content": {<xsl:for-each select="ili:Inhalt/ili:LocalisationCH_V1.MultilingualMText/ili:LocalisedText/ili:LocalisationCH_V1.LocalisedMText">
             "<xsl:value-of select="ili:Language"/>": "<xsl:value-of select="ili:Text"/>"<xsl:if test="not(position() = last())">,</xsl:if>
         </xsl:for-each>
-        }
+        },
+        "extract_index": <xsl:choose><xsl:when test="element-available(ili:AuszugIndex) = 0"><xsl:value-of select="position()"/></xsl:when><xsl:otherwise><xsl:value-of select="ili:AuszugIndex"/></xsl:otherwise></xsl:choose>
```
michmuel, line 18: "extract_index": <xsl:choose><xsl:when test="ili:AuszugIndex"><xsl:value-of select="ili:AuszugIndex"/></xsl:when><xsl:otherwise><xsl:value-of select="position()"/></xsl:otherwise></xsl:choose>

pyramid_oereb/contrib/data_sources/standard/models/main.py
```diff
@@ -38,6 +39,11 @@
 srid = Config.get('srid')
 
 
+class Serial(UserDefinedType):
```
michmuel, line 42: Can be deleted, right?

pyramid_oereb/contrib/data_sources/standard/models/main.py
```diff
@@ -219,12 +225,17 @@ class Disclaimer(Base):
...
         title (str): The title which the disclaimer item has.
         content (str): The content which the disclaimer item has.
+        extract_index (int): index to sort the disclaimers in the extract
     """
     __table_args__ = {'schema': app_schema_name}
     __tablename__ = 'disclaimer'
     id = Column(String, primary_key=True)
     title = Column(JSONType, nullable=False)
     content = Column(JSONType, nullable=False)
+    # the custom class Serial may be used to obtain an auto incremented extract index in the DB
```
michmuel, line 235: Can be deleted, right?

michmuel (dismissed):

svamaa (commented):
Looks good to me. Thanks for finishing it ;)

callsumzg:
Hi, thanks for updating this.

May I ask you why you implement breaking changes in a minor update?

Are you aware that this can cause problems when updating pyramid oereb when one is not aware of this?

We would really, really, from the bottom of our hearts, appreciate it, if you could please at least leave a note in the release notes. For example something like this:


**Version 2.4.7**

- **BREAKING CHANGE:** Add extract_index to disclaimer and general infomation
- Interlis bug fix 
- Library upgrades (geoalchemy2, SQLAlchemy, jsonschema, lxml, responses, urllib3, pypdf)
- Test coverage improvements
- Python 3.8 is no longer explicitly supported
- Remove print proxy xml2pdf, no longer used by the community 

Regards, and thank you for considering this

Christian Sieber

michmuel:
@callsumzg Thanks for your note. You are right. Version 2.4.7 introduces changes to the db structure of the standard model (one additional field "extract_index" for tables "disclaimer" and "general_information" in the main schema). That is not obvious for a user by means of the change log and can require additional time for the implementation of configuration changes and/or the error search. Sorry for the inconvenience. We will try to avoid similar things.

PR #1763, opened 2023-05-26
Update document.py

vvmruder:
Make office at web from oereblex multilingual. As far as I understand, it should be this way.


michmuel (approved):
Yes, that's correct.

PR #1790, opened 2023-07-24
add docu for logo hook configuration

jwkaltz:
Follow-up for #1744


jwkaltz:
@svamaa as I understand, the logo hook configuration is mandatory, is that correct? With this PR, it will be reflected in the release notes.

svamaa:
> @svamaa as I understand, the logo hook configuration is mandatory, is that correct? With this PR, it will be reflected in the release notes.

yes, that's correct. Thanks for the PR!

PR #1805, opened 2023-08-21
Fix pdf archive timestamp

### LINKED ISSUE #1717: Wrong timestamp for archived pdf
In line https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py#L175 a time delta of +2h is added to utcnow() to define the filename for archiving PDF extract.
This composes a wrong timestamp regarding the actual creation date of a PDF (local time) that should be used and differe from the creation date in the logs too. It also uses the 12h notation and not the 24h. 

The date in the filename, the creation_date in the PDF (https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py#L243) and the date in the log files should be identical.

marionb:
fixes #1717 
TODO:
- [x] fix lint
- [x] fix test
- [x]  ~~check if https://github.com/openoereb/pyramid_oereb/pull/1805/commits/ac0e8c9489b142bc66de38888a49f74bb610471d also works for others -> **NOTE:** https://github.com/openoereb/pyramid_oereb/commit/ac0e8c9489b142bc66de38888a49f74bb610471d~~ Is addressed in #1814

@voisardf For changing the default name given to the PDF it would require this change in the mfp config: https://github.com/openoereb/pyramid_oereb_mfp/pull/138

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,6 +33,7 @@ services:
       - oereb-db
     user: "${LOCAL_UID}:${LOCAL_GID}"
     environment:
+      TZ: Europe/Zurich
       PGHOST: ${PGHOST:-oereb-db}
       PGPORT: ${PGPORT:-5432}
       PGUSER: ${PGUSER:-postgres}
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
@@ -119,7 +119,10 @@ class Renderer(JsonRenderer):
                     pdf_reader = PdfReader(pdf)
                     x = []
                     for i in range(len(pdf_reader.outline)):
-                        x.append(pdf_reader.outline[i]['/Page']['/StructParents'])
+                        if isinstance(pdf_reader.outline[i], list):
+                            x.append(pdf_reader.outline[i][0]['/Page']['/StructParents'])
+                        else:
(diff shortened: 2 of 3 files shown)


voisardf:
@marionb it works for me, thanks!
but I had to set the docker containers timezone to Swiss by adding the environment variable in the docker-compose.yml (https://github.com/openoereb/pyramid_oereb/blob/master/docker-compose.yml#L35) Otherwise the time would stay the standard (GMT without summer time = -2h)

    environment:
      TZ: Europe/Zurich

voisardf (approved):
works for me, but needed setting the time zone of the docker container

marionb:
OK great
yes I was expecting that part about the timezone in the container. Currently it is the same for the dev environment. Ideally we would adapt this in the dev einvernommen as well I guess but I am a bit limited on time, I will see what I can do.

What about the change I added here https://github.com/openoereb/pyramid_oereb/commit/ac0e8c9489b142bc66de38888a49f74bb610471d? Here I am not too sure about this perhaps @michmuel or @mki-c2c can you check this? And make sure it everything is still working with the change on your side as well? I am not sure why all of a sudden I get an error here.

voisardf:
@marionb ok and no worries, it's not the first priority but would be a nice to have if the time is available.

Regarding https://github.com/openoereb/pyramid_oereb/commit/ac0e8c9489b142bc66de38888a49f74bb610471d I'm not sure I encountered this issue yet?

PR #1814, opened 2023-09-06
fix error appearing in dev environment

### LINKED ISSUE #1813: Error when creating a pdf compute_toc_pages: true
With the newest MFP images and the configuration:
```
compute_toc_pages: true
```
An error is generated instead of a PDF.
For details see:
- https://github.com/openoereb/pyramid_oereb/commit/ac0e8c9489b142bc66de38888a49f74bb610471d
- https://github.com/openoereb/pyramid_oereb_mfp/issues/137

The error is:
```
File "/usr/local/lib/python3.9/site-packages/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py", line 118,
in __call__ x.append(pdf_reader.outline[i]['/Page']['/StructParents']) 
TypeError: list indices must be integers or slices, not str
```

marionb:
This fixes #1813 
when the table of contented is created an error appears if the following is configured:
```
compute_toc_pages: true
```

Once this is fixed we should set the MFP version back to 3.30 -> revert https://github.com/openoereb/pyramid_oereb_mfp/pull/139

Pleas test as well as possible!

### DIFF
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
@@ -115,7 +115,10 @@ class Renderer(JsonRenderer):
                     pdf_reader = PdfReader(pdf)
                     x = []
                     for i in range(len(pdf_reader.outline)):
-                        x.append(pdf_reader.outline[i]['/Page']['/StructParents'])
+                        if isinstance(pdf_reader.outline[i], list):
+                            x.append(pdf_reader.outline[i][0]['/Page']['/StructParents'])
+                        else:
+                            x.append(pdf_reader.outline[i]['/Page']['/StructParents'])
                     try:
                         true_nb_of_toc = min(x)-1
                     except ValueError:


michmuel (approved):
In combination with ```FROM camptocamp/mapfish_print:3.30``` no problems encountered.

jwkaltz:
> In combination with `FROM camptocamp/mapfish_print:3.30` no problems encountered.

Thanks @michmuel , did you also check if the table of contents is structured as you expect with your data?
MapFish-Print now uses a different library to produce the PDF, therefore the structure of the PDF is different. So far we did not detect any differences visible to the user, but we would feel safer if you can check that with real data.

peterschaer (approved):
I did my tests with the[ 3.30.5](https://github.com/mapfish/mapfish-print/releases/tag/3.30.5) release of mapfish_print (not with the docker image). Without the fix in this PR the PDF creation crashes - as described [earlier](https://github.com/openoereb/pyramid_oereb_mfp/issues/137). With the fix in this PR the PDF creation succeeds. The table of contents of the PDF looks like described [here](https://github.com/openoereb/pyramid_oereb_mfp/issues/137#issue-1780932017). This is ok for me. I have not noticed any other differences in the PDF so far.

michmuel:
The table of content in the PDF-Viewer is the same as documented by @peterschaer. The structuring by first level "themes" and second level "law documents" is ok. However, I would prefer to have only one level "themes" (if there is an easy fix). I do not really see the benefit of having also the laws in the table of content.

We tested the print in our dev environment this morning with the image camptocamp/mapfish_print:3.30. It seems that it runs much less stable in our OpenShift-environment than the image 3:29. I do not have a clue for this observation, but the image 3:30 might not be ready for production.

marionb:
@michmuel and @peterschaer thanks for testing
In this case I will merge this.
Since the image version ist set in the mfp project it will need to be taken care of in there.

PR #1849, opened 2023-10-25
Add Python 3.12, set this version as default

jwkaltz:


jwkaltz:
Closing: using Python 3.12 requires new c2cwsgiutils which should be done separately, in a prior step.

PR #1864, opened 2023-11-27
Update document.py

vvmruder:
Fix broken interface. If the URL is None we won't need a multilingual dict like {"de": None} but a real None. This PR is solving the issue

This is a followup of #1763 which break the tests :grimacing:

### DIFF
--- a/pyramid_oereb/contrib/data_sources/oereblex/sources/document.py
+++ b/pyramid_oereb/contrib/data_sources/oereblex/sources/document.py
@@ -199,12 +199,16 @@ class OEREBlexSource(Base):
             raise AssertionError('Missing authority for document #{0}'.format(document.id))
         if oereblex_doctype is None:
             raise AssertionError('Missing type for document #{0}'.format(document.id))
+        if document.authority_url is None:
+            authority_url = None
+        else:
+            authority_url = {language: document.authority_url}
 
         # Get document type
         document_type = Config.get_document_type_by_data_code(self._code, oereblex_doctype)
 
         # Create related office record
-        office = OfficeRecord({language: authority}, office_at_web={language: document.authority_url})
+        office = OfficeRecord({language: authority}, office_at_web=authority_url)
 
         # Get files
         records = []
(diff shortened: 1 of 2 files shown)


svamaa (commented):
Looks good to me. As soon as the tests are done for the oereblex data_sources the issues from codecov can be addressed.

tests/core/sources/test_document_oereblex.py
svamaa, line 1: I thought we'd use this structure for tests in contrib.data_sources.oereblex? https://github.com/openoereb/pyramid_oereb/tree/master/tests/contrib.data_sources.oereblex

vvmruder:
> Looks good to me. As soon as the tests are done for the oereblex data_sources the issues from codecov can be addressed.

I did the additional test which was necessary :laughing:

tests/core/sources/test_document_oereblex.py
vvmruder: its a matter of the one who will implement aka transfer them to the new structure :grimacing:

Doing it with this PR will probably exceed the scope of a quick fix

I think the tests in the new place will be implemented by: https://github.com/openoereb/pyramid_oereb/issues/1393

pyramid_oereb/contrib/data_sources/oereblex/sources/document.py
```diff
@@ -199,12 +199,16 @@ def _get_document_records(self, document, language):
             raise AssertionError('Missing authority for document #{0}'.format(document.id))
         if oereblex_doctype is None:
             raise AssertionError('Missing type for document #{0}'.format(document.id))
+        if document.authority_url is None:
```
voisardf, line 202: Isn't this a mandatory attribute?

vvmruder: At least not as it is defined now: https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/core/records/office.py#L21-L36

vvmruder: Ah I see. There is a missmatch between Document Record and Office Record.

voisardf (approved):
Looks good to me.

PR #1866, opened 2023-11-27
Improve test coverage for Interlis 2.3 data sources

voisardf:
improving test coverage for hook_methods in the interlis 2.3 sources


voisardf:
@vvmruder @michmuel @svamaa I recycled the fixtures from the standard/conftest.py - so there are more definitions already in there then I use for the hook_methods.
What do you prefer: Should I remove the elements we do not need (yet), adding them again as we add test cases, or should we do a check and remove what is potentially not needed once we finished with adding tests?

svamaa:
> ve what is potentially not needed once we finished with adding t

I'd delete all that's not needed now. We still have all the fixtures in the code history if needed. I think this way it's easier to avoid unneeded code and keep the overview on the needed fixtures.

voisardf:
cleaned up and ready for review

michmuel (approved):
Lokks good to me.

vvmruder (approved):
Here and there it could be improved. But it is a good start. :+1:

svamaa (approved):

voisardf:
@vvmruder I'll wait for your review before I merge

PR #1869, opened 2023-11-28
1381 - add error logs to config.py - no data

svamaa:
When a configuration is faulty, we throw exceptions. But if the configured sources do not contain data the integrator does not get the feedback accordingly.

### DIFF
--- a/pyramid_oereb/core/config.py
+++ b/pyramid_oereb/core/config.py
@@ -886,7 +886,7 @@ class Config(object):
             ConfigurationError
         """
 
-        real_estate_types_config = Config.get('real_estate_type')
+        real_estate_types_config = Config.get_real_estate_type_config()
         if real_estate_types_config is None:
             raise ConfigurationError("Missing configuration for real estate type source config")
         real_estate_type_reader = RealEstateTypeReader(
--- a/pyramid_oereb/core/config.py
+++ b/pyramid_oereb/core/config.py
@@ -269,7 +269,7 @@ class Config(object):
         )
         theme_records = theme_reader.read()
         if len(theme_records) == 0:
-            log.error('No records found for themes.')        
+            log.error('No records found for themes.')
         return theme_records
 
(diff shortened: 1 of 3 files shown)


vvmruder:
@svamaa Are you adding tests for the new lines?

svamaa:
closes https://github.com/openoereb/pyramid_oereb/issues/1381

vvmruder (approved):

pyramid_oereb/core/config.py
```diff
@@ -384,7 +423,11 @@ def _read_availabilities():
             availability_config.get('source').get('class'),
             **availability_config.get('source').get('params')
         )
-        return availability_reader.read()
+
+        avaliablity_records = availability_reader.read()
+        if len(avaliablity_records) == 0:
```
michmuel, line 428: I think the availability table can be empty if all topics are available in all municipalities.

pyramid_oereb/core/config.py
```diff
@@ -886,7 +947,10 @@ def _read_map_layering():
             map_layering_config.get('source').get('class'),
             **map_layering_config.get('source').get('params'))
 
-        return map_layering_reader.read()
+        map_layering_record = map_layering_reader.read()
+        if len(map_layering_record) == 0:
+            log.error('No records found for map layering.')
```
michmuel, line 952: I think it is not necessarily an error if the table is empty.

michmuel (changes requested):
I think two tables can be empty in a correct configuration.

pyramid_oereb/core/config.py
svamaa: Thanks for the review. I think you're right. How about logging a warning? It could still be unintentional to have the table empty? Same for availability.

michmuel: A warning seems good to me. Or even just an info.

PR #1881, opened 2023-11-29
interlis_2_3_bugfix

michmuel:
I encountered a bug in the interlis models. The backref was not correct.


voisardf (approved):
nice catch

svamaa (approved):
Well spotted! Thanks

PR #1889, opened 2023-11-30
 delete-files-and-code-for-xml2pdf

michmuel:
- Remove all code related or required by the xml2pdf print proxy
- Remove xml2pdf from config
- Delete text about "xml2pdf" in docs


svamaa (changes requested):
xml2pdf is also still mentioned in the [.gitignore file](https://github.com/openoereb/pyramid_oereb/blob/ec9c8060b7bad89252837234e3872528b0d87d54/.gitignore#L32C1-L32C44)

We should also adjust the [changes.rst](https://github.com/openoereb/pyramid_oereb/blob/delete-files-and-code-for-xml2pdf/doc/source/changes.rst) file for the updated config.yaml part.

michmuel:
- I removed the line in the .gitignore-file (the grep-cmd did not pick up these files).
- I think that users who do not use xml2pdf have not to change the pyramid_oereb.yml file necessarily. I checked in our configuration and I think we will not have to change anything. However, changes might be required if someone uses a script similar to `create_yaml.py`. 
- The removal of xml2pdf has to be mentioned in the `changes.rst`. I thought that @jwkaltz updates the file whenever publishing a new version. Or shall we already start editing it?

jwkaltz:
> * I thought that @jwkaltz updates the file whenever publishing a new version. Or shall we already start editing it?

I try to think of that when doing a new release, yes. But it's good practice to add that information together with the PR which causes the change.

svamaa:
> * I think that users who do not use xml2pdf have not to change the pyramid_oereb.yml file necessarily. I checked in our configuration and I think we will not have to change anything. However, changes might be required if someone uses a script similar to `create_yaml.py`.
Yes, you're right. The build will still work. I was confused for a sec ;)

michmuel:
> > * I thought that @jwkaltz updates the file whenever publishing a new version. Or shall we already start editing it?
> 
> I try to think of that when doing a new release, yes. But it's good practice to add that information together with the PR which causes the change.

I updated the file changes.rst.

svamaa (approved):

CHANGES.rst
```diff
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+master
+------
+- Remove print proxy "xml2pdf" (#1889)
```
vvmruder, line 6: better use a fully qualified URL for that. Tools other than GitHub.com wont be able to solve the link otherwise

vvmruder (changes requested):

vvmruder (approved):

Issue #1880, opened 2023-11-29
Improve test coverage - adapter.py


svamaa:
https://github.com/openoereb/pyramid_oereb/blob/554311ce89508bf4ab20e7ef8189560074aff428/pyramid_oereb/core/adapter.py

Increase test coverate for adapter.py

PR #1904, opened 2023-12-01
additional-tests-for-adapter

michmuel:
Closes https://github.com/openoereb/pyramid_oereb/issues/1880


svamaa:
> Options:
> 
> * change the scope of the fixture, and explicitly include the function in 20-30 tests that require this patch
> * move the adapter tests in a folder on a higher directory level. Structure of code and tests would differ in this case.
> 
> What shall we do?

How would you estimate the payload for the first option? It seems to me to be the cleaner option but I'm not sure if it exceeds the goal of this ticket?

michmuel:
I guess the first option is doable. Seems to be almost only copy and paste.

michmuel:
I had again a look at 
https://github.com/openoereb/pyramid_oereb/blob/c8201eec7848cf6f655aa9cb63f4dffd7e3586ba/tests/conftest.py#L228
Balancing effort and yield I moved the adapter tests into a separate directory.

tests/core_adapter/test_adapter_database.py
```diff
@@ -0,0 +1,61 @@
...
+    adapter = DatabaseAdapter()
+    adapter.add_connection(db_url)
+    assert 'Connection already exists: {0}'.format(db_url) not in caplog.text
+    adapter.add_connection(db_url)
+    assert 'Connection already exists: {0}'.format(db_url) in caplog.text
+
+
+def test_get_connections():
+    db_url = 'postgresql://adapter_user:adapter_pswd@adapter-tests-02:5432/adapter-db'
+    test_connection = {db_url: {"engine": "test_engne", "session": "test_session"}}
```
voisardf, line 36: small typo: test_engne instead of test_engine - not an issue for the test

voisardf (commented):
Looks good to me and reorganised tests pass in my local instance

tests/core_adapter/test_adapter_database.py
michmuel: Corrected.

svamaa (approved):
Looks good to me! 
Thanks for the work!

voisardf (approved):
sorry I didn't answer earlier. Still ok for me, thanks 👍

PR #1913, opened 2023-12-26
Fix the create_theme script

### LINKED ISSUE #1562: create_theme_tables does not work
create_theme_tables does not work.
Error is: 
  File ".venv/bin/create_theme_tables", line 33, in <module>
    sys.exit(load_entry_point('oereb-server-gr', 'console_scripts', 'create_theme_tables')())
  File "/home/apache2/oereb_server_gr/.venv/lib/python3.7/site-packages/pyramid_oereb/contrib/data_sources/create_tables.py", line 213, in create_theme_tables
    sql_file=sql_file
TypeError: create_theme_tables_() missing 1 required positional argument: 'source_class'

jwkaltz:
Fixes #1562

Example usage (when running in local development setup):

.venv/bin/create_theme_tables --configuration pyramid_oereb.yml --theme ch.Nutzungsplanung --sql-file test.sql

Note that this will only work with SQL file as output, not as direct database statement writing. Otherwise, some code refactoring would be required and I think direct database statement writing is not a good idea anyway when using a script in this way.

### DIFF
--- a/pyramid_oereb/contrib/data_sources/create_tables.py
+++ b/pyramid_oereb/contrib/data_sources/create_tables.py
@@ -198,7 +198,7 @@ def create_main_schema_tables():
 def create_theme_tables():
     parser = optparse.OptionParser(
         usage='usage: %prog [options]',
-        description='Create all content for the standard database'
+        description='Create theme tables SQL'
     )
     parser.add_option(
         '-c', '--configuration',
@@ -241,11 +241,20 @@ def create_theme_tables():
         default=False,
         help='Is the yaml file using a c2ctemplate style (starting with vars)'
     )
+    parser.add_option(
+        '--config-source',
+        type='string',
+        default='pyramid_oereb.contrib.data_sources.standard.sources.plr.DatabaseSource',
+        help='Config source to use, e.g. DatabaseSource, DatabaseOEREBlexSource, etc.'
+        'Note: parameter must match the actual config'
+    )
     options, args = parser.parse_args()
     if not options.configuration:
         parser.error('No configuration file set.')
     if not options.theme:
         parser.error('No theme code specified.')
+    if not options.sql_file:
+        parser.error('No sql filename specified.')
 
     if Config.get_config() is None:
         Config.init(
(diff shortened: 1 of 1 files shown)


pyramid_oereb/contrib/data_sources/create_tables.py
```diff
@@ -241,11 +241,20 @@ def create_theme_tables():
         default=False,
         help='Is the yaml file using a c2ctemplate style (starting with vars)'
     )
+    parser.add_option(
+        '--config-source',
```
michmuel, line 245: If I use
`--config-source pyramid_oereb.contrib.sources.plr_oereblex.DatabaseOEREBlexSource`
sql statements for creating the tables related to the standard data source are generated for the OerebLex topic.

Is it also possible to generate sql statements for creating tables for OerebLex? If not, maybe that should be stated clearer.

michmuel (changes requested):
Looks good, thanks. I have just one question.

pyramid_oereb/contrib/data_sources/create_tables.py
jwkaltz: > Is it also possible to generate sql statements for creating tables for OerebLex? If not, maybe that should be stated clearer.

Yes this should work, if you specify this in your config.yaml and then use this config-source parameter. The code that will be run is the same as the one which is run when the automated test suite is run, and, as I understand, it works fine there. This is why I added this new option to this script, because we get this possibility "for free".

michmuel: In script 
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/contrib/data_sources/oereblex/create_tables.py
the creation of tables is based on
https://github.com/openoereb/pyramid_oereb/blob/fa125c299c46467756f0a21d87cbad094e71ee93/pyramid_oereb/contrib/data_sources/oereblex/create_tables.py#L5

As far as I see, it is always the standard table structure that is created. Or do I miss something?

jwkaltz: Yes I think you are right, I misunderstood and I have now found a "create_oereblex_tables" script.
So the new option does not make sense.
I will simplify this PR accordingly.

michmuel (approved):

PR #1927, opened 2024-01-15
upgrade geolink-formatter to version 2.0.5

michmuel:
- use geolink_formatter version 2.0.5


voisardf (commented):
Code is ok. But we do not use oereblex, so I cannot test functionnality

svamaa (commented):
Sorry, we also do not use oerebLex.

jwkaltz (approved):
Since the tests pass (and we now have automated oereblex tests), this is sufficient for merging.

PR #1930, opened 2024-01-18
Use ST_DWithin instead of ST_Distance for performance reasons

remyguillaume:
Hello everyone,

For performance reasons, it would be great to replace the usage of `ST_Distance` with `ST_DWithin`.

See:
- https://www.keiruaprod.fr/blog/2021/06/11/postgis-st-distance-vs-st-dwithin.html
- https://stackoverflow.com/questions/58429252/postgres-st-distance-best-performance-query-who-finds-closest-points

With this Merge-Request, when generating a big report (for a big real estate), with our data in BS we won:
- about 25% of time for the generation of JSON reports
- about 10% of time for the generation of PDF reports.

Could please someone review this code?
Thanks !

### DIFF
--- a/pyramid_oereb/contrib/data_sources/standard/sources/plr.py
+++ b/pyramid_oereb/contrib/data_sources/standard/sources/plr.py
@@ -524,7 +524,7 @@ class DatabaseSource(BaseDatabaseSource, PlrBaseSource):
         ]
         clause_blocks = [
             text(f'ST_Intersects({extract}, {geometry_string})') if tolerance is None
-            else text(f'ST_DWithin({extract}, {geometry_string}, {round(tolerance, 3)})')
+            else text(f'ST_DWithin({extract}, {geometry_string}, {tolerance})')
             for extract, tolerance in zip([extract_point, extract_line, extract_polygon], tolerance_extracts)
         ]
         return or_(*clause_blocks)


michmuel:
Sounds good. I will have a look at it the next days.

jwkaltz:
For me it's OK to change it like this, but we should do a last minor release before merging this, as there might be some unexpected side effects in some cantons and special cases which are not covered by the automated tests, nor appear in the Basel installations. So I would propose to do a major release after merging this.

svamaa (approved):
Tested locally. Works very well.
Thank you for the contribution!

svamaa:
> @svamaa I think I do not have the right to acces the repo of @remyguillaume ? I was not able to fetch the branch for testing

I suppose the repository is public. https://github.com/remyguillaume/pyramid_oereb
I was able to clone this: https://github.com/remyguillaume/pyramid_oereb.git

jwkaltz:
Hi @remyguillaume , with what versions of Postgres/Postgis did you do your real-life tests? I am wondering if there might be differences in performance behavior depending on the versions, I will also check with @marionb .

voisardf:
> > @svamaa I think I do not have the right to acces the repo of @remyguillaume ? I was not able to fetch the branch for testing
> 
> I suppose the repository is public. https://github.com/remyguillaume/pyramid_oereb I was able to clone this: https://github.com/remyguillaume/pyramid_oereb.git

@svamaa My bad, sorry! I forgot to properly add Remy's repo first 🙈

voisardf:
On building I get a dependency error:  No matching distribution found for c2cwsgiutils==6.0.7

jwkaltz:
> On building I get a dependency error: No matching distribution found for c2cwsgiutils==6.0.7

Strange, that version exists on Pypi. I would rebase the pull request to master, that will probably solve that.

remyguillaume:
Hi @jwkaltz 

We are still using Postgres 11.2 with PostGis 2.5.2.
I think it can be great to do some tests with newer versions, thanks if you can do that.

I just found a better explaination from the official postgis website. I add the link here for documentation: https://postgis.net/workshops/postgis-intro/spatial_relationships.html#st-distance-and-st-dwithin

michmuel (commented):
The PR looks good to me and I can see the advantage of using st_dwithin. However, we do not use "tolerances" in our config, so I can neither test the proper functioning nor the gain in performance in a fast way.

As a comment, I actually think the concepts of tolerances and thresholds may be necessary because of the data quality but it seems to me that, when applying these concepts, the exact outcome of the spatial analysis is not easy to predict.

Maybe @peterschaer can also have a look at it.

peterschaer:
We don't use the tolerance parameter either. So unfortunately I can't test this PR.

jwkaltz (approved):
@marionb did some additional investigations, also concerning the question about the versions. There seems to be no significant difference in this functional domain with postgis3.
We feel the proposed change makes sense and see little to no risk in this change, only potential benefits in terms of performance (if indexes are used).
Please note however that we do not have sufficient test data on our side for full scale tests.

PR #2005, opened 2024-06-25
Add option to not verify certificate when calling the real estate WMS

jwkaltz:
Add option to not verify certificate when calling the real estate WMS. Useful for example if the WMS has a self-signed certificate.

In addition to the CI, this PR was tested locally with
http://localhost:6543/oereb/extract/json?EGRID=CH113928077734&WITHIMAGES=true

with both settings (verify_certificate True and False)

### DIFF
--- a/dev/config/pyramid_oereb.yml.mako
+++ b/dev/config/pyramid_oereb.yml.mako
@@ -238,6 +238,7 @@ pyramid_oereb:
         fr: https://wms.geo.admin.ch/?SERVICE=WMS&REQUEST=GetMap&VERSION=1.3.0&STYLES=default&CRS=EPSG:2056&BBOX=2475000,1065000,2850000,1300000&WIDTH=493&HEIGHT=280&FORMAT=image/png&LAYERS=ch.swisstopo-vd.amtliche-vermessung
       layer_index: 0
       layer_opacity: 1.0
+      verify_certificate: True
     plan_for_land_register_main_page:
       # WMS URL to query the plan for land register specially for static extracts overview page
       reference_wms:
@@ -245,6 +246,7 @@ pyramid_oereb:
         fr: https://wms.geo.admin.ch/?SERVICE=WMS&REQUEST=GetMap&VERSION=1.3.0&STYLES=default&CRS=EPSG:2056&BBOX=2475000,1065000,2850000,1300000&WIDTH=493&HEIGHT=280&FORMAT=image/png&LAYERS=ch.swisstopo-vd.amtliche-vermessung
       layer_index: 0
       layer_opacity: 1.0
+      verify_certificate: True
     visualisation:
       method: pyramid_oereb.core.hook_methods.produce_sld_content
       # Note: these parameters must fit to the attributes provided by the RealEstateRecord!!!!
--- a/dev/config/pyramid_oereb.yml.mako
+++ b/dev/config/pyramid_oereb.yml.mako
@@ -246,6 +246,7 @@ pyramid_oereb:
         fr: https://wms.geo.admin.ch/?SERVICE=WMS&REQUEST=GetMap&VERSION=1.3.0&STYLES=default&CRS=EPSG:2056&BBOX=2475000,1065000,2850000,1300000&WIDTH=493&HEIGHT=280&FORMAT=image/png&LAYERS=ch.swisstopo-vd.amtliche-vermessung
       layer_index: 0
       layer_opacity: 1.0
+      # Option to check certificate for external WMS. Default and recommended setting: True
       verify_certificate: True
     visualisation:
       method: pyramid_oereb.core.hook_methods.produce_sld_content
--- a/dev/config/pyramid_oereb.yml.mako
+++ b/dev/config/pyramid_oereb.yml.mako
@@ -238,6 +238,7 @@ pyramid_oereb:
         fr: https://wms.geo.admin.ch/?SERVICE=WMS&REQUEST=GetMap&VERSION=1.3.0&STYLES=default&CRS=EPSG:2056&BBOX=2475000,1065000,2850000,1300000&WIDTH=493&HEIGHT=280&FORMAT=image/png&LAYERS=ch.swisstopo-vd.amtliche-vermessung
       layer_index: 0
       layer_opacity: 1.0
+      # Option to check certificate for external WMS. Default and recommended setting: True
       verify_certificate: True
     plan_for_land_register_main_page:
       # WMS URL to query the plan for land register specially for static extracts overview page
--- a/pyramid_oereb/contrib/data_sources/standard/models/__init__.py
+++ b/pyramid_oereb/contrib/data_sources/standard/models/__init__.py
@@ -16,7 +16,7 @@ As of the moment of writing this includes:
 
 
 from sqlalchemy import Column, ForeignKey
-from sqlalchemy import String, Integer, Float, Date
+from sqlalchemy import String, Integer, Float, Date, Boolean
 from geoalchemy2.types import Geometry as GeoAlchemyGeometry
 from sqlalchemy.orm import relationship
 from sqlalchemy_utils import JSONType
@@ -174,7 +174,8 @@ def get_view_service(base, schema_name, pk_type):
             reference_wms (dict of str): The actual url which leads to the desired cartographic
                 representation (multilingual).
             layer_index (int): Index for sorting the layering of the view services for a theme
-            layer_opacity (float): Opacity of a view service
+            layer_opacity (float): Opacity of a view servicA
+            verify_certificate (boolean): Whether the certificate of the WMS shall be verified
         """
         __table_args__ = {'schema': schema_name}
         __tablename__ = 'view_service'
@@ -182,6 +183,7 @@ def get_view_service(base, schema_name, pk_type):
         reference_wms = Column(JSONType, nullable=False)
         layer_index = Column(Integer, nullable=False)
         layer_opacity = Column(Float, nullable=False)
+        verify_certificate = Column(Boolean, nullable=True)
 
     return ViewService
 
--- a/pyramid_oereb/contrib/data_sources/standard/models/__init__.py
+++ b/pyramid_oereb/contrib/data_sources/standard/models/__init__.py
@@ -16,7 +16,7 @@ As of the moment of writing this includes:
 
 
 from sqlalchemy import Column, ForeignKey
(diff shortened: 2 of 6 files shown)


pyramid_oereb/contrib/data_sources/standard/sources/plr.py
```diff
@@ -162,6 +162,7 @@ def from_db_to_view_service_record(self, view_service_from_db, legend_entry_reco
             Config.get('default_language'),
             Config.get('srid'),
             Config.get('proxies'),
+            Config.get('verify_certificate'),
```
michmuel, line 165: Where is 'verify_certificate' defined in file 'pyramid_oereb.yml.mako'?

pyramid_oereb/core/readers/real_estate.py
```diff
@@ -63,7 +63,8 @@ def read(self, params, nb_ident=None, number=None, egrid=None, geometry=None):
             plan_for_land_register_config.get('layer_opacity'),
             Config.get('default_language'),
             Config.get('srid'),
-            Config.get('proxies')
+            Config.get('proxies'),
+            Config.get('verify_certificate')
```
michmuel, line 67: plan_for_land_register_config.get('verify_certificate', True) ?

pyramid_oereb/core/readers/real_estate.py
```diff
@@ -73,7 +74,8 @@ def read(self, params, nb_ident=None, number=None, egrid=None, geometry=None):
             plan_for_land_register_main_page_config.get('layer_opacity'),
             Config.get('default_language'),
             Config.get('srid'),
-            Config.get('proxies')
+            Config.get('proxies'),
+            Config.get('verify_certificate')
```
michmuel, line 78: plan_for_land_register_main_page_config.get('verify_certificate', True) ?

pyramid_oereb/core/records/view_service.py
```diff
@@ -80,7 +80,7 @@ class ViewServiceRecord(object):
     """
 
     def __init__(self, reference_wms, layer_index, layer_opacity, default_language,
-                 srid, proxies=None, legends=None):
+                 srid, proxies=None, verify_certificate=True, legends=None):
```
michmuel, line 83: "verify_certificate" is inserted before "legends". Is that fine for every use of this class, or would it be better to have it at the end?

Can you update the documentation of the class?

michmuel (changes requested):
Thanks @jwkaltz. I added some comments.

It seems to me that it is a bit a partial solution to a very specific problem of a particular oereb server instance. And not a general use case. If an extract with geometry is requested, the wms for a topic could have as well a self-signed certificate. But this is not possible so for with the standard source. In addition, neglecting the certificates is not advised. Do we really need this MR?

jwkaltz:
> Thanks @jwkaltz. I added some comments.
> 
> It seems to me that it is a bit a partial solution to a very specific problem of a particular oereb server instance. And not a general use case. If an extract with geometry is requested, the wms for a topic could have as well a self-signed certificate. But this is not possible so for with the standard source. In addition, neglecting the certificates is not advised. Do we really need this MR?

Hi @michmuel , thanks for the review, I will look at your proposed changes!

Regarding the general context:
In fact the requirement does not come from within the oereb instance itself, but from the oereb server which calls a service with a self-signed certificate. I know of at least 2 cantons which use self-signed certificated for their services, because they believe it saves money to use self-signed certificates. I did advise them to not use self-signed certificates, but this is currently not an option. 
We did have a similar situation with the oereb_client, which is why this possibility was added there (https://github.com/openoereb/oereb_client/commit/be16b594e3c145fceac5ef42b79280ce04f8e245).

As far as I could tell, adding this option to the requests.get call in view_service.py is in fact a full solution for this issue. There is also the swisstopo/address.py, but swisstopo has proper certificates so this not an issue there.

pyramid_oereb/contrib/data_sources/standard/sources/plr.py
jwkaltz: at the wms level, like layer opacity - so your following comments are accurate, I have changed the code accordingly.

pyramid_oereb/core/readers/real_estate.py
jwkaltz: You are correct, I have now adapted the code

pyramid_oereb/core/records/view_service.py
jwkaltz: I moved it to the end, to reduce risks. 
Documentation is now also updated.

pyramid_oereb/core/readers/real_estate.py
jwkaltz: You are correct, I have now adapted the code

dev/config/pyramid_oereb.yml.mako
```diff
@@ -238,13 +238,15 @@ pyramid_oereb:
         fr: https://wms.geo.admin.ch/?SERVICE=WMS&REQUEST=GetMap&VERSION=1.3.0&STYLES=default&CRS=EPSG:2056&BBOX=2475000,1065000,2850000,1300000&WIDTH=493&HEIGHT=280&FORMAT=image/png&LAYERS=ch.swisstopo-vd.amtliche-vermessung
       layer_index: 0
       layer_opacity: 1.0
+      verify_certificate: True
```
svamaa, line 241: Insert comment to advice users not to use verify_certificate: False

svamaa (commented):
I agree with @michmuel, it's not ideal to open pyramid_oereb to use self-signed certificates. Wouldn't it be possible to add a self-signed certificate to the docker image of pyramid_oereb?
However, the changes don't affect the security of users, that use proper certificates.
Though, I would strongly advice in pyramid_oereb.yml.mako not to use verify_certificate: False with a comment.

pyramid_oereb/contrib/data_sources/standard/models/__init__.py
```diff
@@ -174,14 +174,16 @@ class ViewService(base):
...
+            layer_opacity (float): Opacity of a view servicA
+            verify_certificate (boolean): Whether the certificate of the WMS shall be verified
         """
         __table_args__ = {'schema': schema_name}
         __tablename__ = 'view_service'
         id = Column(pk_type, primary_key=True, autoincrement=False)
         reference_wms = Column(JSONType, nullable=False)
         layer_index = Column(Integer, nullable=False)
         layer_opacity = Column(Float, nullable=False)
+        verify_certificate = Column(Boolean, nullable=True)
```
michmuel, line 186: This would require an effort for all the users that employ the standard models as they would have to change the db structure.

dev/config/pyramid_oereb.yml.mako
jwkaltz: Thanks, I have now added a comment

pyramid_oereb/contrib/data_sources/standard/models/__init__.py
jwkaltz: Thanks for pointing this out, indeed I think that this should not be required. 
I have adapted the code, and now the standard db model does not require this attribute, we just set it to True by default.

svamaa (approved):
Looks good to me. Thank you for adding the comment.

michmuel (changes requested):
The branch works fine in our test environment. However, I spotted a few things.

dev/config/pyramid_oereb.yml.mako
```diff
@@ -238,13 +238,16 @@ pyramid_oereb:
...
```
michmuel, line 241: Add "# Option to check certificate for external WMS. Default and recommended setting: True"

pyramid_oereb/contrib/data_sources/standard/models/__init__.py
```diff
@@ -174,7 +174,7 @@ class ViewService(base):
             reference_wms (dict of str): The actual url which leads to the desired cartographic
                 representation (multilingual).
             layer_index (int): Index for sorting the layering of the view services for a theme
-            layer_opacity (float): Opacity of a view service
+            layer_opacity (float): Opacity of a view servicA
```
michmuel, line 177: "service" instead of "serviceA"

pyramid_oereb/core/readers/real_estate.py
```diff
@@ -63,7 +63,8 @@ def read(self, params, nb_ident=None, number=None, egrid=None, geometry=None):
             plan_for_land_register_config.get('layer_opacity'),
             Config.get('default_language'),
             Config.get('srid'),
-            Config.get('proxies')
+            Config.get('proxies'),
+            verify_certificate=plan_for_land_register_config.get('verify_certificate')
```
michmuel, line 67: I think "verify_certificate=plan_for_land_register_config.get('verify_certificate', True)" would be clearer.

pyramid_oereb/core/readers/real_estate.py
```diff
@@ -73,7 +74,8 @@ def read(self, params, nb_ident=None, number=None, egrid=None, geometry=None):
             plan_for_land_register_main_page_config.get('layer_opacity'),
             Config.get('default_language'),
             Config.get('srid'),
-            Config.get('proxies')
+            Config.get('proxies'),
+            verify_certificate=plan_for_land_register_main_page_config.get('verify_certificate')
```
michmuel, line 78: I think "verify_certificate=plan_for_land_register_main_page_config.get('verify_certificate', True)" would be clearer.

michmuel (changes requested):

pyramid_oereb/contrib/data_sources/standard/models/__init__.py
jwkaltz: Thanks, it is corrected now

pyramid_oereb/core/readers/real_estate.py
jwkaltz: OK, changed

pyramid_oereb/core/readers/real_estate.py
jwkaltz: OK, changed

dev/config/pyramid_oereb.yml.mako
jwkaltz: The comment is already there, subsequently to review by Anne

jwkaltz: Thanks, copied comment here as well

michmuel (approved):

PR #2006, opened 2024-06-27
Restrict the version of Numpy used

jwkaltz:
Restrict the version of Numpy used, as incompatibilities with other libraries have been detected in an operational environment (Jura).
This is quick-fix pull request; because the current CI does not detect this issue, the root cause needs to be investigated further.


domfre (approved):

jwkaltz:
After further analysis, this change would only be necessary for old versions of Python, which are no longer supported by this project.
The solution for users is to update their version of Python to at least 3.9.

PR #2009, opened 2024-07-04
update change log

michmuel:
- add a note related to the approach how packages are imported
- extend the information for version 2.4.7

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,7 @@ Changelog
 
 Notes:
 ------
-- This python package specifies the version numbers only for directly imported python packages. This approach may result in a build failure of the package if there arise incompatibilities between imported packages over time. The build process of the master branch is regularely tested in an automatic process.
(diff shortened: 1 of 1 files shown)


voisardf (approved):
thanks

CHANGES.rst
```diff
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+Notes:
+------
+- This python package specifies the version numbers only for directly imported python packages. This approach may result in a build failure of the package if there arise incompatibilities between imported packages over time. The build process of the master branch is regularely tested in an automatic process.
```
jwkaltz, line 6: Can we define "of the package" -> "of older versions of the project"

jwkaltz, line 6: regularly

jwkaltz (approved):

jwkaltz (approved):
Since we suggest that build might fail over time, we could also suggest how to work in a productive environment (create docker images for your images which you tag and do not delete as long as they are used in production)

michmuel:
> Since we suggest that build might fail over time, we could also suggest how to work in a productive environment (create docker images for your images which you tag and do not delete as long as they are used in production)

That's a good point. We could address it at the next user meeting.

PR #2034, opened 2024-08-20
Provide a general WMS verify certificate option

jwkaltz:
Provide a general WMS verify certificate option, for cantonal setups where all WMS are served with self-signed certificates

Improves https://github.com/openoereb/pyramid_oereb/pull/2005 , tested in canton Jura infrastructure.


dev/config/pyramid_oereb.yml.mako
```diff
@@ -1413,6 +1409,9 @@ pyramid_oereb:
           transfer_code: Hinweis
           extract_code: Hint
 
+  # Option to check certificate for external WMS calls. Default and recommended setting: True
```
michmuel, line 1412: Add a note that this concerns only the wms of themes using the standard and oereblex models.

michmuel (changes requested):

dev/config/pyramid_oereb.yml.mako
jwkaltz: Note added

michmuel (approved):

PR #2042, opened 2024-08-27
Default toc length config

voisardf:
A new parameter `default_toc_length` in the print configuration allows to deactivate the `compute_toc_pages` function and set a default toc page length which applies in almost every case (exceptions, see description)

- [x] new print config parameter(s)
- [x] tests

The TOC page breaks depend on the overall number of published topics, the settings regarding the land register elements and the individual disclaimer options (QR code). 
In Neuchâtel, with only one or two concerned themes, we get one TOC page, with three or more concerned topics the TOC will be at least two pages.


pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -116,6 +119,7 @@ def __call__(self, value, system):
             data=json.dumps(spec)
         )
         try:
+            log.debug('Validation of the TOC length with compute_toc_pages set to {} and default_toc_length set to {}'.format(print_config.get('compute_toc_pages'), print_config.get('default_toc_length'))) # noqa
             if Config.get('print', {}).get('compute_toc_pages', False):
```
svamaa, line 123: we can leave the if statement away.
The check if the toc pages from the actual pdf fits the extract_as_dict['nbTocPages'] we've set above should be done anyways and will be the same for all three cases.

svamaa:
I've tested this PR with compute_toc_pages=false and default_toc_length=2 and found a 30% faster pdf extract!
Great work! thank you @voisardf

doc/source/changes.rst
```diff
@@ -6,6 +6,13 @@ Changes/Hints for migration
 This chapter will give you hints on how to handle version migration, in particular regarding what you may need
 to adapt in your project configuration, database etc. when upgrading to a new version.
 
+Version 2.6.0
+-------------
+* New parameter 'default_toc_length' allows to define a default table of content pages number avoiding a second
```
michmuel, line 11: I would omit the statemtents about the 95% and the default setting. I guess default is to not set this parameter.

dev/config/pyramid_oereb.yml.mako
```diff
@@ -88,6 +88,16 @@ pyramid_oereb:
...
+    # To avoid the potentially time consuming second computing of the PDF extract and skip the the computation
+    # of the estimated TOC length, you can specify a default length for the number of TOC pages.
+    # For most of the cantons the length of the TOC is pretty consistent unless a real estate is concerned by none 
+    # or a huge number of restrictions.
+    # An additional page break might also occur if the number of published topics is close to a threshold number
+    # where the TOC fits just about on one or two pages. - for those case estimate the TOC length ist preferable.
+    # In both cases (computing an estimated length or setting a default length) the exact number of TOC pages is
+    # extracted from the created PDF and if it is different from the expected value the PDF extract is called a 
+    # second time with the correct page numbers.
+    default_toc_length: 2
```
michmuel, line 100: Maybe the name of the variable could be more precise. E.g. "most_frequent_toc_length"

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -73,17 +73,20 @@ def __call__(self, value, system):
         extract_as_dict = self._render(extract_record, value[1])
         feature_geometry = mapping(extract_record.real_estate.limit)
 
+        print_config = Config.get('print', {})
```
michmuel, line 76: If the variable 'print_config' is defined, it should thoroughly be used in subsequent parts of the code.

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -73,17 +73,20 @@ def __call__(self, value, system):
...
+
         if Config.get('print', {}).get('compute_toc_pages', False):
             extract_as_dict['nbTocPages'] = TocPages(extract_as_dict).getNbPages()
         else:
-            extract_as_dict['nbTocPages'] = 1
+            if Config.get('print', {}).get('default_toc_length', False):
+                extract_as_dict['nbTocPages'] = print_config.get('default_toc_length', 2)
```
michmuel, line 82: I would not specify a default, but test if it is a positive number.

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
```diff
@@ -131,14 +135,43 @@ def __call__(self, value, system):
                     except ValueError:
                         true_nb_of_toc = 1
 
+                    log.debug('True number of TOC pages is {}'.format(true_nb_of_toc))
```
michmuel, line 138: Maybe the expected number can also be written to the log (in case it is computed).

pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py
michmuel: Agree.

michmuel (changes requested):

michmuel:
Thanks for the pull request! Looks good to me. I added some remarks, some of them are discussible.

lopo977 (approved):
Looks good to me.

svamaa:
Looks good to me! Thank you @voisardf !

doc/source/changes.rst
```diff
@@ -6,6 +6,13 @@ Changes/Hints for migration
 This chapter will give you hints on how to handle version migration, in particular regarding what you may need
 to adapt in your project configuration, database etc. when upgrading to a new version.
 
+Version 2.6.0
+-------------
+* New parameter 'expected_toc_length' allows to define a default table of content pages number avoiding a second
+call for the pdf extract in most cases. This value should be set if most of the PDF extracts have the same number
+of TOC pages.
+Default setting: 2
```
michmuel, line 14: That is omitted now.

A change in this pull request is that the pdf is built a second time whenever the numbers of toc pages disagree between extract_as_dict['nbTocPages'] (computed/expected/1, depending on the setting) and the one in the returned pdf.

michmuel (changes requested):
Looks good to me. Thanks! I noted one small thing (maybe its good to change it here in order to not forget it).

doc/source/changes.rst
michmuel: In the current version it is only rebuild if 'compute_toc_pages' is true.

voisardf:
thanks for all your input!

Issue #1875, opened 2023-11-28
swisstopo address source: add timeout?


michmuel:
Shall we introduce a timeout at https://github.com/openoereb/pyramid_oereb/blob/1a3ba4bc9010d479473d5d88b00e31bab5ac1e7e/pyramid_oereb/contrib/data_sources/swisstopo/address.py#L64 ?

voisardf:
@michmuel personally I would go for an slightly higher value - maybe 5s?

vvmruder:
3s is enaough I would say. Otherwise we could make it configurable.

PR #2043, opened 2024-08-27
add-timout-in-address-source

michmuel:
closes https://github.com/openoereb/pyramid_oereb/issues/1875


voisardf (approved):
thanks!
should we mention the fixed timeout in the doc? at least in the change log I think

svamaa (approved):

PR #2044, opened 2024-08-27
compress all responses

svamaa:
Depending on the accept-encoding header of the request, the response will be compressed in brotli or gzip. If the header is not present or both are not supported, no compression is done.


svamaa:
Testing this locally with large parcels and many concerned themes, the compression done in python takes more time than the time win from downloading less data.
Also tested in debugtoolbar, compression itself takes several 100 ms. Download much less.
Possible solution: use nginx as proxy before oereb-server, which can compress data automatically and more performant

- drawback: more complexity and only small time win (if at all)

My suggestion: Not to implement the compression.

Issue #2037, opened 2024-08-27
Load geometries more efficiently


svamaa:
Now, there are 3 db requests to intersect the real estate with the plr.

michmuel:
There are repeated queries on geometry in the standard, oereblex and interlis sources.

standard
https://github.com/openoereb/pyramid_oereb/blob/ec86cb5af4671edd850d5a26c37b5c71dc03ad3b/pyramid_oereb/contrib/data_sources/standard/sources/plr.py#L664C1-L666C22
https://github.com/openoereb/pyramid_oereb/blob/ec86cb5af4671edd850d5a26c37b5c71dc03ad3b/pyramid_oereb/contrib/data_sources/standard/sources/plr.py#L685C1-L690C88

oereblex
[inherited from standard source]

interlis
https://github.com/openoereb/pyramid_oereb/blob/ec86cb5af4671edd850d5a26c37b5c71dc03ad3b/pyramid_oereb/contrib/data_sources/interlis_2_3/sources/plr.py#L528C1-L530C22
https://github.com/openoereb/pyramid_oereb/blob/ec86cb5af4671edd850d5a26c37b5c71dc03ad3b/pyramid_oereb/contrib/data_sources/interlis_2_3/sources/plr.py#L548C1-L553C88

Puprose of this issue: analysis if/how these repeated queries can be simplified.

svamaa:
was addressed in this PR https://github.com/openoereb/pyramid_oereb/pull/2050

PR #2045, opened 2024-08-29
reduce-number-of-spatial-queries

michmuel:
closes https://github.com/openoereb/pyramid_oereb/issues/2037


vvmruder:
@michmuel Just a quick pointer: The amount of requests was intrduced to be able to handle geometrycollections. I think this is still a case because some topics may encapsulate all three simple type kinds (point, line, polygon) in one table.

michmuel:
@vvmruder That part is not removed. The handling of geometry collections results in only one db query as well. The issue here is that the retrieval of the legend entries requires for each law status an additional spatial query.

svamaa (commented):
I've finally managed to test the branch locally with our adjustments in BS.
Seems like the json extract did not get faster, but slightly slower for the very large parcel.
However, on top of collecting the legend entries from the db, we have to filter them again. I guess this is what dissolves the time win from this branch on standard pyramid_oereb.

Issue #2038, opened 2024-08-27
TOC - review calculation


svamaa:

peterschaer:
This issue is about checking the calculation of the size of the table of contents (i.e. number of pages) made in https://github.com/openoereb/pyramid_oereb/blob/ec86cb5af4671edd850d5a26c37b5c71dc03ad3b/pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py#L8
The number of pages that the table of contents needs is used in mapfish_print.py to be compared with the actual length of the table of contents in the PDF. If the two numbers do not match, the PDF is created a second time to correct the page numbers in the table of contents. The most accurate calculation possible prevents the PDF from being created a second time and therefore increases performance.

peterschaer:
First of all, I looked at some parcels in the canton of Bern to see whether the PDF would be generated a second time. I looked at the following examples:

* Parcel with only one topic: https://www.oereb2.apps.be.ch/extract/pdf?egrid=CH368946357207&lang=de
* Parcel with two hits (but 3 subtopics): https://www.oereb2.apps.be.ch/extract/pdf?egrid=CH743546874207&lang=de
* Parcel with 5 topics (incl. 4 subtopics): https://www.oereb2.apps.be.ch/extract/pdf?egrid=CH368746359439&lang=de
* Parcel with 5 topics (incl. 4 subtopics): https://www.oereb2.apps.be.ch/extract/pdf?egrid=CH643546232754&lang=de

In all these cases, the length of the TOC is two pages. In no case will the PDF be generated a second time. In our current constellation, there is no plot in the canton of Bern that generates a table of contents of more than two pages, so that in our case at least, the PDF is only ever generated once.

@svamaa @michmuel @voisardf @lopo977 Are there examples in your cantons where the PDF is generated twice? I can't really test this from the outside.

svamaa:
Extracts with wrong calculation (results in second print request):

https://api.oereb.bs.ch/extract/pdf/?EGRID=CH198970136775
https://api.oereb.bs.ch/extract/pdf/?EGRID=CH516710897827 (only one theme concerned)

We always have 2 pages table of contents. But I could not find a case where the calculation was correct.

michmuel:
PDF computation in BL:

- 1x
  - https://oereb.geo.bl.ch/extract/pdf?egrid=CH780507397978

 - 2x
   - https://oereb.geo.bl.ch/extract/pdf?egrid=CH507859371772
[...]

PR #2047, opened 2024-08-29
Revise toc calculation

peterschaer:
adresses #2038 

I completely reworked the calculation:

- all the pixel values where verified in the templates and updated when needed
- unused variables were deleted
- qrcode (if activated) is now included
- landregister disclaimer is now included
- the ConcernedTheme-Heading cannot be calculated at runtime. The label is defined in the templates which are not accessible from the python code.

@svamaa @michmuel @voisardf @lopo977 could you test with your data?

### DIFF
--- a/dev/config/pyramid_oereb.yml.mako
+++ b/dev/config/pyramid_oereb.yml.mako
@@ -89,8 +89,9 @@ pyramid_oereb:
     # not set it will assume that only one TOC page exists, and this can lead to wrong numbering in the TOC, which
     # will be fixed by a second PDF extract call that has an impact on performance.
     compute_toc_pages: false
-    # The calculation of the TOC pages is very detailed in pyramid_oereb. However there are "earlier" page breaks, that are
-    # not (yet) within the calculation. With setting this value, the page breaks (in this case) 10 px earlier than default.
+    # The calculation of the number of TOC pages is very detailed in pyramid_oereb. However, it is not complete, 
+    # and page breaks occur slightly earlier in page than expected. With setting the value "page_break_difference"
+    # (default: 10px), the apparent (not the real) page height is reduced in order to account for this issue.
     page_break_difference: 10
     # In order to skip the computation of the estimated number of TOC pages which might return an erroneous result
     # for your setting, you can specify a default for the number of TOC pages. For most of the cantons the number of
--- a/dev/config/pyramid_oereb.yml.mako
+++ b/dev/config/pyramid_oereb.yml.mako
@@ -89,6 +89,9 @@ pyramid_oereb:
     # not set it will assume that only one TOC page exists, and this can lead to wrong numbering in the TOC, which
     # will be fixed by a second PDF extract call that has an impact on performance.
     compute_toc_pages: false
+    # The calculation of the TOC pages is very detailed in pyramid_oereb. However there are "earlier" page breaks, that are
+    # not (yet) within the calculation. With setting this value, the page breaks (in this case) 10 px earlier than default.
+    page_break_difference: 10
     # In order to skip the computation of the estimated number of TOC pages which might return an erroneous result
     # for your setting, you can specify a default for the number of TOC pages. For most of the cantons the number of
     # TOC pages is pretty constant unless a real estate is concerned by none or a huge number of restrictions.
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
@@ -6,7 +6,6 @@ log = logging.getLogger(__name__)
 
 
 class TocPages:
-
     def __init__(self, extract):
         self.total_height = 842
         self.header_height = self.compute_header()
@@ -76,7 +75,9 @@ class TocPages:
         total_size += (
             blank_space_above + not_concerned_themes_title_height + blank_space_between
         )
-        total_size += len(self.extract["NotConcernedTheme"]) * not_concerned_themes_item_height
+        total_size += (
+            len(self.extract["NotConcernedTheme"]) * not_concerned_themes_item_height
+        )
         if total_size > self.d3_height:
             log.debug(f"d3 total_size: {total_size}")
             return total_size
@@ -95,9 +96,8 @@ class TocPages:
         total_size = 0
         theme_without_data_item_height = 12  # themelist.jrxml
         total_size += (
-            (len(self.extract["ThemeWithoutData"]))
-            * theme_without_data_item_height
-        )
+            len(self.extract["ThemeWithoutData"])
+        ) * theme_without_data_item_height
         if total_size > self.d5_height:
             log.debug(f"d5 total_size: {total_size}")
             return total_size
@@ -119,7 +119,7 @@ class TocPages:
 
         space_between_info_and_disclaimer = 6  # general_info_and_disclaimer.jrxml
         total_size += space_between_info_and_disclaimer
-        
+
         # LandRegister-Disclaimer (1 title, 1 item)
         land_register_disclaimer_title_line_height = (
             8  # general_info_and_disclaimer.jrxml
--- a/pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
+++ b/pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
@@ -5,13 +5,15 @@ import textwrap
 log = logging.getLogger(__name__)
 
 
-class TocPages():
+class TocPages:
 
     def __init__(self, extract):
         self.total_height = 842
         self.header_height = self.compute_header()
         self.footer_height = self.compute_footer()
-        self.disposable_height = 842 - self.header_height - self.footer_height  # A4 size - (footer + header)
+        self.disposable_height = (
+            842 - self.header_height - self.footer_height
+        )  # A4 size - (footer + header)
         self.d1_height = 77  # toc.jrxml
         self.d2_height = 29  # toc.jrxml
         self.d3_height = 61  # toc.jrxml
@@ -21,21 +23,21 @@ class TocPages():
         self.d6_left_height = 38  # toc.jrxml
         self.d6_right_height = 20  # toc.jrxml
         self.extract = extract
-        self.display_qrcode = self.extract['Display_QRCode']
+        self.display_qrcode = self.extract["Display_QRCode"]
         self.total_length = self.compute_total_lenght()
 
     def compute_header(self):
         total_size = 0
-        page_top_margin = 28 # toc.jrxml
-        header_height = 60 # toc.jrxml
+        page_top_margin = 28  # toc.jrxml
+        header_height = 60  # toc.jrxml
         total_size += page_top_margin + header_height
         log.debug(f"header total_size: {total_size}")
         return total_size
 
     def compute_footer(self):
         total_size = 0
-        page_bottom_margin = 20 # toc.jrxml
-        footer_height = 10 # toc.jrxml
+        page_bottom_margin = 20  # toc.jrxml
+        footer_height = 10  # toc.jrxml
         total_size += page_bottom_margin + footer_height
         log.debug(f"header total_size: {total_size}")
         return total_size
@@ -54,7 +56,7 @@ class TocPages():
         total_size += blank_space_above + page_label_height
         toc_item_height = 17  # toc.jrxml (20 in tocConcernedTheme.jrxml)
         unique_concerned_themes = []
-        for concerned_theme in self.extract['ConcernedTheme']:
+        for concerned_theme in self.extract["ConcernedTheme"]:
             if concerned_theme not in unique_concerned_themes:
                 unique_concerned_themes.append(concerned_theme)
         total_size += len(unique_concerned_themes) * toc_item_height
@@ -70,8 +72,12 @@ class TocPages():
         not_concerned_themes_title_height = 15  # toc.jrxml
         blank_space_between = 5  # toc.jrxml
         not_concerned_themes_item_height = 15  # toc.jrxml (12 in themelist.jrxml)
-        total_size += blank_space_above + not_concerned_themes_title_height + blank_space_between
-        total_size += len(self.extract['NotConcernedTheme']) * not_concerned_themes_item_height
+        total_size += (
+            blank_space_above + not_concerned_themes_title_height + blank_space_between
+        )
+        total_size += (
+            len(self.extract["NotConcernedTheme"]) * not_concerned_themes_item_height
+        )
         log.debug(f"d3 total_size: {total_size}")
         if total_size > self.d3_height:
             return total_size
@@ -88,7 +94,9 @@ class TocPages():
     def compute_d5(self):
         total_size = 0
         theme_without_data_item_height = 15  # toc.jrxml (12 in themelist.jrxml)
-        total_size += len(self.extract['ThemeWithoutData'] * theme_without_data_item_height)
+        total_size += len(
+            self.extract["ThemeWithoutData"] * theme_without_data_item_height
+        )
         log.debug(f"d5 total_size: {total_size}")
         if total_size > self.d5_height:
             return total_size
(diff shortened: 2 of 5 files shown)


lopo977:
> adresses #2038
> 
> I completely reworked the calculation:
> 
>     * all the pixel values where verified in the templates and updated when needed
> 
>     * unused variables were deleted
> 
>     * qrcode (if activated) is now included
> 
>     * landregister disclaimer is now included
> 
>     * the ConcernedTheme-Heading cannot be calculated at runtime. The label is defined in the templates which are not accessible from the python code.
> 
> 
> @svamaa @michmuel @voisardf @lopo977 could you test with your data?

It looks good to me.

[CH657302880741-1.pdf](https://github.com/user-attachments/files/16801060/CH657302880741-1.pdf)

michmuel:
https://github.com/openoereb/pyramid_oereb/blob/90affa3945017f14dcfcda32f05a6c2470fb43a4/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py#L79

Are the law status already taken into account here? If a theme has plrs "in force" and plrs "changes" the topic is listed twice in the toc of the pdf.

peterschaer:
> https://github.com/openoereb/pyramid_oereb/blob/90affa3945017f14dcfcda32f05a6c2470fb43a4/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py#L79
> 
> Are the law status already taken into account here? If a theme has plrs "in force" and plrs "changes" the topic is listed twice in the toc of the pdf.

Yes, the law status is taken into account here:
https://github.com/openoereb/pyramid_oereb/blob/90affa3945017f14dcfcda32f05a6c2470fb43a4/pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py#L36
If a theme has plrs of type `inForce` and plrs of type `changeWithoutPreEffect` there are two entries in the `extract['ConcernedTheme']` list and therefore are included in the calculation.

peterschaer:
> > https://github.com/openoereb/pyramid_oereb/blob/90affa3945017f14dcfcda32f05a6c2470fb43a4/pyramid_oereb/contrib/print_proxy/mapfish_print/mapfish_print.py#L79
> > 
> > Are the law status already taken into account here? If a theme has plrs "in force" and plrs "changes" the topic is listed twice in the toc of the pdf.
> 
> Yes, the law status is taken into account here:
> 
> https://github.com/openoereb/pyramid_oereb/blob/90affa3945017f14dcfcda32f05a6c2470fb43a4/pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py#L36
> 
> 
> If a theme has plrs of type `inForce` and plrs of type `changeWithoutPreEffect` there are two entries in the `extract['ConcernedTheme']` list and therefore are included in the calculation.

@michmuel your comment on the daily was correct. the toc calculation is executed before the json is transformed for mapfish_print. Therefore it does not take into account different law_status. I will try to fix this behaviour this afternoon.

peterschaer:
The calculation now takes into account the law_status. The PR can now be tested...

svamaa:
I tested this branch within our oereb-server. I could not find any parcel, that calculates a wrong toc, thus.
Works for us ;) However, we should find out what's different in BL.

voisardf:
Same here, the calculation works for us too. 👍

pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
```diff
@@ -8,101 +8,158 @@
...
+        header_height = 60 # toc.jrxml
+        total_size += page_top_margin + header_height
+        log.debug(f"header total_size: {total_size}")
+        return total_size
+
+    def compute_footer(self):
+        total_size = 0
+        page_bottom_margin = 20 # toc.jrxml
+        footer_height = 10 # toc.jrxml
+        log.debug(f"header total_size: {total_size}")
```
michmuel, line 39: `total_size` is always zero.

michmuel:
I have a question concerning https://github.com/openoereb/pyramid_oereb_mfp/blob/8d9da5ab3a8d1a8469fd693e4f86c5870ee9f6dc/print-apps/oereb/toc.jrxml#L225-L243

What is the use of this?

pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
peterschaer: Well spotted, thanks! I just commited a bugfix.

peterschaer:
> @voisardf, @peterschaer I have a question concerning https://github.com/openoereb/pyramid_oereb_mfp/blob/8d9da5ab3a8d1a8469fd693e4f86c5870ee9f6dc/print-apps/oereb/toc.jrxml#L225-L243 What is the use of this?

This detail contains the NoData heading ("Öffentlich-rechtliche Eigentumsbeschränkungen, zu denen noch keine Daten vorhanden sind").

michmuel:
> > @voisardf, @peterschaer I have a question concerning https://github.com/openoereb/pyramid_oereb_mfp/blob/8d9da5ab3a8d1a8469fd693e4f86c5870ee9f6dc/print-apps/oereb/toc.jrxml#L225-L243 What is the use of this?
> 
> This detail contains the NoData heading ("Öffentlich-rechtliche Eigentumsbeschränkungen, zu denen noch keine Daten vorhanden sind").

I selected the wrong lines. I meant these lines:
https://github.com/openoereb/pyramid_oereb_mfp/blob/8d9da5ab3a8d1a8469fd693e4f86c5870ee9f6dc/print-apps/oereb/toc.jrxml#L244-L255

peterschaer:
> > > @voisardf, @peterschaer I have a question concerning https://github.com/openoereb/pyramid_oereb_mfp/blob/8d9da5ab3a8d1a8469fd693e4f86c5870ee9f6dc/print-apps/oereb/toc.jrxml#L225-L243 What is the use of this?
> > 
> > 
> > This detail contains the NoData heading ("Öffentlich-rechtliche Eigentumsbeschränkungen, zu denen noch keine Daten vorhanden sind").
> 
> I selected the wrong lines. I meant these lines: https://github.com/openoereb/pyramid_oereb_mfp/blob/8d9da5ab3a8d1a8469fd693e4f86c5870ee9f6dc/print-apps/oereb/toc.jrxml#L244-L255

This detail contains the names of the topics having no data. It can be empty.

pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
```diff
@@ -5,116 +5,197 @@
...
+        log.debug(f"d4 total_size: {self.d4_height}")
         return self.d4_height
 
     def compute_d5(self):
-        x = len(self.extract['ThemeWithoutData'] * self.theme_without_data_item_height)
-        if x > self.d5_height:
-            return x
+        total_size = 0
+        theme_without_data_item_height = 15  # toc.jrxml (12 in themelist.jrxml)
+        total_size += len(
```
svamaa, line 97: total_size should be calculated like this: **len(**self.extract["ThemeWithoutData"]**)** * theme_without_data_item_height
Also, the theme_without_data_item_height is 15 only for one entry. 12 is the line height of an individual entry, plus padding. (@voisardf )

peterschaer: I just pushed a bugfix adressing this. Did I understand it correctly?

voisardf:
@peterschaer @michmuel I checked the templates and found, that some spacings and alignments were (in my opinion) over time not defined as they should have or could have. I adapted them slightly and pushed a draft version to https://github.com/openoereb/pyramid_oereb_mfp/pull/147

I'm not perfectly sure if the size of the fonts in the new templates is ok, but it does not affect line heights. I will check that

voisardf:
TOC_adapted_spacing_line_heights

peterschaer:
I have adapted the calculation code to the changes made by François in the templates (https://github.com/openoereb/pyramid_oereb_mfp/pull/147) and made a few minor corrections to the calculation of the line breaks in the disclaimers. My tests with our data look good. You can test again.

pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
```diff
@@ -5,116 +5,206 @@
...
         else:
-            return content_min_size
+            log.debug("d6 right total_size: {}".format(self.d6_right_height))
+            return self.d6_right_height
 
     def compute_d6(self):
-        x = max(self.compute_d6_left(), self.compute_d6_right()) + self.d6_stuff_y_location
-        if x > self.d6_height:
-            return x
+        total_size = max(self.compute_d6_left(), self.compute_d6_right())
```
svamaa, line 184: D6 also has a top margin of 26px. Did you take that into account?

svamaa (commented):
@peterschaer thank you for your updates! 
I think there's still an issue with Detail 6.
I have now the usecase, that an extract with 3 concerned themes, has 2 pages TOC. However, the calculations seem right (even with the 26px top margin on the detail 6). I think we still miss something here. When measuring D6 in the final pdf in cm, it looks as if it should fit into the space between D5 and the Footer.
Maybe we can have a look at it together?

pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
```diff
@@ -5,116 +5,206 @@
...
+        self.d6_right_height = 20  # toc.jrxml
         self.extract = extract
-        self.total_length = self.compute_total_lenght()
+        self.display_qrcode = self.extract["Display_QRCode"]
+        self.total_length = self.compute_total_length()
+
+    def compute_header(self):
+        total_size = 0
+        page_top_margin = 28  # toc.jrxml
+        header_height = 60  # toc.jrxml
```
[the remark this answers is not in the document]
peterschaer: the header band has 60px height, the subreports contained in the header band have 55px height. Which one is correct?
header

pyramid_oereb/contrib/print_proxy/mapfish_print/toc_pages.py
peterschaer: No, I didn't. I made a bugfix. Thanks for the hint.

peterschaer:
> @peterschaer thank you for your updates! I think there's still an issue with Detail 6. I have now the usecase, that an extract with 3 concerned themes, has 2 pages TOC. However, the calculations seem right (even with the 26px top margin on the detail 6). I think we still miss something here. When measuring D6 in the final pdf in cm, it looks as if it should fit into the space between D5 and the Footer. Maybe we can have a look at it together?

Yes, let's look at it together. Can you call me tomorrow friday?

svamaa:
I have added a 10px buffer since the page break happens just slightly earlier than calculated. We don't know why.
In the example [CH698977976775](https://api.oereb.bs.ch/extract/pdf/?EGRID=CH698977976775) the page breaks at least 10px earlier. 
This way, the calculation works for all cases in BS.
Would you also find some time for further tests with the latest commits @voisardf @michmuel @lopo977 

Otherwise as a solution to conclude this PR, the calculation has greatly improved. one could use a page break (depending on the number of topics) between page 1 and 2, so that the page break is forced, the number of toc pages is mostly fixed and the expected_toc_length can be used instead of the calculation.

michmuel:
I tested the newest version of the branch and there is still one pdf extract for which the computed toc length is not correct. However, I do not know why.

As a question: mixing "mm" and "px" is ok, right?
[search for value="mm"](https://github.com/search?q=repo%3Aopenoereb%2Fpyramid_oereb_mfp%20value%3D%22mm%22&type=code)

svamaa:
> I tested the newest version of the branch and there is still one pdf extract for which the computed toc length is not correct. However, I do not know why.
> 
> As a question: mixing "mm" and "px" is ok, right? [search for value="mm"](https://github.com/search?q=repo%3Aopenoereb%2Fpyramid_oereb_mfp%20value%3D%22mm%22&type=code)

yeah, mixing measurement units should be fine according to [this](https://community.jaspersoft.com/knowledgebase/best-practices/measure-units-jaspersoft-studio/)

voisardf:
@michmuel @svamaa I have the same questions as Anne; if it's only one specific case (or really few cases), I think we could stop inquiring. There is still the conversion between mm and px where there might be a rounding error just adding the one pixel generating a page break. Or a letter g, p, q ... responsible for the missing pixel due to the underline

voisardf:
@michmuel @svamaa I have the same questions as Anne; if it's only one specific case (or really few cases), I think we could stop inquiring. There is still the conversion between mm and px where there might be a rounding error just adding the one pixel generating a page break. Or a letter g, p, q ... responsible for the missing pixel due to the underline

michmuel:
There are a few PDFs for which the TOC computation is not correct.

For example: 

https://oereb-dev.geo.bl.ch/extract/pdf?egrid=CH507859371772

2025-01-07 10:24:42,357 DEBUG [...] header total_size: 88
2025-01-07 10:24:42,357 DEBUG [...] footer total_size: 30
2025-01-07 10:24:42,357 DEBUG [...] d1 total_size: 77
2025-01-07 10:24:42,357 DEBUG [...] d2 total_size: 172
2025-01-07 10:24:42,357 DEBUG [...] d3 total_size: 271
2025-01-07 10:24:42,357 DEBUG [...] d4 total_size: 44
2025-01-07 10:24:42,357 DEBUG [...] d5 total_size: 15
2025-01-07 10:24:42,358 DEBUG [...] d6 left total_size: 120
2025-01-07 10:24:42,432 DEBUG [...] d6 right total_size: 104
2025-01-07 10:24:42,433 DEBUG [...] d6 total_size: 120
2025-01-07 10:24:42,433 DEBUG [...] TOC total page length : 699
2025-01-07 10:24:42,433 DEBUG [...] disposable height: 714
2025-01-07 10:24:42,433 DEBUG [...] number of pages: 1
2025-01-07 10:25:10,534 DEBUG [...] Validation of the TOC length with compute_toc_pages set to True and expected_toc_length set to None
2025-01-07 10:25:11,135 DEBUG [...] True number of TOC pages is 2, expected number was 1
2025-01-07 10:25:11,135 WARNI [...] nbTocPages in result pdf: 2 are not equal to the one predicted : 1, request new pdf
2025-01-07 10:25:11,135 DEBUG [...] Secondary PDF extract call STARTED

I am not so confident that all is correct.

svamaa:
Thanks @michmuel for your feedback. I quickly checked your case. In my opinion, there should not be a page break before the disclaimer. To me, the calculation is rather right. but I cannot understand why the page breaks at all since the content fits into the page (visually and manually measured). 
Either we're still missing a margin somewhere or I'm clueless on how jasperreports breaks pages.

So how can we proceed with this ticket? I see three options for users of this repo:

1. calculation works: enable calculation.

michmuel:
I agree that we should finalize this PR. Two last points:
- It seems that it works better for those printing a QR code. However, the templates for printing or not printing the QR code are correct, right?
- We set `self.total_height = 832` although we do not know the reason for the difference to 842. I tend to define the difference as a configuration parameter. Maybe, values different from 10 are better suited for some settings.

What do you think @peterschaer, @svamaa, @voisardf ?

svamaa:
Thank you for your comments.
I could not find anything wrong regarding the qr-code calculations, with or without it.
I think we can set the page break difference to a variable. That way everyone can set their threshold independently. If you agree I can quickly implement this.

svamaa:
@peterschaer , @michmuel , @voisardf I have added the optional parameter "page_break_difference".
Could you give me a short feedback if that's ok with you?
Once we agree we'll have to fix the tests so that we can merge this PR :)

dev/config/pyramid_oereb.yml.mako
```diff
@@ -89,6 +89,9 @@ pyramid_oereb:
     # not set it will assume that only one TOC page exists, and this can lead to wrong numbering in the TOC, which
     # will be fixed by a second PDF extract call that has an impact on performance.
     compute_toc_pages: false
+    # The calculation of the TOC pages is very detailed in pyramid_oereb. However there are "earlier" page breaks, that are
```
michmuel, line 92: I would change the text slightly: "The calculation of the number of TOC pages is very detailed in pyramid_oereb. However, it is not complete, and page breaks occur slightly earlier in page than expected. With setting the value "page_break_difference" (default: 10px), the apparent (not the real) page height is reduced in order to account for this issue."

michmuel:
Except of the little remark concerning the description of the parameter, I would finalize the tests and merge.

peterschaer:
It's also okay for me to merge the PR.

dev/config/pyramid_oereb.yml.mako
svamaa: Done in this commit: https://github.com/openoereb/pyramid_oereb/pull/2047/commits/7a9de6900e339374716e87bce18ad10548782823

PR #2050, opened 2024-09-03
change-legend-entries-retrieval

michmuel:


svamaa (commented):
Thank you @michmuel for this PR!
I prefer this solution to the other PR.
However, I could neither measure any performance win within the standart pyramid_oereb example nor within our environment in BS. Did your extracts improve in BL?

michmuel:
@svamaa The request for the xml-extract for egrid CH958382784904 (large real estate) seems to be faster by about 1s (10s to 9s in total).

svamaa:
Good to hear, could we get any more tests from @voisardf @peterschaer @lopo977 ?

lopo977:
> Good to hear, could we get any more tests from @voisardf @peterschaer @lopo977 ?

I just tried this branch with a 32-page extract, and here in Tessin, we see less than a 5% difference. However this branch also seems to be 5% faster.

michmuel:
The case where performance could be improved is limited to real estates that are covered by public law restrictions with law status "inKraft" AND "laufende Aenderungen". However, extracts for real estates only covered by plrs of law status "inKraft" should not get slower.

svamaa:
To conclude on the recent testing: 
- in 2 cases (BE & BS) it did not make any difference (I tested with parcels that do have both law status).
- in 2 cases it slightly does (TI & BL). 
- There are no errors found so far.
How shall we proceed with this PR? We could merge it if it does not harm anyone but reduces the time slightly. I'd be fine with that.

peterschaer:
It's ok for me to merge.

voisardf:
Same here. In favour for merging - even if the performance does not improve much maybe it is still better to have the calculation fixed. 👍

svamaa (approved):

svamaa (approved):

PR #2059, opened 2024-10-28
initialize processor on server boot

### LINKED ISSUE #2056: Improve performance of data collection
### Intro
Coming back to #1544 we can see that there are many redundant steps which are executed. This is/was mainly introduced to have a configurable server which can be altered at runtime. Means the data integrator can change content on the database without restarting the server. It turned out, that this usecase is a rare one. Probably it is not used at all. Instead, new data is provided by a regular deploy which means the underlying data and the server is completely re setup.

### Initialising things once

One of the most interesting points which was not touched in the recent refactorings and performance improvements is the initialisation of the processor. It is initialized everytime an ÖREB related endpoint is called. If we agree on the statement made in the intro, it would be one of the most efficient performance catches if we refactor pyramid_oereb to initialize the processor only once at boot time. This would cut down all initilisation process which is done in this method.

- https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/core/views/webservice.py#L198
- https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/core/views/webservice.py#L217
- https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/core/views/webservice.py#L254
- https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/core/views/webservice.py#L270
The first 3 pointers are already improved that way, that they initialize only the real estate source and ommit the rest. This was done to improve performance. We could improve it in general if we do that processor initialisation once on server start up.

=> this has to be discussed, as it is a organizational decision to make pyramid_oereb recognizing the configuration and datasorurces only at boot time and not on every request.

###  Parallelisation

I see on potential place where we could hook in for proper take advantage of parallel processing:
https://github.com/openoereb/pyramid_oereb/blob/master/pyramid_oereb/core/readers/extract.py#L51-L104

Here all iterative querying to the sources is bundled and here we could take action. However we should discuss which technique we want to use and if this should be configurable.

#### asyncio

Since we build up onto recent python versions in this project, we are able to use asyncio in combination with [SQLAlchemy](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html). This is probably the best solution in terms of future proof setup. However it comes with some down sides. Asyncio is not 100% available in all python stack and libs we may depend on. So a major task would be to research where we might be blocked to use that.
The most up side of this solution is its scalability and the resource saving solution we would have.

#### [multiprocessing](https://docs.python.org/3/library/multiprocessing.html) / [threading](https://docs.python.org/3/library/threading.html#module-threading)

A well known way of implementing iterative parallel tasks. We easily could implement that. The main disatvantage here is the forking. Threads in one solution or processes in the other, may introduce much more load onto the metal server in the end. So we should discuss how we can avoid bruteforcing wether the database or our servers in the end. In my opinion we could avoid that with some additional configuration where one can set the number of threads or processes to be allowed.

### SQLAlchemy Session Management

A thing we also need to research, is the way we currently implement our session sharing. It is some home made way to improve things for long time running servers to not collect too many open DB sessions. Currently Iam not aware of the influence that would have in a parallel context. Not for Threading NOR Processing NOR Asyncio.

vvmruder:
Fixes #2056 

As discussed this PR changes pyramid_oereb behaviour. It loads Processor and therefore all related theme configuration on server boot time. Before the theme configuration was initialized on every process. This should provide a recognizable gain in performance.

pyramid_oereb will fail to start if Processor could not be initialized for any reason.

TODO:

- [x]  fix tests
- test in example infrastructures
  - BS ✅
  - BL✅
  - NE ✅
  - TI ✅
  - BE ✅


vvmruder:
@michmuel @voisardf @svamaa : I think this is worth a try now in your infra. Please let me know if you need anything.

vvmruder:
I did a quick manual check. On the master branch the [test request](http://localhost:6543/oereb/extract/json?EGRID=CH113928077734) lasts around 1.7seconds on my laptop. While this branch delivers the same request in 0.24seconds. A nice gain I think.

lopo977:
When running a local test on my workstation, your implementation of the singleton pattern processes the XML request (CH452802850772) in 4.37 seconds, down from 6.26 seconds. Looks good to me.

vvmruder:
> When running a local test on my workstation, your implementation of the singleton pattern processes the JSON request (CH452802850772) in 4.37 seconds, down from 6.26 seconds. Looks good to me.

@lopo977 Thank you for testing. Did you run the request multiple times? For me, the first was a bit longer, the following were much faster than. Can you confirm that?

lopo977:
> > When running a local test on my workstation, your implementation of the singleton pattern processes the JSON request (CH452802850772) in 4.37 seconds, down from 6.26 seconds. Looks good to me.
> 
> @lopo977 Thank you for testing. Did you run the request multiple times? For me, the first was a bit longer, the following were much faster than. Can you confirm that?

I ran the same request multiple times. The first was is longer also for me. I confirm.

svamaa (commented):
I have tested this branch within our local dev environment. Overall I discovered a 9% improvment over 3 different parcels with each 20 requests. However, the improvement is significantly smaller, the larger the parcel and the associated data get. This makes sense.

However, I discovered in general a large variety in response time (this happens also on the **master** branch). 
Even within a small parcel the response time varies up to 2 seconds between the same request. Does anyone else also finds this behaviour on their systems? I did not connect a db on my machine but on our intranet. The network should actually not be a large factor here.

voisardf:
I also am under the impression our extract creation times do vary for the same pdf even between the 2nd and following extracts. I plan to do some more structured testing next week.

peterschaer:
I finally found time to do some tests:

| EGRID  | Description | No. of DB-Queries master | No. of DB-Queries PR | Processing time master | Processing time PR |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| [CH523822463517](https://oerebview.apps.be.ch/#/d/CH523822463517)  | parcel with only one PLR  | 198 | 157 | 4.8s | 0.7s |
| [CH743546874207](https://oerebview.apps.be.ch/#/d/CH743546874207)  | parcel with two topics  | 208 | 167 | 4.9s | 1.2s |
| [CH762046353030](https://oerebview.apps.be.ch/#/d/CH762046353030)  | parcel with six topics  | 261 | 220 | 9.1s | 4.7s |

I made all my tests on my development machine (Win10, no Docker, DB remote (but staging DB, so not a lot of load to be expected)). All values were taken from the Pyramid Debug Toolbar. All requested extracts were XML extracts.

All in all a very significant gain!

michmuel:
I compared the branch of this PR with the master branch on our dev instance. The request of a XML extract is faster using the branch of this PR. The gain is in the order as outlined by @peterschaer. That's great!

We use the oereblex source for many oereb topics. In our extended oereblex sources we handle the indexing of the legal provisions which are retrieved for the relevant public law restrictions at each request from oereblex. The index is required for the extract and shall be in the range -1000 to 1000. At least in our instance of oereblex, legal provisions do not have an index (only laws have). When all the sources are initialised only at server start up, we will have to change the reset of this index. That should be faesible.

Just important to note that request related information attached to sources may have to be handled differently with this changes.

svamaa:
Thank you all for testing! 
@vvmruder, I think you can start implementing the testing.

michmuel:
With the new approach, oereblex links are cached from server start-up (https://github.com/openoereb/pyramid_oereb/blob/176b4b3245ed243d2892c9a604e370751283a98f/pyramid_oereb/contrib/data_sources/oereblex/sources/plr_oereblex.py#L99). This means that changes in oereblex for links that have been invoked by pyramid oereb will not take effect in the app till server restart.

Shall we change the caching behavior? Any opinions? (e.g. @jwkaltz?)

jwkaltz:
> With the new approach, oereblex links are cached from server start-up (
> 
> https://github.com/openoereb/pyramid_oereb/blob/176b4b3245ed243d2892c9a604e370751283a98f/pyramid_oereb/contrib/data_sources/oereblex/sources/plr_oereblex.py#L99
> 
> ). This means that changes in oereblex for links that have been invoked by pyramid oereb will not take effect in the app till server restart.
> Shall we change the caching behavior? Any opinions? (e.g. @jwkaltz?)

@michmuel We already have a caching of oereblex links (see https://github.com/openoereb/pyramid_oereb/pull/993), this fills the links as needed. I don't think we should cache all links on server start-up.
Maybe I misunderstand?

michmuel:
> > With the new approach, oereblex links are cached from server start-up (
> > https://github.com/openoereb/pyramid_oereb/blob/176b4b3245ed243d2892c9a604e370751283a98f/pyramid_oereb/contrib/data_sources/oereblex/sources/plr_oereblex.py#L99
> > 
> > ). This means that changes in oereblex for links that have been invoked by pyramid oereb will not take effect in the app till server restart.
> > Shall we change the caching behavior? Any opinions? (e.g. @jwkaltz?)
> 
> @michmuel We already have a caching of oereblex links (see #993), this fills the links as needed. I don't think we should cache all links on server start-up. Maybe I misunderstand?

Let me know if I do not understand correctly, but I think, today, it works like this:
1. Server start
2. Request for parcel XY -> get public law restrictions -> download corresponding oereblex xmls (different plrs may have the same lexlinks but every lexlink is downloaded only once due to caching)
3. Request for parcel XY again -> get public law restrictions -> download corresponding oereblex xmls (different plrs may have the same lexlinks but every lexlink is downloaded only once due to caching)

Changes / corrections in Oereblex are active at least for the next request.

With this pull request the behavior changes as the oereblex classes are initialised only on server start:
1. Server start
2. Request for parcel XY -> get public law restrictions -> download corresponding oereblex xmls (different plrs may have the same lexlinks but every lexlink is downloaded only once due to caching)
3. Request for parcel XY again -> get public law restrictions -> no lexlink is downloaded because all lexlinks are cached

Changes / corrections in Oereblex are active immediately for lexlinks that have not been requested since server start and after next server start for lexlinks that have been requested since server start.

vvmruder:
@michmuel You were right with your assumptions.

It took me a while to follow the path and find a maybe nice solution without changing too much in the code. However, I would be happy if you could check this out first on some real world data. Is that possible?

michmuel:
> @michmuel You were right with your assumptions.
> 
> It took me a while to follow the path and find a maybe nice solution without changing too much in the code. However, I would be happy if you could check this out first on some real world data. Is that possible?

Thanks @vvmruder 

Looking into the logs I found that a specific geolink is retrieved once for every request. If more than one plr with the same geolink is found for a real estate, the geolink is not retrieved again.

However, is it robust in an uwsgi-setting with many threads? Can you provide some reasoning?

vvmruder:
@michmuel You made a valid point. I fixed it now. You may have a look if that is how you expected it.

michmuel:
Hi @vvmruder 
Your solution concerning OerebLex looks good to me. Thanks!

vvmruder:
Happy new year everyone and thanks for the feedback. I now fix the tests.

peterschaer (approved):
Looks good to me. Thanks!

svamaa:
Thanks for your fixes. It works for me too.
How did we leave things regarding test coverage? Could you provide the respecting test coverage for your new code?
I also saw, there's an issue from the  [Codacy Static Code Analysis](https://app.codacy.com/gh/openoereb/pyramid_oereb/pull-requests/2059/issues#issue-167bf629e14e1e25613b6269261ab9f8).

michmuel:
I agree with @svamaa that the tests should be successful.

michmuel:
@vvmruder, @svamaa How do we proceed?

vvmruder:
@svamaa & @michmuel I now did my best to satisfy codecov. For some reason it points out the plr.py file which I actually did not really touch. The init file in pyramid_oereb lib root is for server start up and was never really tested. Since this is a bit bogus to do, I would now like to state:
That task is finished :grimacing:

svamaa (approved):

michmuel (approved):

PR #2068, opened 2024-11-04
Sphinx and Sphinx_rtd simulateneous update due to mutual dependency

jwkaltz:


jwkaltz:
Requires Python > 3.9, but 3.9 is not yet end-of-life

PR #2102, opened 2025-03-20
2101 fix daily ci

domfre:
The CI failed due to a failing test (see changes). The code uses the Shapely dependency which itself uses Numpy. However, the version of Numpy used depends on the underlying Python base interpreter. For Python versions 3.10, 3.11 and 3.12, Numpy v2.2.4  is used, whereas for Python version 3.9 Numpy v2.0.2 is used. Somehow, the different Numpy versions apply different calculations resulting in different values for the y-coordinate in the affected test (again, see changes). This PR provides a temporary workaround to handle the different Python versions. Needs to be corrected once support for Python 3.9 is dropped.


jwkaltz (approved):
Well spotted, thanks!

PR #2196, opened 2026-07-14
2191 getegrid   postalcode localisation number optional

domfre:
This pull request makes the parameter `street_number` in the `get_egrid` method, specifically in the `_get_egrid_address` method of the `pyramid_oereb/core/views/webservice.py` optional.

Furthermore, it propagates this change to any concerned module.

At the same time, certain improvements regarding type annotation were made in order to make the code base more readable.

### DIFF
--- a/pyramid_oereb/contrib/data_sources/swisstopo/address.py
+++ b/pyramid_oereb/contrib/data_sources/swisstopo/address.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
 import requests
 from pyreproj import Reprojector
+from requests import Response
 from shapely.geometry import Point
 
 from pyramid_oereb import Config
 from pyramid_oereb.core.records.address import AddressRecord
 from pyramid_oereb.core.sources.address import AddressBaseSource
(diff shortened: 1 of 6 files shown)


pyramid_oereb/contrib/data_sources/swisstopo/address.py
```diff
@@ -37,52 +39,65 @@ def __init__(self, **kwargs):
...
         if self._referer is not None:
             headers.update({
                 'Referer': self._referer
             })
-        request_params = {
+        request_params: dict[str, str | None] = {
             'type': self._type,
             'origins': self._origins,
-            'searchText': u'{0} {1} {2}'.format(zip_code, street_name, street_number)
+            'searchText': f'{zip_code} {street_name} {street_number}'
```
michmuel: If "NUMBER" is not set as a query parameter, "street_number" is "None". In this case, the "searchText" should be "f'{zip_code} {street_name}'" instead of f'{zip_code} {street_name} {street_number}'.

michmuel (changes requested):

